PR #5, opened 2015-08-10
fixes for building on os x

wjwwood:
These are just a few fixes for compiler errors and warnings on OS X.


richiprosima:
Last commit from our internal Git contains your suggestions. Thanks

PR #6, opened 2015-08-10
add package.xml to enable ament builds

wjwwood:
See: https://github.com/eProsima/Fast-CDR/pull/2


dirk-thomas:
With ament/ament_tools#49 `ament_tools` is able to identify the CMake project as well as (some of) its dependencies automatically without having a manifest. Therefore this seems to be not necessary anymore.

richiprosima:
Thanks, I close the pull request

PR #7, opened 2015-09-16
add requested boost components to the error message

dirk-thomas:


dirk-thomas:
Another approach would be to pass `REQUIRED` in the line above:

    find_package(Boost REQUIRED COMPONENTS ${ARGN})

Then you can remove all the custom checking and error messaging since CMake will fail directly if the dependencies can't be found.

richiprosima:
Last commit from our internal Git contains your last suggestion. Thanks.

dirk-thomas:
Will the change be available in this repository at some point?

dirk-thomas:
Seems to be in this repo by now: https://github.com/eProsima/Fast-RTPS/blob/eb6ecddf018ddd0552c25a3ed298546e2c4fe92f/cmake/dev/boost_libraries.cmake#L21

PR #45, opened 2016-06-04
Add mutex to functions in ReaderHistory

jacquelinekay:
Using [Clang ThreadSanitizer](http://clang.llvm.org/docs/ThreadSanitizer.html) with some ROS 2 examples, I discovered a data race in ReaderHistory: `m_historyRecord` could be written to from `ReaderHistory::add_change` while other functions reading `m_historyRecord` are called in other threads. I have added acquisition of `mp_mutex` to other functions which access m_historyRecord.

I thought this would fix some issues with mysterious values appearing in serialization of bools that @dirk-thomas is debugging right now, but it does not appear to fix the problem on my side.


richiprosima:
Hi @jacquelinekay

Your PR was added by us in @0d6e9234a before the change to Apache License
2.0. Sorry for the inconveniences. I will close this PR and from now on with
the new license we will improve our pull request practices. Thanks for your contributions.

Issue #44, opened 2016-06-04
BlackBox tests fail for an OSX build with build type Debug


jacquelinekay:
On the ROS 2 CI  farm, we observe that the BlackBox tests fail in the case "BlackBox.AsyncPubSubAsNonReliableData300kb" on an OSX machine built with -DCMAKE_BUILD_TYPE=Debug.

http://ci.ros2.org/view/nightly/job/nightly_osx_debug/101/testReport/projectroot.test/blackbox/BlackboxTests/

richiware:
We will check it in a OSX machine.

gerkey:
Here's the console output from a recent failure:
```
11:15:03 1: [ RUN      ] BlackBox.AsyncPubSubAsNonReliableData300kb
11:15:03 1: Writer waiting for discovery...
11:15:03 1: Writer discovery phase finished
11:15:03 1: Reader waiting for discovery...
11:15:03 1: Reader discovery phase finished
11:21:44 1: Samples not received: 11 12 13 14 15 16 17 18 19 20 21 22 23 24 25 26 27 28 29 30 
11:21:44 1: /Users/osrf/jenkins/workspace/nightly_osx_debug/ws/src/eProsima/Fast-RTPS/test/blackbox/BlackboxTests.cpp:720: Failure
11:21:44 1: Value of: 0
11:21:44 1: Expected: data.size()
11:21:44 1: Which is: 20
11:21:44 1: [  FAILED  ] BlackBox.AsyncPubSubAsNonReliableData300kb (400770 ms)
```

It looks like `AsyncPubSubAsNonReliableData300kb` takes ~6.5 minutes to run (perhaps related to #46) and fails.

gerkey:
I can reproduce the delay and failure locally, having built Fast-CDR and Fast-RTPS in Debug:
```
$ ./test/blackbox/BlackboxTests --gtest_filter="*AsyncPubSubAsNonReliableData300kb"
Note: Google Test filter = *AsyncPubSubAsNonReliableData300kb
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from BlackBox
[ RUN      ] BlackBox.AsyncPubSubAsNonReliableData300kb
Writer waiting for discovery...
Writer discovery phase finished
Reader waiting for discovery...
Reader discovery phase finished
Samples not received: 10 12 13 14 15 16 17 18 19 20 21 22 23 24 27 28 29 30 
[...]

PR #47, opened 2016-06-15
Disable failing test on OSX

gerkey:
Temporary workaround for #44.

This obviously isn't the right fix, but if it will take some time to track down the underlying cause, it would help to clean up our CI to have this merged temporarily.


richiprosima:
Hi @gerkey

I will close your PR because in new release 1.2.0 non-reliable tests were
refactored and now they should work successfully in OSX.

Thanks.

dirk-thomas:
Not all of them, see #50.

PR #58, opened 2016-07-06
Set default durability for WriterQos to match documentation

dhood:
The documentation specifies [here](https://github.com/eProsima/Fast-RTPS/blob/3cf7996c9cea21a4bdc4ee5f8dc6763069cd85f2/include/fastrtps/qos/QosPolicies.h#L60) that the default durability QoS for publishers/writers should be `TRANSIENT_LOCAL_DURABILITY_QOS` however it is not currently being set. As a result publishers currently have the default durability settings of `VOLATILE_DURABILITY_QOS` due to https://github.com/eProsima/Fast-RTPS/blob/3cf7996c9cea21a4bdc4ee5f8dc6763069cd85f2/include/fastrtps/qos/QosPolicies.h#L74

### DIFF
--- a/src/cpp/qos/WriterQos.cpp
+++ b/src/cpp/qos/WriterQos.cpp
@@ -28,6 +28,7 @@ static const char* const CLASS_NAME = "WriterQos";
 WriterQos::WriterQos()
 {
 	this->m_reliability.kind = RELIABLE_RELIABILITY_QOS;
+	this->m_durability.kind = TRANSIENT_LOCAL_DURABILITY_QOS;
 }
 
 WriterQos::~WriterQos()


dhood:
I've noticed that this line was removed in https://github.com/eProsima/Fast-RTPS/commit/93088db71c088dc04ffa77418c2492d5012b8b13

This leads me to believe that now `VOLATILE_DURABILITY_QOS` is intended as the default for both reader and writer and that the documentation is just out of date. Is this the case? (it would match with the [durability defaults for RTI Connext](http://community.rti.com/rti-doc/510/ndds/doc/html/api_cpp/group__DDSDurabilityQosModule.html#ga4bc6b72c72928cb9ed6432e3e6a83e2e)).

richiprosima:
Thanks for your help

PR #61, opened 2016-08-25
Add pthread, needed on some ARM platforms

gerkey:
As reported in ros2/ros2#260, `-lpthread` is needed when linking one of the examples on some ARM/Linux machines (e.g., rpi3).

### DIFF
--- a/examples/C++/DeadlineQoSExample/CMakeLists.txt
+++ b/examples/C++/DeadlineQoSExample/CMakeLists.txt
@@ -37,7 +37,11 @@ file(GLOB DEADLINEQOS_EXAMPLE_SOURCES "*.cxx")
 
 add_executable(DeadlineQoSExample ${DEADLINEQOS_EXAMPLE_SOURCES})
 target_include_directories(DeadlineQoSExample PRIVATE ${Boost_INCLUDE_DIR})
-target_link_libraries(DeadlineQoSExample fastrtps fastcdr ${Boost_LIBRARIES})
+if(UNIX)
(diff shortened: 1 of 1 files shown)


examples/C++/DeadlineQoSExample/CMakeLists.txt
```diff
@@ -37,7 +37,11 @@ file(GLOB DEADLINEQOS_EXAMPLE_SOURCES "*.cxx")
 
 add_executable(DeadlineQoSExample ${DEADLINEQOS_EXAMPLE_SOURCES})
 target_include_directories(DeadlineQoSExample PRIVATE ${Boost_INCLUDE_DIR})
-target_link_libraries(DeadlineQoSExample fastrtps fastcdr ${Boost_LIBRARIES})
+if(UNIX)
+  target_link_libraries(DeadlineQoSExample fastrtps fastcdr ${Boost_LIBRARIES} pthread)
+else()
+  target_link_libraries(DeadlineQoSExample fastrtps fastcdr ${Boost_LIBRARIES})
+endif()
```
dirk-thomas: To avoid duplicating all this logic I would suggest to update the patch to:

```
target_link_libraries(DeadlineQoSExample fastrtps fastcdr ${Boost_LIBRARIES})
if(UNIX)
  target_link_libraries(DeadlineQoSExample pthread)
endif()
```

scpeters: +1

richiprosima:
Thanks for the PR. I've also applied @dirk-thomas  suggestion.

PR #65, opened 2016-09-30
introduce user transport option to set time to live

Karsten1987:
I extended the `UDPv*TransportDescriptor.h` of a field called `TTL` in order to specify the time to live for each packet.
By default it's set to 1, but totally up to discussing whether or not this should be done like that or not.

Addresses the conflicts in https://github.com/eProsima/Fast-RTPS/issues/64

### DIFF
--- a/include/fastrtps/transport/UDPv4Transport.h
+++ b/include/fastrtps/transport/UDPv4Transport.h
@@ -170,6 +170,7 @@ protected:
    uint32_t mMaxMessageSize;
    uint32_t mSendBufferSize;
    uint32_t mReceiveBufferSize;
+   uint8_t mTTL;
 
    boost::asio::io_service mService;
    std::unique_ptr<boost::thread> ioServiceThread;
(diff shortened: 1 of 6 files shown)


wjwwood:
Anyone on the eProsima team have time to weigh in on this? @richiprosima maybe?

I think this a pretty important configuration to have available for ROS 2. In fact I think we would default it to TTL 0 and have a global configuration to change it to TTL 1. The reason being is that if someone tries our image pipeline demos (which push a lot of data) then it can take down a wifi network because of the mass multicast data.

If you guys don't have time to merge it and release it, we will probably distribute ROS 2 with this patch on top of Fast-RTPS so we can control it, but even in that case it would be good to get feedback from you guys about whether or not this is a reasonable approach.

richiprosima:
@Karsten1987 Can you reopen this PR? @JavierIH was working last week checking multicast scenarios with this PR. After his work he will merge it. Thanks

Karsten1987:
@richiprosima thanks for catching up on this. @JavierIH Please feel free to contact me if I can help in any case.

PR #81, opened 2017-03-09
Added additional NDK variables to the Android build options

spiderkeys:
This is just to make sure that any android variables passed to the FastRTPS build make their way to the FastCDR build.


richiprosima:
Thanks

PR #91, opened 2017-04-17
Check the CMake registry for pre-configured Asio and TinyXML2 packages.

nuclearsandwich:
Find modules override the CMake package registry by default.  On Windows, we have no standard file system hierarchy to check for installed libraries and using the CMake package registry [[1]] is the easiest way to advertise installed dependencies to dependents.

These changes will force CMake to look in the package registry by passing the CONFIG directive to find_package, if no package is found, the existing find module logic will be used.

This change should not affect builds on systems where the existing logic is successful unless those systems also include entries in the CMake package registry and those entries are invalid.

[1]: https://cmake.org/cmake/help/v3.5/manual/cmake-packages.7.html#id20

### DIFF
--- a/cmake/modules/FindAsio.cmake
+++ b/cmake/modules/FindAsio.cmake
@@ -3,7 +3,7 @@
 
 find_package(Asio CONFIG QUIET)
 if(Asio_FOUND)
-    message(STATUS "Using Asio package from CMake registry")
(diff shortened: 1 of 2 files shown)


richiprosima:
Thanks @nuclearsandwich

PR #110, opened 2017-06-20
Fix cross compiling examples

theomta:
pass down cmake toolchain arg to examples subproject

Signed-off-by: Teodor Madan <teodor.madan@nxp.com>


cmake/dev/compile_example.cmake
```diff
@@ -22,6 +22,7 @@ macro(compile_example example example_directory)
         # Separate CMAKE_PREFIX_PATH
         string(REPLACE ";" "|" CMAKE_PREFIX_PATH_ "${CMAKE_PREFIX_PATH}")
         set(${example}_CMAKE_ARGS
+			"-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}"
```
richiware: Can you indent this line like the others?

richiware:
Since version 1.6.0 we are not using ExternalProject_Add to compile examples. Instead of it we use add_subdirectory. This PR is obsolete. I will close it.

PR #114, opened 2017-06-23
work around to support TinyXML2 5.0.0

wjwwood:
This is the most minimal patch I could make that fixes https://github.com/eProsima/Fast-RTPS/issues/113.

A more complete fix might be to convert that custom FindTinyXML2 file to always produce an imported target and update the rest of the project's CMake code to not depend on things like `TINYXML2_INCLUDE_DIR` and `TINYXML2_SOURCE_DIR`.


richiprosima:
Thanks @wjwwood

PR #119, opened 2017-07-04
Fix compilation issue with gcc v7.1

rojkov:
When compiling with gcc v7.1 the following error message gets
trggered:
```
  /home/rojkov/work/ros/build/tmp-glibc/work/i586-oe-linux/fastrtps/git-r0/git/include/fastrtps/rtps/writer/RTPSWriter.h:66:54: error: 'function' in namespace 'std' does not name a template type
       RTPS_DllAPI CacheChange_t* new_change(const std::function<uint32_t()>& dataCdrSerializedSize,
                                                        ^~~~~~~~
```
The patch adds a missing header required by the compiler.


richiprosima:
This was also fixed on #112. Thanks @rojkov

PR #124, opened 2017-07-27
Generalize 32 and 64 bit support

mcharleb:
ARM compiler doesn't support the -m flag. Add new platform for arm to omit the flag. 

Signed-off-by: Mark Charlebois <charlebm@gmail.com>


richiware:
Thanks for your contribution

PR #130, opened 2017-08-07
Fix typo in XML API

tparys:
Was using the XML endpoint the other day, and ran across "fillSubscribertAttributes" and "fillPublishertAttributes" methods which looked to be typos (extra T in each).

Feel free to ignore if intentional.


richiware:
Thanks for fixing the typo.

PR #139, opened 2017-08-23
Allow XML profiles tag to be nested inside a root tag.

### LINKED ISSUE #138: Feature: Allow end-user application XML config files to include non-FastRTPS settings
It would be useful in some applications to include settings meant for other parts of the code in the same XML config containing the FastRTPS settings so the application only has one configuration file. This is possible with the current implementation but the loadXMLFile routine expects the ```<profiles>``` tag to be the root tag of the XML document.  Other settings can be dumped inside that tag but FastRTPS will warn about unexpected tags in its configuration.

I propose that if the ```<profiles>``` tag is not the root tag, it is searched for as a first level child.  This would allow for the inclusion of FastRTPS configuration alongside user settings with XML files like:
```xml
<?xml version="1.0"?>

<any-root-tag>

    <user-application-configs>
        <setting>7</setting>
    </user-application-configs>

    <profiles>
        <!--FastRTPS configs -->

        <participant profile_name="application">
            <!--Participant details -->
        </participant>

        <publisher profile_name="raw_rx">
            <!--Publisher details -->
        </publisher>
    </profiles>
</any-root-tag>
```

While this is not endlessly flexible, it does allow FastRTPS to play nicely with other content in XML files with the restriction that ```<profiles>``` be a top level child tag.

I could also understand the philosophy of keeping FastRTPS configs separate and expecting any other settings to be in a different file but a single file would be a cleaner solution in our current use case.

wtsnyder:
resolves #138

### DIFF
--- a/src/cpp/xmlparser/XMLProfileParser.cpp
+++ b/src/cpp/xmlparser/XMLProfileParser.cpp
@@ -80,16 +80,39 @@ XMLP_ret XMLProfileParser::loadXMLFile(const std::string &filename)
 
     logInfo(XMLPROFILEPARSER, "File '" << filename << "' opened successfully");
 
-    XMLElement* p_root = xmlDoc.FirstChildElement(PROFILES);
-    if (nullptr == p_root)
+    XMLElement* p_profiles = xmlDoc.FirstChildElement(PROFILES);
+    if (nullptr == p_profiles)
     {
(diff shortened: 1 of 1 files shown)


richiware:
As mentioned in #138 currently we don't want to change XML formats because there are users using it. Sorry.

PR #153, opened 2017-10-06
Add ability to apply the whitelist to input sockets and locators.

deng02:
The UDP descriptors have been modified to support flags that
allow the interface white list to be applied to output sockets,
input sockets and/or locators.

If applied to input sockets, rather than binding to IP address
any (0.0.0.0), the input socket will bind to the multicast IP
if the locator is multicast or to specific allowed interface IPs.

If applied to locators, then the default unicast locator list
will be filtered against the allowed interface IPs.  It is not
applied in the case where the user has specified a default list.

### DIFF
--- a/include/fastrtps/rtps/network/NetworkFactory.h
+++ b/include/fastrtps/rtps/network/NetworkFactory.h
@@ -77,6 +77,8 @@ class NetworkFactory
 
         void NormalizeLocators(LocatorList_t& locators);
 
+        void FilterLocators(LocatorList_t& locators);
+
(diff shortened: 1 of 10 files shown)


deng02:
Addresses https://github.com/eProsima/Fast-RTPS/issues/150

deng02:
@richiware  Is there any interest in this PR? Due to the nature of the network in which we are working in, we are in need of something that allows us to isolate input channels as well as output.

MiguelCompany:
@deng02  As this has been rebased on #231 and it is being discussed there, I will mark this as duplicate and close this PR

MiguelCompany:
Closing as duplicate of #231

PR #160, opened 2017-10-27
Change --std=c++11 to -std=c++11.

clalancette:
While gcc accepts either, older clang on Ubuntu Xenial (3.8) only accepts `-std=c++11` (single dash).


richiware:
Thanks for the contribution

PR #162, opened 2017-11-17
Fix issue of wrapper object use after free

gaoethan:
The internal representation of temporary of type
`std::string` is freed by its destructor in `c_str()`

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>


richiware:
Thanks for your contribution

Issue #165, opened 2017-12-05
make API const correct


dirk-thomas:
Many APIs could / should be `const` but are not declared as such. E.g. [SubscriberHistory::getUnreadCount()](https://github.com/eProsima/Fast-RTPS/blob/bec4a979c088a1f978af3dfdea2a72ae2a78b52a/include/fastrtps/subscriber/SubscriberHistory.h#L111) which simply returns the value of a member could be declared `const`.

But this applies to a lot of functions in the API. Is there a reason why those are not declared as `const`?

richiware:
First developer of FastRTPS doesn't have the habit of use ``const``. Now, as we keep modifying the code, we try to use it. Thanks for your suggestions.

dirk-thomas:
It might be an option to add `const` functions beside the current non-const versions so that users can start using those. That would allow to deprecate and at some point remove the non-const ones sometime in the future.

Are you interested in doing this and/or would you accept pull requests towards this direction?

richiware:
At this moment we have no time to do this. We appreciate any PR. Thanks

PR #170, opened 2017-12-15
add const where appropriate

dirk-thomas:
Addresses some cases of #165.


richiware:
Thanks for your contribution

PR #175, opened 2018-01-04
Fix warning on windows about hidden local declaration

sloretz:
Currenly `SecurityManager.cpp:201` and `SecurityManager.cpp:215` cause the warning `declaration of 'exception' hides previous local declaration` on windows. This moves outer declaration  lower to eliminate them.


richiware:
Thanks. We will check the PR next week and merge it.

richiware:
Thanks for the contribution

PR #188, opened 2018-03-02
Fix windows build for VS2015 and newer

cmpute:
There's a little bug in CMakeList prevent me from successfully building the project. I guess it was a forgotten fix.

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -416,7 +416,8 @@ if(MSVC OR MSVC_IDE)
             PATTERN "*.pdb"
             )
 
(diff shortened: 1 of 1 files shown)


cmpute:
I'm using MSVC v140 toolset in VS2017

richiware:
Finally I installed VS2017 in my machine. I've compiled fastrtps succesfully with next CMake configuration (VS2017 with MSVC v140):

`cmake -DTHIRDPARTY=ON -T v140`

Can you paste the output with your error?

MiguelCompany:
As we cannot reproduce the problem, and this was answered long ago, we are closing this. Feel free to reopen if you think it is still an issue for you.

Issue #190, opened 2018-03-09
Mutex lock ordering problem between PDP and EDP classes [4627]


guillaumeautran:
There seems to be some mutex ordering problems between the EDP and PDP class such it causes a deadlock when a participant endpoint is deleted and a subscription to that endpoint is created at the same time.

guillaumeautran:
Hellgrind trace looks like this:
{code}
==18153== ----------------------------------------------------------------
==18153== 
==18153== Thread #15: lock order "0x16E56540 before 0x16E4AD70" violated
==18153== 
==18153== Observed (incorrect) order is: acquisition of lock at 0x16E4AD70
==18153==    at 0x4C321BC: ??? (in /usr/lib/valgrind/vgpreload_helgrind-amd64-linux.so)
==18153==    by 0x8D7EBFF: __gthread_mutex_lock(pthread_mutex_t*) 
==18153==    by 0x8D7EC4F: __gthread_recursive_mutex_lock(pthread_mutex_t*) 
==18153==    by 0x8D7F5A7: std::recursive_mutex::lock() 
==18153==    by 0x8D7F7F7: std::lock_guard<std::recursive_mutex>::lock_guard(std::recursive_mutex&) 
==18153==    by 0x8E57FC6: eprosima::fastrtps::rtps::PDPSimple::removeReaderProxyData(eprosima::fastrtps::rtps::GUID_t const&) 
==18153==    by 0x8E6D887: eprosima::fastrtps::rtps::EDPSimple::removeLocalReader(eprosima::fastrtps::rtps::RTPSReader*) 
==18153==    by 0x8E55868: eprosima::fastrtps::rtps::BuiltinProtocols::removeLocalReader(eprosima::fastrtps::rtps::RTPSReader*) 
==18153==    by 0x8DFBB10: eprosima::fastrtps::rtps::RTPSParticipantImpl::deleteUserEndpoint(eprosima::fastrtps::rtps::Endpoint*) 
==18153==    by 0x8E01DC7: eprosima::fastrtps::rtps::RTPSDomain::removeRTPSReader(eprosima::fastrtps::rtps::RTPSReader*) 
==18153==    by 0x8E1ADB5: eprosima::fastrtps::SubscriberImpl::~SubscriberImpl() 
==18153==    by 0x8E1AE7F: eprosima::fastrtps::SubscriberImpl::~SubscriberImpl() 
==18153== 
==18153==  followed by a later acquisition of lock at 0x16E56540
[...]

PR #192, opened 2018-03-10
Fixed the lock ordering between EDP and PDP

guillaumeautran:
Fixed a deadlock problem when removing writer proxies / reader proxies releasing the PDP mutex pior to going down proxies. 

Also, converted some pointers to smart pointers to avoid memory leakage.

Issue: #190

### DIFF
--- a/include/fastrtps/rtps/builtin/data/ParticipantProxyData.h
+++ b/include/fastrtps/rtps/builtin/data/ParticipantProxyData.h
@@ -21,6 +21,7 @@
 #define PARTICIPANTPROXYDATA_H_
 #ifndef DOXYGEN_SHOULD_SKIP_THIS_PUBLIC
 #include <mutex>
+#include <memory>
 #include "../../../qos/QosList.h"
 #include "../../../qos/ParameterList.h"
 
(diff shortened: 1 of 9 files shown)


dhood:
This PR resolves the deadlocking described in https://github.com/eProsima/Fast-RTPS/issues/200

dhood:
Could you let us know if this is likely to be reviewed in the next week, please? (We understand that your team is busy, we just need to do something about the ROS 2 tests that are failing from https://github.com/eProsima/Fast-RTPS/issues/200, and an ETA on this PR will help us decide if we should pin fastrtps or just wait). Thanks!

JaimeMartin:
Hi dhood,

yes, we will review this PR next week.

richiware:
@guillaumeautran Last week a fellow worker tells me he detected a segmentation fault using your PR. I've checked it right now and get it too. 

```
1: [ RUN      ] BlackBox_PreallocMem.EndpointRediscovery
1: Writer is waiting discovery...
1: Writer discovery finished...
1: Reader is waiting discovery...
1: Reader discovery finished...
1: Writer is waiting removal...
1: Writer removal finished...
1: Writer is waiting discovery...
1/1 Test #1: BlackboxTests_PreallocMem ........***Exception: SegFault 28.27 sec
```

It's good your intention of use shared_ptr, but I believe it has repercutions  knowing how that part is designed.
Anyway you can run our tests using `ctest`. Previously our `CMakelists.txt` has to have detected googletest libraries.

guillaumeautran:
Thanks for the feedback! Let me have a look.

guillaumeautran:
Yeap, I see it now. The problem is the suicidal timer [TimedEventImpl.cpp](https://github.com/eProsima/Fast-RTPS/blob/master/src/cpp/rtps/resources/TimedEventImpl.cpp#L221) deleting itself. With shared pointers, once the last reference is cleared, the memory is deleted without delay. 

Now, does that not worry you to have a class basically doing a delete on the this pointer?

PR #199, opened 2018-03-14
Fix potential unused-parameter error in GCC code generated by fastrtp…

scherererer:
Generated serializeKey() functions that have an empty body could produce the error.


MiguelCompany:
Hey @scherererer !

Thanks for this, but we already did it on commit 4724e919d7b19b5b5e5448ffb559eaab514c45c7

scherererer:
Thanks, closing because of fix in 4d5f197

PR #216, opened 2018-05-13
Fix cross-compilation by forwarding cmake_toolchain_file

dgoel:
Without this change, compiler detection fails for eprosima libraries
since the toolchain file, which may contain the sysroot path necessary
for compilation to work properly, is not forwarded to cmake.


dgoel:
See this comment for more details about the build failure that this PR is trying to address:

https://github.com/ros2-for-arm/ros2/issues/4#issuecomment-389682844

richiware:
Sorry for the delay. We had drastic changes on the file you are modifying in an internal branch. Now we are not using Add_ExternalProjectAdd. Instead of it we are using add_subdirectory and those kind of configuration parameters now are not needed to be passed. These changes are already on master branch. I hope it works for you.

dgoel:
Thanks. I'll close the PR then.

PR #222, opened 2018-05-29
make RTPSDomain a singleton

cvilas:
[rtps_multiparticipant_test.zip](https://github.com/eProsima/Fast-RTPS/files/2048714/rtps_multiparticipant_test.zip)

The attached example program highlights an issue where createRTPSReader() or createRTPSWriter() fails *sometimes* when trying to create a reader and writer attached to different participants in the same process. Could you please investigate? The example program creates one reader and one writer and runs them in separate threads. Each endpoint is attached to a unique participant. To reproduce the problem, just run the executable a couple of times.  A segfault will occur due to create...() returning null. 

The pull request converts RTPSDomain into a singleton, and protects access to calls via a mutex. This seems to fix the issue.

### DIFF
--- a/include/fastrtps/rtps/RTPSDomain.h
+++ b/include/fastrtps/rtps/RTPSDomain.h
@@ -27,6 +27,7 @@
 
 #include <set>
 #include <mutex>
+#include <atomic>
 
 namespace eprosima{
 namespace fastrtps{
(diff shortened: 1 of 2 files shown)


cvilas:
Hi @MiguelCompany. Seems to solve my issue. Thank you!

PR #224, opened 2018-06-08
Fix potential nullptr dereferenced issue in XMLElementParser.cpp

testkit:
elem should be checked before be referenced.


richiware:
Thanks

PR #229, opened 2018-06-19
Updates for Android NDK 17.

esteve:
This updates the support for Android by enabling move semantics for Asio, adding support for `std::string_view` for C++14 and fixing linking errors. This PR also ensures that TinyXML2 and Asio are embedded is the `THIRDPARTY` flag, even if they're found in the system.


esteve:
@richiware this one has all the pieces that the other PR was missing.

esteve:
Here's two CI build with these changes:

- Linux https://travis-ci.org/esteve/ros2_java/jobs/393989674
- Android https://travis-ci.org/esteve/ros2_java/jobs/393989673

CMakeLists.txt
```diff
@@ -119,6 +119,17 @@ else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic -Wextra -Wno-unknown-pragmas -std=c++11")
 endif()
 
+add_definitions(
```
richiware: This block is already set in `src/cpp/CMakeLists.txt`. Is there some other piece failing besides the main library?

CMakeLists.txt
```diff
@@ -128,6 +139,14 @@ if("${CMAKE_CXX_COMPILER_LAUNCHER}" STREQUAL "ccache" AND
     add_compile_options(-fdiagnostics-color=always)
 endif()
 
+if(NOT WIN32)
+  if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
richiware: Can you change this block's location to https://github.com/eProsima/Fast-RTPS/blob/master/CMakeLists.txt#L119?

cmake/modules/FindAsio.cmake
```diff
@@ -2,7 +2,7 @@
 # ASIO_INCLUDE_DIR
 
 find_package(Asio CONFIG QUIET)
-if(Asio_FOUND)
+if(Asio_FOUND AND NOT THIRDPARTY)
```
richiware: If Asio was found in the system, why not use it? is there a reason to prefer the thirdparty asio?

cmake/modules/FindTinyXML2.cmake
```diff
@@ -5,15 +5,15 @@
 option(TINYXML2_FROM_SOURCE "Integrate TinyXML2 source code inside Fast RTPS" OFF)
 
 find_package(TinyXML2 CONFIG QUIET)
-if(TinyXML2_FOUND)
+if(TinyXML2_FOUND AND NOT THIRDPARTY)
```
richiware: If TinyXML2 was found in the system, why not use it? is there a reason to prefer the thirdparty TinyXML2?

cmake/modules/FindAsio.cmake
esteve: This ensures that the version bundled with Fast-RTPS will always work, independent from what's installed (e.g. if the user has a buggy/incompatible version of Asio installed).

cmake/modules/FindTinyXML2.cmake
esteve: The binaries in the system will be of a different architecture than the target when crosscompiling. Besides, it's easier to ship it all in one Fast-RTPS binary than figuring out the location of the system TinyXML2 binaries. So, if I want to crosscompile and later on compile Fast-RTPS natively later on, I have to uninstall and reinstall TinyXML2 every time I want to switch between native and crosscompiling.

esteve:
@richiware 1.10.8 https://github.com/chriskohlhoff/asio/releases/tag/asio-1-10-8, but it should work with 1.12.0 as well.

CMakeLists.txt
esteve: Done in https://github.com/eProsima/Fast-RTPS/pull/229/commits/4db96d431a914bf8712df542a41ced3a9324a781

CMakeLists.txt
esteve: Done in https://github.com/eProsima/Fast-RTPS/pull/229/commits/19f5566173aff200c0e8733f74cc1583fa174490

richiware: In https://github.com/eProsima/Fast-RTPS/blob/master/src/cpp/CMakeLists.txt#L213 you should add the c++14 check.

cmake/modules/FindAsio.cmake
richiware: But if `find_package` is called, it will set some variables, and if you have `THIRDPARTY=ON` also will set some variables. It should be a problem there, shouldn't?

richiware:
I agree with @mikaelarguedas . I prefer maintain the current version (I think it is 1.11) or update it. Not downgrade it.

CMakeLists.txt
esteve: I had already removed that in 19f5566

cmake/modules/FindAsio.cmake
esteve: I haven't had any issue whatsoever with that, to be honest. It's been working for me with both asio installed or not.

esteve:
@richiware 1.11 was never released, Fast-RTPS was tracking a random commit that I used at the time when I ported Fast-RTPS to C++11

@mikaelarguedas I've switched to 1.12.0, which is the latest tagged release of asio on GitHub, 1.12.1 doesn't seem to be tagged https://github.com/chriskohlhoff/asio/issues/309

cmake/modules/FindAsio.cmake
richiware: In the past ROS2 had problems with this macro. @mikaelarguedas do you mind check this change doesn't break anything for you?

richiware:
This PR also runs succesfully into our CI. Do you mind to squash all commits in one?

esteve:
@richiware done, let me know if there's anything else.

mikaelarguedas:
> mikaelarguedas I've switched to 1.12.0, which is the latest tagged release of asio on GitHub, 1.12.1 doesn't seem to be tagged chriskohlhoff/asio#309

:+1: that's the version we use on Windows and Mac :+1:

CMakeLists.txt
```diff
@@ -117,6 +117,22 @@ if(MSVC OR MSVC_IDE)
     endif()
 else()
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic -Wextra -Wno-unknown-pragmas -std=c++11")
+    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
mikaelarguedas: `CMAKE_COMPILER_IS_GNUCXX` ?

cmake/modules/FindAsio.cmake
mikaelarguedas: hmmm I dont remember this issue. Looking at the change I don't see a reason it wouldnt work as we never clone recursively and never pass THIRDPARTY when building. I'll give it a shot tomorrow though to make sure but it seems fine to me

CMakeLists.txt
esteve: CMake now recommends using `CMAKE_CXX_COMPILER_ID` instead https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILER_IS_GNUCXX.html

mikaelarguedas: oh good to know, thanks :+1:

richiware:
Thanks for the contribution @esteve

CMakeLists.txt
richiware: @esteve I'm rebasing certain code over your PR and I see this block was introduced again. I didn't check it again after your rebase. Could it be an error during your rebase? Can I remove this block?

esteve: That's strange. Yeah, just remove it, it should work without it.

esteve:
@richiware thanks for the quick response :-)

Issue #150, opened 2017-09-28
Interface white list does not restrict the input channels or the multicast group membership


deng02:
The UDPv4Transport::mInterfaceListWhiteList does not restrict what interfaces are used in the input channels nor when multicast group membership is performed.  The case is same for UDPv6 but for brevity I'm just going to focus on v4.

The comments in the UDPv4Transport class say:
 >    - It is possible to provide a white list at construction, which limits the interfaces the transport 
 >       will ever be able to interact with. If left empty, all interfaces are allowed.
 >
>     - Opening an input channel by passing a locator will open a socket listening on the given port on 
>       every whitelisted interface, and join the multicast channel specified by the locator address.

However when an input channel is created, the socket binds to any 0.0.0.0 regardless of what the white list says: https://github.com/eProsima/Fast-RTPS/blob/master/src/cpp/transport/UDPv4Transport.cpp#L455

Also if the input channel is a multicast address, it joins the group on all local IPs regardless of what the whitelist says: https://github.com/eProsima/Fast-RTPS/blob/master/src/cpp/transport/UDPv4Transport.cpp#L210

Ideally the interface white list should be fixed to be used in these scenarios so that it truly limits the interfaces the UDPv4Transport is configured to deal with on both.

MiguelCompany:
This should have been fixed by #285

PR #231, opened 2018-06-21
Use the interface whitelist to bind multicast input sockets and filte…

guillaumeautran:
…r locators.

The UDP descriptors have been modified to support flags that
allow the interface white list to be applied to output sockets,
multicast input sockets and/or locators.  By default the white
list is only applied to output sockets.

If applied to multicast input sockets, rather than binding to IP address
any (0.0.0.0), the input socket will bind to specific IPs.

If applied to locators, then the default unicast locator list
will be filtered against the allowed interface IPs.  It is not
applied in the case where the user has specified a default list.

The XML profile has been modified tosupport configuring the
interface white list via the user transport tag.

Issue: #190


MiguelCompany:
> Issue: #190

I guess you refer to issue #150, as this is basically a rebase of #153.

We are now working on a refactor of the transport layer of Fast-RTPS, and will then take this into consideration

guillaumeautran:
@MiguelCompany Yes that's correct. 
Do you happen to have timeline and details wrt the refactor of the transport layer? 
I can help if needed (I am currently merging the IPv4 and IPv6 classes which are 90% identical).

MiguelCompany:
>I am currently merging the IPv4 and IPv6 classes which are 90% identical

That is one of the things we are doing. We are also adding support for TCP.
Regarding timeline, we expect to have it ready and published on github next week.

guillaumeautran:
Sounds good. I'll hold until the new transport layer comes online. We'll see then what needs to be done to incorporate this functionality (which is basically adding the ability to quarantine DDS traffic to a specific interface/subnet) .

guillaumeautran:
@MiguelCompany any update with this? How is the refactor of the transport layer work going?

richiware:
We are reviewing the refactoring. There have been a lot of changes and we have to be carefully and test it very well.  
Thanks for your patience.
↪ 2018-07-19 jue 21:26, Guillaume Autran <notifications@github.com>:

guillaumeautran:
@richiware ping...

richiware:
This PR depends on #251

richiware:
@guillaumeautran Finally TCP transport was merged. We will start looking your PR. Can you rebase it? We can do it for you.

guillaumeautran:
@richiware That's great! 

I won't have the bandwidth to rebase this work until late this week, and since it impacts the transport layer, there is likely some re-implementation needed... 
If you have the bandwidth to help, that would be much appreciated. :)

JuanCarlos-eProsima:
Hi @guillaumeautran,

After rebasing with master I have detected a mismatch in the transport layer, because we made a change to improve the performance and we removed the additional thread for asio and this change has a conflict with the current implementation of the whitelists branch. The main issue is the transformation of the mInputSockets to a map of vectors. 

It seems that is not necessary because the transports will check if there is any socket created before opening a second one, so this vector only will have the first socket. 

Is it ok for you to revert this attribute to a map of sockets ?

JuanCarlos-eProsima:
Hi @guillaumeautran 

Do you need the extra boolean attributes to filter if the whitelist is going to be applied for input, output and locators?  Or, could be enough to check the whitelist on these checks always?

JuanCarlos-eProsima:
@guillaumeautran  friendly ping...

guillaumeautran:
Hi @JuanCarlos-eProsima Sorry for the delay. missed this altogether. 
The purpose of this change is to be able to whitelist specific interface or interfaces from the XML config. 
The mInputSockets was turned into a vector because there may be more than one interface whitelisted. For example, you computer has eth0, wlan0, eth1 and vpn0 network interface. You should be able to whitelist only wlan0 and vpn0 interface and receive/send DDS traffic only on these two interface while the others are completely ignored. 

Reverting mInputSocket to a basic type means the socket needs to be bound to all interfaces and the filtering of traffic must be done some other way.

guillaumeautran:
Regarding your second comment, Our primary usecase is for binding traffic to a subset of interfaces on a machine. The filtering really should be applied to all (input, output as well as locator).

guillaumeautran:
Thanks @JuanCarlos-eProsima , I was suspecting this would need to be re-implemented. My schedule is clearing up next week. I can help more actively. But hopefully, you understand the intent of the changes here.

PR #233, opened 2018-06-22
add thirdparty directory to CMAKE_PREFIX_PATH only if THIRDPARTY optio…

mikaelarguedas:
…n is specified

Otherwise it will use the local clone instead of the system version of the dependency.

Without the patch:
```
-- Found asio: /tmp/fastrtps_test/src/eProsima/Fast-RTPS/thirdparty/asio/asio/include
```

With this patch:
```
-- Found asio: /usr/include 
```

@richiware FYI


dirk-thomas:
@richiware Friendly ping.

PR #241, opened 2018-07-10
Error serialized payload has not been declared in PubSubTypes.cxx

Muneeb0693:
I am trying to communicate between two laptop using helloworld pubsub techniques but in the end I am facing error when I have to make file which says 
**In file included from mztzpePubSubTypes.cxx:26:0:
mztzpePubSubTypes.h:42:29: error: ‘SerializedPayload_t’ has not been declared**

kindly resolve this issue
capture


MiguelCompany:
Why are you trying to merge from master on feature/fastrtpsgen?
Why don't you use the code in master?

The code generated by fastrtpsgen on the master branch is including the full namespace path on the types `SerializedPayload_t` and `InstanceHandle_t`

It seems you are using an old version of fastrtpsgen with the new version of fastrtps

On a side note: you shouldn't open a PR for things like this, unless you are fixing the problem with it. You should open an issue instead.

PR #252, opened 2018-09-04
Making 'max' tag optional

### LINKED ISSUE #245: [XMLPARSER] Parser fails on id_range if 'min' tag present but not 'max' tag
My understanding [reading this XSD](https://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_governance.xsd) is that if the min tag is specified, the max tag is optional (`minOccurs="0"`):
```xml
<xs:complexType name="DomainIdRange">
 <!--  DDSSEC-130  -->
  <xs:choice>
    <xs:sequence>
      <xs:element name="min" type="DomainId"/>
      <xs:element name="max" type="DomainId" minOccurs="0"/>
    </xs:sequence>
    <xs:element name="max" type="DomainId"/>
  </xs:choice>
</xs:complexType>
```

But if I provide a governance file with a domain rule like this
```xml
            <domains>
              <id_range>
                <min>0</min>
              </id_range>
            </domains>
```

Fast-RTPS fails to validate the artifacts with the following error:
```
[XMLPARSER Error] Expected max tag. Line  -> Function parse_domain_id_set
[SECURITY Error] Error validating the local participant permissions. (Malformed governance file file:///<PATH>/governance.p7s (/root/bouncy_ws/src/eProsima/Fast-RTPS/src/cpp/security/accesscontrol/Permissions.cpp:437)) -> Function init
```

Is this because the max tag will become mandatory in a future version of DDS-Security ? or is it something that should be fixed in the parser?

MiguelCompany:
This fixes #245 making domains->id_range->max optional on the governance file

### DIFF
--- a/src/cpp/security/accesscontrol/CommonParser.cpp
+++ b/src/cpp/security/accesscontrol/CommonParser.cpp
@@ -88,33 +88,29 @@ bool eprosima::fastrtps::rtps::security::parse_domain_id_set(tinyxml2::XMLElemen
 
                     if(returned_value && (subnode = subnode->NextSiblingElement()) != nullptr)
                     {
+                        // "Max" parameter is optional
                         if(strcmp(subnode->Name(), Max_str) == 0)
                         {
                             uint32_t max_domain_id = 0;
-
                             if(tinyxml2::XMLError::XML_SUCCESS == subnode->QueryUnsignedText(&max_domain_id))
                             {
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Thanks @MiguelCompany @richiware !

It looks like the case where only the `max` tag is specified is still failing, I ticketed the follow-up at https://github.com/eProsima/Fast-RTPS/issues/255

PR #258, opened 2018-09-11
Refs #3376. Fixed set of WriteParams with volatile durability.

richiware:
After fixed in eProsima/Fast-RTPS@7b2c5b9 a incorrect behavior (removing samples) when a publisher's durability is set to volatile, the old code for getting WriteParams with this configuration (volatile durability) stops working. This PR updates this code to work properly.

This error was detected in ros2/rmw_fastrtps#226.

### DIFF
--- a/include/fastrtps/rtps/history/WriterHistory.h
+++ b/include/fastrtps/rtps/history/WriterHistory.h
@@ -27,6 +27,7 @@ namespace fastrtps{
 namespace rtps {
 
 class RTPSWriter;
+class WriteParams;
 
 /**
  * Class WriterHistory, container of the different CacheChanges of a writer
@@ -55,6 +56,9 @@ class WriterHistory : public History
      * @return True if added.
      */
     RTPS_DllAPI bool add_change(CacheChange_t* a_change);
+
+    RTPS_DllAPI bool add_change(CacheChange_t* a_change, WriteParams &wparams);
+
     /**
      * Remove a specific change from the history.
      * @param a_change Pointer to the CacheChange_t.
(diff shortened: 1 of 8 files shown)


include/fastrtps/rtps/history/WriterHistory.h
```diff
@@ -55,6 +56,9 @@ class WriterHistory : public History
      * @return True if added.
      */
     RTPS_DllAPI bool add_change(CacheChange_t* a_change);
+
+    RTPS_DllAPI bool add_change(CacheChange_t* a_change, WriteParams &wparams);
```
MiguelCompany: This function should be doxygen'd

clalancette:
This seems to fix the problem reported in https://github.com/ros2/rmw_fastrtps/issues/226 .  Thanks!

include/fastrtps/rtps/history/WriterHistory.h
richiware: Done

test/blackbox/BlackboxTests.cpp
```diff
@@ -5269,6 +5269,23 @@ BLACKBOXTEST(BlackBox, AsyncVolatileKeepAllPubReliableSubNonReliableHelloWorld)
     reader.block_for_at_least(2);
 }
 
+// Regression test of Refs #3376, github ros2/rmw_fastrtps #226
+BLACKBOXTEST(BlackBox, ReqRepVolatileHelloworldRequesterCheckWriteParams)
+{
+    ReqRepAsReliableHelloWorldRequester requester;
+    ReqRepAsReliableHelloWorldReplier replier;
+
+    requester.durability_kind(eprosima::fastrtps::VOLATILE_DURABILITY_QOS).init();
```
MiguelCompany: I think the replier should also be set to volatile for this test to work

richiware: The replier is not needed in this test. I removed it.

MiguelCompany: LGTM

PR #267, opened 2018-10-04
Fix localhost

clalancette:
We were running into a situation where Fast-RTPS would not pass any data when all network interfaces were down except for localhost.  It turns out that Fast-RTPS was ignoring localhost as a valid communications interface in some circumstances.  These fixes things so that Fast-RTPS will now consider localhost
as a communications interface, allowing this to work.

I've run the basic tests against this patchset, and they seem to pass.  I also ran this patchset as part of the ROS2 CI ([linux](https://ci.ros2.org/job/ci_linux/5248), [windows](https://ci.ros2.org/job/ci_windows/5221), [osx](https://ci.ros2.org/job/ci_osx/4362)), and things seem to be happy there as well.  I don't know what other knock-on effects this might have, so thoughts/review on that would be appreciated.  Finally, I'm pretty sure there is a similar problem in UDPv6, but I'm much less familiar with IPv6 and I don't have time to debug it right now.

### DIFF
--- a/src/cpp/utils/IPFinder.cpp
+++ b/src/cpp/utils/IPFinder.cpp
@@ -83,7 +83,6 @@ bool IPFinder::getIPs(std::vector<info_IP>* vec_name, bool return_loopback)
 
                 if (family == AF_INET || family == AF_INET6) //IP4
                 {
-                    //printf("\t%s ",  family == AF_INET ? "IPv4":"IPv6");
                     memset(buf, 0, BUFSIZ);
                     getnameinfo(ua->Address.lpSockaddr, ua->Address.iSockaddrLength, buf, sizeof(buf), NULL, 0, NI_NUMERICHOST);
                     info_IP info;
@@ -132,7 +131,7 @@ bool IPFinder::getIPs(std::vector<info_IP>* vec_name, bool return_loopback)
 
     if (getifaddrs(&ifaddr) == -1) {
         perror("getifaddrs");
-        exit(EXIT_FAILURE);
+	return false;
     }
 
     for (ifa = ifaddr; ifa != NULL; ifa = ifa->ifa_next)
@@ -149,7 +148,7 @@ bool IPFinder::getIPs(std::vector<info_IP>* vec_name, bool return_loopback)
             if (s != 0) {
                 printf("getnameinfo() failed: %s\n", gai_strerror(s));
                 freeifaddrs(ifaddr);
-                exit(EXIT_FAILURE);
+		return false;
             }
             info_IP info;
             info.type = IP4;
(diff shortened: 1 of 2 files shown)


src/cpp/utils/IPFinder.cpp
```diff
@@ -83,7 +83,6 @@ bool IPFinder::getIPs(std::vector<info_IP>* vec_name, bool return_loopback)
 
                 if (family == AF_INET || family == AF_INET6) //IP4
                 {
-                    //printf("\t%s ",  family == AF_INET ? "IPv4":"IPv6");
```
dirk-thomas: Why should the patch remove this comment?

src/cpp/utils/IPFinder.cpp
```diff
@@ -132,7 +131,7 @@ bool IPFinder::getIPs(std::vector<info_IP>* vec_name, bool return_loopback)
 
     if (getifaddrs(&ifaddr) == -1) {
         perror("getifaddrs");
-        exit(EXIT_FAILURE);
+	return false;
```
dirk-thomas: The indentation doesn't seem to align and also uses tabs instead of spaces.

Same below.

src/cpp/utils/IPFinder.cpp
MiguelCompany: This would produce undesired console output, so I don't mind it being removed.

src/cpp/utils/IPFinder.cpp
clalancette: Yep, good call.  Fixed now in 8e75a59.

dirk-thomas:
@richiware Friendly ping.

richiware:
@miguelcompany and me were reviewing the PR, but we think the solution  is not ideal. With this PR the locator 127.0.0.1 
is sent to remote endpoints through discovery. I'm working on a better solution in branch 
[hotfix/#3559](https://github.com/eProsima/Fast-RTPS/tree/hotfix/%233559). I will be informing you.
↪ 2018-10-23 mar 00:14, Dirk Thomas <notifications@github.com>:

clalancette:
> @MiguelCompany and me were reviewing the PR, but we think the solution is not ideal. With this PR the locator 127.0.0.1 is sent to remote endpoints through discovery. I'm working on a better solution in branch [hotfix/#3559](https://github.com/eProsima/Fast-RTPS/tree/hotfix/%233559). I will be informing you.

Great, thanks for the update @richiware

MiguelCompany:
Hi @clalancette and @dirk-thomas 

I've seen a better way to address the case when running on a machine with all the network interfaces down except localhost. You can see the code on #303. I have tested the HelloWorldExample on Ubuntu 18 and Windows and the publisher and subscriber match and communicate correctly.

When that PR is merged it will close this one, so I wanted to inform you so you can check it solves the problem originally posted here.

clalancette:
@MiguelCompany I just got a chance to test out PR #303 (really, I tested the develop branch), and things seem to work great there!  It actually fixes the problem better than this PR, since I no longer even need to setup lo for MULTICAST.  Thank you so much for the fixes.  I'm going to close this PR out now.

PR #269, opened 2018-10-08
Refs #3468 Added methods to ease dds-security adaptation.

LuisGP:
Added some methods to easy Time_t (and Duration_t) class management:
[] Constructor from long double seconds.
[] to_ns method returns stored time as nanoseconds
[] >, >=, + and - operators.


include/fastrtps/rtps/common/Time_t.h
```diff
@@ -127,11 +153,52 @@ static inline bool operator<=(const Time_t& t1,const Time_t& t2)
...
+			return false;
+	}
+}
+
 inline std::ostream& operator<<(std::ostream& output,const Time_t& t)
 {
 	return output << t.seconds<<"."<<t.fraction;
 }
 
+static inline Time_t operator+(const Time_t &ta, const Time_t &tb)
```
MiguelCompany: This should be doxygen'd

include/fastrtps/rtps/common/Time_t.h
```diff
@@ -127,11 +153,52 @@ static inline bool operator<=(const Time_t& t1,const Time_t& t2)
...
+{
+	Time_t result(ta.seconds + tb.seconds, ta.fraction + tb.fraction);
+	if (result.fraction < ta.fraction) // Overflow is detected by any of them
+	{
+		++result.seconds;
+	}
+	return result;
+}
+
+static inline Time_t operator-(const Time_t &ta, const Time_t &tb)
```
MiguelCompany: This should be doxygen'd

include/fastrtps/rtps/common/Time_t.h
```diff
@@ -50,7 +50,22 @@ struct RTPS_DllAPI Time_t{
...
+		fraction = (sec - seconds) * 4294967296;
+	}
+
+	/**
+	 *  Returns stored time as nanoseconds
+	 */
+	inline int64_t to_ns()
+	{
+		int64_t nano = seconds * 1000000000;
+		nano += (fraction  * 1000000000) / 4294967296; // 1 fraction = 1/(2^32) seconds
```
MiguelCompany: This will use 32 bit arithmetic. Please use ULL literals to all constants to force 64-bit arithmetic

include/fastrtps/rtps/common/Time_t.h
```diff
@@ -50,7 +50,22 @@ struct RTPS_DllAPI Time_t{
 		seconds = sec;
 		fraction = frac;
 	}
-	
+
+	Time_t(long double sec)
+	{
+		seconds = static_cast<uint32_t>(sec);
+		fraction = (sec - seconds) * 4294967296;
```
MiguelCompany: I think this can produce a warning on some systems, as the result of the calculation is a long double. So I think a cast is required here.

LuisGP:
Done, pong!

PR #271, opened 2018-10-09
Refs #3443. Added equality operator for attributes.

julianbermudez:
According with the `XRCE` standard, entities attributes must be compared in order to use the `REUSE` flag.
For this reason `operator==` has been implemented for all attributes classes, such as `ParticipantAttributes`, `PublisherAttributes` and so on.


MiguelCompany:
This was already addressed as part of #251 and #276

PR #283, opened 2018-10-25
Refs #3571 Added xml configuration for wait tcp negotiation.

LuisGP:
Added parameter to configure wait for negotiation tcp transport descriptor attribute


LuisGP:
Dupped.

PR #286, opened 2018-10-29
Decreased default participant announcement

richiware:
To avoid failing randomly some tests, I decreased the default participant announcement.


JuanCarlos-eProsima:
LGTM

PR #285, opened 2018-10-26
Input sockets whitelist

JuanCarlos-eProsima:
[Refs #3549](https://eprosima.easyredmine.com/issues/3549) Implement a new feature to apply the interface whitelist to the input sockets in UDP and TCP transports

PR #292, opened 2018-11-02
Fix deadlock on TCP

MiguelCompany:
This PR fixes an error introduced by #285. When the socket is closed, the receive call returns 0 (i.e. 0 bytes were read) and the thread keeps looping because the channel has not been closed.


JuanCarlos-eProsima:
LGTM

PR #297, opened 2018-11-05
Refs #3641 Fixed localhost in TCP. 

LuisGP:
- Improved locators shrink.
- Improved mutex usage on SocketConnected method.
- TransactionId is now accessed by mutex.

### DIFF
--- a/src/cpp/rtps/reader/StatefulReader.cpp
+++ b/src/cpp/rtps/reader/StatefulReader.cpp
@@ -77,6 +77,9 @@ bool StatefulReader::matched_writer_add(RemoteWriterAttributes& wdata)
 
     getRTPSParticipant()->createSenderResources(wdata.endpoint.remoteLocatorList, false);
 
+
+    wdata.endpoint.unicastLocatorList =
+        mp_RTPSParticipant->network_factory().ShrinkLocatorLists({wdata.endpoint.unicastLocatorList});
     WriterProxy* wp = new WriterProxy(wdata, this);
 
     wp->mp_initialAcknack->restart_timer();
--- a/src/cpp/transport/TCPTransportInterface.cpp
+++ b/src/cpp/transport/TCPTransportInterface.cpp
@@ -1036,6 +1036,7 @@ LocatorList_t TCPTransportInterface::ShrinkLocatorLists(const std::vector<Locato
                     Locator_t loopbackLocator;
                     FillLocalIp(loopbackLocator);
                     IPLocator::setPhysicalPort(loopbackLocator, IPLocator::getPhysicalPort(*it));
+                    IPLocator::setLogicalPort(loopbackLocator, IPLocator::getLogicalPort(*it));
                     pendingUnicast.push_back(loopbackLocator);
                     break;
                 }
@@ -1091,13 +1092,15 @@ void TCPTransportInterface::SocketAccepted(TCPAcceptor* acceptor, const asio::er
         }
         else
         {
-            logError(RTPC, "Incomming connection from unknown Acceptor: " << IPLocator::getPhysicalPort(acceptor->mLocator));
+            logError(RTPC, "Incomming connection from unknown Acceptor: "
+                << IPLocator::getPhysicalPort(acceptor->mLocator));
             return;
         }
     }
     else
     {
         logInfo(RTCP, " Accepting connection failed (error: " << error.message() << ")");
+        eClock::my_sleep(200); // Wait a little to accept again.
     }
 
     if (error.value() != eSocketErrorCodes::eConnectionAborted) // Operation Aborted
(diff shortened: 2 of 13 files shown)


src/cpp/rtps/reader/StatefulReader.cpp
```diff
@@ -77,6 +77,9 @@ bool StatefulReader::matched_writer_add(RemoteWriterAttributes& wdata)
 
     getRTPSParticipant()->createSenderResources(wdata.endpoint.remoteLocatorList, false);
 
+
+    wdata.endpoint.unicastLocatorList =
+        mp_RTPSParticipant->network_factory().ShrinkLocatorLists({wdata.endpoint.unicastLocatorList});
```
MiguelCompany: Should we also create sender resources for unicast locators?

Should we make a similar change on StatelessReader?

src/cpp/transport/TCPTransportInterface.cpp
```diff
@@ -1087,13 +1092,15 @@ void TCPTransportInterface::SocketAccepted(TCPAcceptor* acceptor, const asio::er
...
-            logError(RTPC, "Incomming connection from unknown Acceptor: " << IPLocator::getPhysicalPort(acceptor->mLocator));
+            logError(RTPC, "Incomming connection from unknown Acceptor: "
+                << IPLocator::getPhysicalPort(acceptor->mLocator));
             return;
         }
     }
     else
     {
         logInfo(RTCP, " Accepting connection failed (error: " << error.message() << ")");
+        eClock::my_sleep(200); // Wait a little to accept again.
```
MiguelCompany: Could you explain the reason for this sleep?

src/cpp/transport/TCPTransportInterface.cpp
```diff
@@ -1142,6 +1151,10 @@ void TCPTransportInterface::SocketConnected(Locator_t locator, const asio::error
         }
         else
         {
+            if (error.value() == 111)
```
MiguelCompany: Please use a constant for this. I think it corresponds to asio::error::basic_errors::connection_refused

src/cpp/transport/TCPTransportInterface.cpp
```diff
@@ -1142,6 +1151,10 @@ void TCPTransportInterface::SocketConnected(Locator_t locator, const asio::error
...
+            {
+                eClock::my_sleep(200); // Wait a little before try again.
```
MiguelCompany: Also explain the reason for this sleep.

src/cpp/transport/tcp/RTCPMessageManager.cpp
```diff
@@ -187,6 +187,7 @@ void RTCPMessageManager::fillHeaders(TCPCPMKind kind, const TCPTransactionId &tr
     //logInfo(RTCP, "Send (CRC= " << header.crc << ")");
 
     // LOG
+    /*
```
MiguelCompany: Why have you removed these logs?

src/cpp/transport/tcp/RTCPMessageManager.cpp
```diff
@@ -579,7 +577,7 @@ bool RTCPMessageManager::processRTCPMessage(TCPChannelResource *pChannelResource
     {
     case BIND_CONNECTION_REQUEST:
     {
-        logInfo(RTCP_SEQ, "Receive [BIND_CONNECTION_REQUEST] Seq: " << controlHeader.transactionId);
+        //logInfo(RTCP_SEQ, "Receive [BIND_CONNECTION_REQUEST] Seq: " << controlHeader.transactionId);
```
MiguelCompany: Why have you removed all the logs?

src/cpp/rtps/reader/StatefulReader.cpp
LuisGP: If I remember correctly StatelessReader doesn't create SenderResources. I make the minimal possible changes to this part of code.

src/cpp/transport/TCPTransportInterface.cpp
LuisGP: Sometimes, the acceptor was being stuck on error accepting incoming connections until I included the sleep. It is possible that the error wasn't related to the wait, but I think it doesn't hurt.

src/cpp/transport/TCPTransportInterface.cpp
LuisGP: Changed.

src/cpp/transport/TCPTransportInterface.cpp
LuisGP: Explained in the next commit.

src/cpp/transport/tcp/RTCPMessageManager.cpp
LuisGP: RTCP_SEQ logs were added with debug purposes and in its current state can introduce some noise.

src/cpp/transport/tcp/RTCPMessageManager.cpp
LuisGP: RTCP_SEQ logs were added with debug purposes and in its current state can introduce some noise.

src/cpp/rtps/reader/StatefulReader.cpp
MiguelCompany: And I still think that sender resources should be created for the shrinked unicast locator list

src/cpp/transport/TCPTransportInterface.cpp
MiguelCompany: If I understand correctly, this is to prevent an infinite loop of connect->refuse->connect->refuse calls.

src/cpp/rtps/reader/StatefulReader.cpp
LuisGP: I had a conversation with @richiware the last week with these questions, and the conclusion was, leave as it is. Maybe you both can argument again and see the consequences. But please, share the conclusions!

src/cpp/transport/TCPTransportInterface.cpp
LuisGP: Exactly, once the OS exhausts all out possible file descriptors it will fail forever.

src/cpp/rtps/reader/StatefulReader.cpp
MiguelCompany: Ok. We will leave it as it is.

PR #298, opened 2018-11-06
Log FileConsumer [3709]

LuisGP:
- New FileConsumer
- Log now allows to empty its consumers
- Updated XML parser to allow log configuration.


include/fastrtps/log/FileConsumer.h
```diff
@@ -0,0 +1,49 @@
...
+#include <fastrtps/log/Log.h>
+
+#include <fstream>
+
+namespace eprosima {
+namespace fastrtps {
+
+class FileConsumer : public LogConsumer {
+public:
+    RTPS_DllAPI FileConsumer();
```
JuanCarlos-eProsima: Please, review this file to add DOXYGEN comments

include/fastrtps/xmlparser/XMLParser.h
```diff
@@ -128,7 +128,7 @@ class XMLParser
     RTPS_DllAPI static XMLP_ret parseXMLProfiles(tinyxml2::XMLElement& profiles, up_base_node_t& root);
     RTPS_DllAPI static XMLP_ret parseProfiles(tinyxml2::XMLElement* p_root, BaseNode& profilesNode);
     RTPS_DllAPI static XMLP_ret parseRoot(tinyxml2::XMLElement* p_root, BaseNode& rootNode);
-
+    RTPS_DllAPI static XMLP_ret parseLogConfig(tinyxml2::XMLElement* p_root);
```
JuanCarlos-eProsima: Please add the documentation in doxygen format

include/fastrtps/xmlparser/XMLParser.h
```diff
@@ -139,6 +139,8 @@ class XMLParser
     RTPS_DllAPI static XMLP_ret parseXMLCommonTransportData(tinyxml2::XMLElement* p_root, sp_transport_t p_transport);
     RTPS_DllAPI static XMLP_ret parseXMLCommonTCPTransportData(tinyxml2::XMLElement* p_root, sp_transport_t p_transport);
 
+    RTPS_DllAPI static XMLP_ret parseXMLConsumer(tinyxml2::XMLElement& consumer);
```
JuanCarlos-eProsima: Please add the documentation in doxygen format

src/cpp/log/FileConsumer.cpp
```diff
@@ -0,0 +1,77 @@
...
+   PrintHeader(entry);
+   mFile << entry.message;
+   PrintContext(entry);
+   mFile << std::endl;
+   mFile.flush();
+}
+
+void FileConsumer::PrintHeader(const Log::Entry& entry)
+{
+    auto now = std::chrono::system_clock::now();
```
richiware: This code is common with Stdconsumer. Cannot be in the parent (LogConsumer)?

src/cpp/xmlparser/XMLParser.cpp
```diff
@@ -102,6 +114,10 @@ XMLP_ret XMLParser::parseXML(tinyxml2::XMLDocument& xmlDoc, up_base_node_t& root
                 {
                     ret = parseXMLTopicData(node, *root);
                 }
+                else if (strcmp(tag, LOG) == 0)
+                {
+                    ret = parseLogConfig(node);
+                }
```
richiware: Can you add an `else`? It has to print and return an error.

src/cpp/xmlparser/XMLParser.cpp
```diff
@@ -1462,6 +1478,160 @@ XMLP_ret XMLParser::parseDynamicTypes(tinyxml2::XMLElement* p_root)
...
+    */
+
+    XMLP_ret ret = XMLP_ret::XML_OK;
+    tinyxml2::XMLElement *p_aux0 = nullptr;
+    p_aux0 = p_root->FirstChildElement(LOG);
+    tinyxml2::XMLElement* p_element = nullptr;
+
+    if (p_aux0 != nullptr)
+    {
+        p_element = p_aux0->FirstChildElement(USE_DEFAULT);
```
richiware: Too much splitting between `p_aux0` and `p_root`. Join all code getting the XMLElement in an auxiliary variable and using it. Also be more restrictive. Right now this code permits any other not support tag.

src/cpp/xmlparser/XMLParser.cpp
```diff
@@ -1462,6 +1478,160 @@ XMLP_ret XMLParser::parseDynamicTypes(tinyxml2::XMLElement* p_root)
...
+                    property = property->NextSiblingElement(PROPERTY);
+                }
+
+                std::unique_ptr<FileConsumer> consumer(new FileConsumer(outputFile, append));
+                Log::RegisterConsumer(std::move(consumer));
+            }
+        }
+        else
+        {
+            logError(XMLParser, "Unknown log consumer class: " << classStr);
```
richiware: Also return an error when an error occur in all this code. Use a variable to store the returned value instead of a lot of `return` sentences.

test/unittest/logging/CMakeLists.txt
```diff
@@ -39,5 +49,76 @@ if(NOT ((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER))
                 )
         endif()
         add_gtest(LogTests SOURCES ${LOGTESTS_SOURCE})
+
+        set(LOGFILETESTS_SOURCE
+            ${PROJECT_SOURCE_DIR}/src/cpp/xmlparser/XMLProfileManager.cpp
```
richiware: Too much code files to an unit test. Maybe if you separate th logging XML part to a separate CPP, we don't have to depend one dynamic type and so on..

richiware: Another alternative is the xml testing part go to the unit testing of xmlparser.

src/cpp/log/FileConsumer.cpp
JuanCarlos-eProsima: I have merged the timestamp block into a common method in the parent class

src/cpp/xmlparser/XMLParser.cpp
JuanCarlos-eProsima: Added

src/cpp/xmlparser/XMLParser.cpp
JuanCarlos-eProsima: I have rewritten this section to validate all the tags and I have removed the split version unifying the management.

src/cpp/xmlparser/XMLParser.cpp
JuanCarlos-eProsima: Updated following your suggestions

test/unittest/logging/CMakeLists.txt
JuanCarlos-eProsima: I have split the tests between the log management and log parsing, moving the second group to XMLProfileTests

src/cpp/log/FileConsumer.cpp
```diff
@@ -0,0 +1,72 @@
...
+   PrintContext(entry);
+   mFile << std::endl;
+   mFile.flush();
+}
+
+void FileConsumer::PrintHeader(const Log::Entry& entry)
+{
+    PrintTimestamp(mFile);
+
+    switch (entry.kind)
```
richiware: This code also can be joined.

test/unittest/logging/CMakeLists.txt
```diff
@@ -39,5 +39,37 @@ if(NOT ((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER))
                 )
         endif()
         add_gtest(LogTests SOURCES ${LOGTESTS_SOURCE})
+
+        set(LOGFILETESTS_SOURCE
```
richiware: Split in `LOGFILETESTS_TEST_SOURCE` adding only `LogFileTests.cpp` and `LOGFILETESTS_SOURCE` will add `LOGFILETESTS_TEST_SOURCE` and the rest of the sources.
Then `add_gtest` will receive only `LOGFILETESTS_TEST_SOURCE`.

For more info see `test/blackbox/CMakeLists.txt:100`

test/unittest/logging/CMakeLists.txt
```diff
@@ -39,5 +39,37 @@ if(NOT ((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER))
...
+        target_include_directories(LogFileTests PRIVATE ${GTEST_INCLUDE_DIRS}
+            ${PROJECT_SOURCE_DIR}/include ${PROJECT_BINARY_DIR}/include)
+        target_link_libraries(LogFileTests ${GTEST_LIBRARIES} ${MOCKS} )
+        if(MSVC OR MSVC_IDE)
+            target_link_libraries(LogFileTests ${PRIVACY}
+                iphlpapi Shlwapi
+                )
+        endif()
+        target_link_libraries(LogFileTests ${PRIVACY} fastcdr)
+        if(TINYXML2_LIBRARY)
```
richiware: Instead of a conditional IF, use a cmake generator expresion inside `target_link_libraries`.

test/unittest/logging/CMakeLists.txt
```diff
@@ -39,5 +39,37 @@ if(NOT ((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER))
...
+
+        include_directories(${TINYXML2_INCLUDE_DIR})
+
+        add_executable(LogFileTests ${LOGFILETESTS_SOURCE})
+        target_compile_definitions(LogFileTests PRIVATE FASTRTPS_NO_LIB)
+        target_include_directories(LogFileTests PRIVATE ${GTEST_INCLUDE_DIRS}
+            ${PROJECT_SOURCE_DIR}/include ${PROJECT_BINARY_DIR}/include)
+        target_link_libraries(LogFileTests ${GTEST_LIBRARIES} ${MOCKS} )
+        if(MSVC OR MSVC_IDE)
+            target_link_libraries(LogFileTests ${PRIVACY}
```
richiware: Instead of a conditional IF, use a cmake generator expresion inside `target_link_libraries`.

test/unittest/xmlparser/XMLProfileParserTests.cpp
```diff
@@ -523,10 +536,118 @@ TEST_F(XMLProfileParserTests, XMLParserSecurity)
 
 #endif
 
+TEST_F(XMLProfileParserTests, file_xml_consumer_append)
+{
+    // First remove previous executions file
+    std::remove("test1.log");
```
richiware: I think these tests doesn't have to check the behavior of Logger. For this purpose we have its unit tests. These test should check the XML parser do the right thing. Thus, check the consumers or logger configuration was loaded correctly.

test/unittest/xmlparser/CMakeLists.txt
```diff
@@ -173,7 +184,7 @@ if(NOT ((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER))
             ${PROJECT_SOURCE_DIR}/include
             ${PROJECT_BINARY_DIR}/include
             )
-            
+
```
richiware: As you will change it in unitest/log/CMakeLists.txt, change all `target_link_libraries` to cmake generator expressions. See the main src/cpp/CMakeLists.txt to see and example.

test/unittest/xmlparser/XMLProfileParserTests.cpp
LuisGP: Changed. Now, these tests only check that the logger is configured correctly.

LuisGP:
A few win and mac warnings, a leak, and a condition race fixed. CI passing.

PR #305, opened 2018-11-08
TCP bugfixing

LuisGP:
- Fixed issue when server rejects a bind connection request due to protocol version causing the client to retry forever.
- Fixed protocol version used.
- Improved Fast-RTPS current protocol version management.
- Added possibility to expand both client and server behaviour on errors in RTCP while negotiation.
- Added "virtual physical port" on tcp clients instead of always report 0. This fixes several clients on the same machine issue.
- Added System.h file to utils folder as helper when retrieving system information.

### DIFF
--- a/src/cpp/utils/System.cpp
+++ b/src/cpp/utils/System.cpp
@@ -1,0 +1,24 @@
+#include <fastrtps/utils/System.h>
+
+#if defined(_WIN32)
+#include <process.h>
+#else
+#include <unistd.h>
+#endif
+
+namespace eprosima {
+namespace fastrtps {
+
+int System::GetPID()
+{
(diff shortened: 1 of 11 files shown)


src/cpp/utils/System.cpp
```diff
@@ -0,0 +1,24 @@
...
+#else
+#include <unistd.h>
+#endif
+
+namespace eprosima {
+namespace fastrtps {
+
+int System::GetPID()
+{
+    #if defined(__cplusplus_winrt)
```
MiguelCompany: Please put conditional directives at the begining of the lines

MiguelCompany:
CI has failed. You have to add the new utils/System.cpp file to the all the unit tests requiring it

PR #320, opened 2018-11-13
Improve udp close mechanism

richiware:


src/cpp/transport/UDPTransportInterface.cpp
```diff
@@ -398,18 +397,20 @@ bool UDPTransportInterface::Receive(UDPChannelResource* pChannelResource, octet*
...
         return false;
     }
 
     try
     {
+        channel->Disable();
+
         ip::udp::socket socket(mService);
         socket.open(GenerateProtocol());
+        socket.bind(GenerateEndpoint("127.0.0.1", 0));
```
MiguelCompany: How does this work on UDPv6? I think we should call a GenerateLocalhostEndpoint method or something similar.

PR #328, opened 2018-11-19
Fix ROS2 test_security failures.

MiguelCompany:
Fixing block counter initialization detected when running ROS2 tests.


JuanCarlos-eProsima:
LGTM

PR #333, opened 2018-11-20
Fixed TCP max msg size and TCPv4 Tests.

LuisGP:
- TCP internal buffers are now configured by maxMessageSize instead of socket sizes.
- If TCP detects a message bigger than its buffer capacity, now drops all the expected bytes that don't fit instead of trying to read a TCP header again.
- TCPv4 tests updated to use default message, and sockets buffer sizes.

### DIFF
--- a/src/cpp/transport/TCPTransportInterface.cpp
+++ b/src/cpp/transport/TCPTransportInterface.cpp
@@ -119,7 +119,7 @@ void TCPTransportInterface::Clean()
         std::unique_lock<std::mutex> scopedLock(mSocketsMapMutex);
         for (auto it = mSocketAcceptors.begin(); it != mSocketAcceptors.end(); ++it)
         {
-            for (auto acceptorIt : it->second)
+            for (TCPAcceptor* acceptorIt : it->second)
             {
                 delete acceptorIt;
             }
@@ -145,7 +145,7 @@ void TCPTransportInterface::Clean()
         mUnboundChannelResources.clear();
     }
 
-    std::for_each(vDeletedSockets.begin(), vDeletedSockets.end(), [this](auto it)
+    std::for_each(vDeletedSockets.begin(), vDeletedSockets.end(), [this](TCPChannelResource* it)
     {
         this->DeleteSocket(it); // Disable all added TCPChannelResources
     });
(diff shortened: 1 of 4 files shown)


src/cpp/transport/TCPTransportInterface.cpp
```diff
@@ -782,21 +780,30 @@ bool TCPTransportInterface::Receive(TCPChannelResource *pChannelResource, octet*
...
-                    //memcpy(&tcp_header, header, TCPHeader::getSize());
                     size_t body_size = tcp_header.length - static_cast<uint32_t>(TCPHeader::getSize());
-                    //logInfo(RTCP_MSG_IN, " Received [TCPHeader] (CRC=" << tcp_header.crc << ")");
 
                     if (body_size > receiveBufferCapacity)
                     {
                         logError(RTCP_MSG_IN, "Size of incoming TCP message is bigger than buffer capacity: "
-                            << static_cast<uint32_t>(body_size) << " vs. " << receiveBufferCapacity << ".");
+                            << static_cast<uint32_t>(body_size) << " vs. " << receiveBufferCapacity << ". " <<
+                            "The full message will be drop.");
```
MiguelCompany:
```suggestion
                            "The full message will be dropped.");
```

test/unittest/transport/TCPv4Tests.cpp
```diff
@@ -139,18 +139,18 @@ TEST_F(TCPv4Tests, send_and_receive_between_ports)
     std::regex filter("RTCP(?!_SEQ)");
     Log::SetCategoryFilter(filter);
     TCPv4TransportDescriptor recvDescriptor;
-    recvDescriptor.maxMessageSize = 5;
-    recvDescriptor.sendBufferSize = 5;
-    recvDescriptor.receiveBufferSize = 5;
+    //recvDescriptor.maxMessageSize = 5;
```
MiguelCompany: Just remove all the commented code.

MiguelCompany:
Please take the chance to also add default values to the in_use field of class ReceiverInUseCV

src/cpp/transport/TCPTransportInterface.cpp
LuisGP: I'll change it along the other requested changes to avoid an additional commit.

PR #334, opened 2018-11-22
Create Environment Variable for default XML file

JuanCarlos-eProsima:
[#Refs 3937](https://eprosima.easyredmine.com/issues/3937) Create the management to check if exists an environment variable to load the DEFAULT_FASTRTPS_PROFILES.xml. The name of the environment variable is **FASTRTPS_DEFAULT_PROFILES_FILE**

### DIFF
--- a/src/cpp/xmlparser/XMLProfileManager.cpp
+++ b/src/cpp/xmlparser/XMLProfileManager.cpp
@@ -15,6 +15,7 @@
 #include <tinyxml2.h>
 #include <fastrtps/xmlparser/XMLProfileManager.h>
 #include <fastrtps/xmlparser/XMLTree.h>
+#include <cstdlib>
 
 namespace eprosima {
 namespace fastrtps {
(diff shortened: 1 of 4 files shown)


src/cpp/xmlparser/XMLProfileManager.cpp
```diff
@@ -104,6 +107,19 @@ void XMLProfileManager::getDefaultTopicAttributes(TopicAttributes& topic_attribu
 
 XMLP_ret XMLProfileManager::loadDefaultXMLFile()
 {
+#ifdef _WIN32
```
MiguelCompany: Please use spaces instead of TABS

src/cpp/xmlparser/XMLProfileManager.cpp
```diff
@@ -104,6 +107,19 @@ void XMLProfileManager::getDefaultTopicAttributes(TopicAttributes& topic_attribu
...
+	char file_path[MAX_PATH];
+	size_t size = MAX_PATH;
+	if (getenv_s(&size, file_path, size, DEFAULT_FASTRTPS_ENV_VARIABLE) == 0 && size > 0)
```
richiware: Remove return value of this function, Anyone is getting it.
Also print errors each time a loadXMLFile fails in this function.

src/cpp/xmlparser/XMLProfileManager.cpp
JuanCarlos-eProsima: Changed

src/cpp/xmlparser/XMLProfileManager.cpp
JuanCarlos-eProsima: I have changed the returned value to void, and the errors are already logged by the `loadXMLFile` function.

PR #346, opened 2018-12-05
Fix user of interface whitelist on linux [4084]

richiware:


src/cpp/transport/UDPTransportInterface.cpp
```diff
@@ -406,25 +414,42 @@ bool UDPTransportInterface::ReleaseInputChannel(const Locator_t& locator, UDPCha
...
+            auto localEndpoint = GenerateLocalEndpoint(localLocator, port);
+            socket.send_to(asio::buffer("EPRORTPSCLOSE", 13), localEndpoint);
+
+            // We then send to the address of the input locator
+            auto destinationEndpoint = GenerateLocalEndpoint(locator, port);
+            socket.send_to(asio::buffer("EPRORTPSCLOSE", 13), destinationEndpoint);
+        }
+        else
+        {
+            auto interface_address = asio::ip::address_v4::from_string(channel->GetInterface());
```
MiguelCompany: This is not IP version agnostic.

richiware: Done

PR #361, opened 2018-12-26
Fixed size strings [4216]

MiguelCompany:
# Fixed size strings

A `template<size_t N> fixed_string` class containing an array of `N + 1` characters. Content is truncated when longer than N characters.

Basic operations implemented:
* Basic construction
* Construct from a `const char*` / `std::string`
* Converters to `const char*` / `std::string`
* Assignment operators from `fixed_string` / `const char*` / `std::string`
* (In)equality comparisons with `fixed_string` / `const char*` / `std::string`

Participant name, topic name and type name changed to be `fixed_string<255>`


dejanpan:
@dejanpan review

PR #366, opened 2019-01-04
Domain Thread Safe [4278]

JuanCarlos-eProsima:
Create a new mutex on the Domain class to make it thread-safe

### DIFF
--- a/include/fastrtps/Domain.h
+++ b/include/fastrtps/Domain.h
@@ -185,7 +185,7 @@ public:
 
 private:
 
-    static std::recursive_mutex m_mutex;
+    static std::mutex m_mutex;
     static std::vector<t_p_Participant> m_participants;
     static bool default_xml_profiles_loaded;
 
--- a/include/fastrtps/Domain.h
+++ b/include/fastrtps/Domain.h
@@ -21,6 +21,7 @@
 #define DOMAIN_H_
 
 #include "attributes/ParticipantAttributes.h"
+#include <mutex>
 
 namespace eprosima{
 namespace fastrtps{
@@ -184,6 +185,7 @@ public:
 
 private:
 
+    static std::recursive_mutex m_mutex;
     static std::vector<t_p_Participant> m_participants;
     static bool default_xml_profiles_loaded;
 
--- a/src/cpp/Domain.cpp
+++ b/src/cpp/Domain.cpp
@@ -84,10 +84,10 @@ void Domain::stopAll()
 
 bool Domain::removeParticipant(Participant* part)
 {
-    std::lock_guard<std::mutex> guard(m_mutex);
-
     if(part!=nullptr)
     {
+        std::lock_guard<std::mutex> guard(m_mutex);
+
         for(auto it = m_participants.begin();it!= m_participants.end();++it)
         {
             if(it->second->getGuid() == part->getGuid())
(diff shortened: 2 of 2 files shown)


src/cpp/Domain.cpp
```diff
@@ -274,6 +290,8 @@ bool Domain::registerType(Participant* part, TopicDataType* type)
 
 bool Domain::registerDynamicType(Participant* part, types::DynamicPubSubType* type)
 {
+    std::lock_guard<std::recursive_mutex> guard(m_mutex);
```
guillaumeautran: Are you sure this lock is necessary? The only two area of concerns would be if the `TypeObjectFactory` and `DynamicTypeBuilderFactory` instances are not thread safe.

src/cpp/Domain.cpp
```diff
@@ -62,6 +63,8 @@ Domain::~Domain()
 
 void Domain::stopAll()
 {
+    std::lock_guard<std::recursive_mutex> guard(m_mutex);
```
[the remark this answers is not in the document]
JuanCarlos-eProsima: You are right, I'm going to create a new scope only for the "remove participants" calls.

src/cpp/Domain.cpp
JuanCarlos-eProsima: We can remove this one because we added the lock inside the `registerType` method that secures the management over participants.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1069/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1039/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1053/)

src/cpp/Domain.cpp
```diff
@@ -314,6 +332,8 @@ bool Domain::registerDynamicType(Participant* part, types::DynamicPubSubType* ty
 
 bool Domain::unregisterType(Participant* part, const char* typeName)
 {
+    std::lock_guard<std::recursive_mutex> guard(m_mutex);
+
```
guillaumeautran: Nitpick: extra blank line.

include/fastrtps/Domain.h
```diff
@@ -184,6 +185,7 @@ class Domain
 
 private:
 
+    static std::recursive_mutex m_mutex;
```
richiware: Why recursive mutex at this level? With a normal mutex is enough.

JuanCarlos-eProsima: You are right. Changed.

guillaumeautran: Not sure if that is correct. `Domain::stopAll()` calls `Domain::removeParticipant()` in a loop. And that second method also acquires the lock. 
Because `Domain::stopAll()`  loops over `m_participants.size()` and that is a shared variable, I think a recursive mutex is required...

JuanCarlos-eProsima: @guillaumeautran we have changed this method in a second push to just remove all the elements of the list of participants without calling `Domain::removeParticipant()`, because these searches are not needed.

guillaumeautran: I see it now. Missed that change before though. 
Yes, in that case, a recursive mutex is not needed. Thanks!

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1093/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1063/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1077/)

PR #401, opened 2019-02-05
Forced c++11.

richiware:


mauropasse:
@richiware  @MiguelCompany 
Forcing C++=11 on https://github.com/eProsima/Fast-DDS/blob/master/CMakeLists.txt#L162
makes the build fail:
```
In file included from /root/sysroot/usr/include/c++/5.2.0/shared_mutex:35:0,
                 from /root/ws/src/eProsima/Fast-DDS/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp:21:
/root/sysroot/usr/include/c++/5.2.0/bits/c++14_warning.h:32:2: error: #error This file requires compiler and library support for the forthcoming ISO C++ 2014 standard. This support is currently experimental, and must be enabled with the -std=c++1y or -std=gnu++1y compiler options.
 #error This file requires compiler and library support for the forthcoming \
  ^
```
Removing the FORCE_CXX line fixes the issue.
Any reason for the existence of that line?

EduPonz:
Hi @mauropasse ,

This is getting solved today in eProsima/Fast-DDS#1545. Shared mutex is not used anymore

PR #431, opened 2019-02-25
Add option for sanitizers enabled so that pthread library can be expl… [4772]

bhatsach:
…icitly linked

Sanitizers like ASan and TSan have their own flavor thread libraries which prevent
cmake from linking the pthread library. Explicity linking this library when SANITIZER_ENABLED
option is on.

Connects to ros2/ci#245

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -277,19 +277,16 @@ elseif(NOT EPROSIMA_INSTALLER)
         set(PRIVACY "PRIVATE")
     endif()
 
+    # Define option to link pthread library when sanitizers are enabled
+    option(SANITIZER_ENABLED "Add pthread library when sanitizers are enabled" OFF)
+
     # Link library to external libraries.
     target_link_libraries(${PROJECT_NAME} ${PRIVACY} fastcdr
-        ${CMAKE_THREAD_LIBS_INIT} ${CMAKE_DL_LIBS}
-        ${TINYXML2_LIBRARY}
+        $<$<BOOL:${SANITIZER_ENABLED}>:pthread$<SEMICOLON>${CMAKE_THREAD_LIBS_INIT}>
+        ${CMAKE_DL_LIBS} ${TINYXML2_LIBRARY}
         $<$<BOOL:${SECURITY}>:OpenSSL::SSL$<SEMICOLON>OpenSSL::Crypto>
         $<$<BOOL:${WIN32}>:iphlpapi$<SEMICOLON>Shlwapi>
         )
-
-    # Define option to link pthread library when sanitizers are enabled
-    option(SANITIZER_ENABLED "Add pthread library when sanitizers are enabled" OFF)
-    if(SANITIZER_ENABLED)
-        target_link_libraries(${PROJECT_NAME} ${PRIVACY} pthread)
-    endif()
 
     if(MSVC OR MSVC_IDE)
         set_target_properties(${PROJECT_NAME} PROPERTIES RELEASE_POSTFIX -${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR})
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -285,6 +285,12 @@ elseif(NOT EPROSIMA_INSTALLER)
         $<$<BOOL:${WIN32}>:iphlpapi$<SEMICOLON>Shlwapi>
         )
 
+    # Define option to link pthread library when sanitizers are enabled
+    option(SANITIZER_ENABLED "Add pthread library when sanitizers are enabled" OFF)
+    if(SANITIZER_ENABLED)
+        target_link_libraries(${PROJECT_NAME} ${PRIVACY} pthread)
+    endif()
+
     if(MSVC OR MSVC_IDE)
         set_target_properties(${PROJECT_NAME} PROPERTIES RELEASE_POSTFIX -${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR})
         set_target_properties(${PROJECT_NAME} PROPERTIES RELWITHDEBINFO_POSTFIX -${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR})
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -282,8 +282,9 @@ elseif(NOT EPROSIMA_INSTALLER)
 
     # Link library to external libraries.
     target_link_libraries(${PROJECT_NAME} ${PRIVACY} fastcdr
-        $<$<BOOL:${SANITIZER_ENABLED}>:pthread$<SEMICOLON>${CMAKE_THREAD_LIBS_INIT}>
-        ${CMAKE_DL_LIBS} ${TINYXML2_LIBRARY}
+        ${CMAKE_THREAD_LIBS_INIT} ${CMAKE_DL_LIBS}
+        $<$<BOOL:${SANITIZER_ENABLED}>:pthread>
+        ${TINYXML2_LIBRARY}
         $<$<BOOL:${SECURITY}>:OpenSSL::SSL$<SEMICOLON>OpenSSL::Crypto>
         $<$<BOOL:${WIN32}>:iphlpapi$<SEMICOLON>Shlwapi>
         )


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1291/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1261/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1274/)

src/cpp/CMakeLists.txt
```diff
@@ -285,6 +285,12 @@ elseif(NOT EPROSIMA_INSTALLER)
         $<$<BOOL:${WIN32}>:iphlpapi$<SEMICOLON>Shlwapi>
         )
 
+    # Define option to link pthread library when sanitizers are enabled
+    option(SANITIZER_ENABLED "Add pthread library when sanitizers are enabled" OFF)
+    if(SANITIZER_ENABLED)
+        target_link_libraries(${PROJECT_NAME} ${PRIVACY} pthread)
```
richiware: Can you use a cmake generator expression to have only one `target_link_libraries` call? As we do in line 284. Thanks

dirk-thomas:
I can't provide a better proposal since I haven't tried it yet but I am not convinced that adding a CMake option for this is the best path forward.

src/cpp/CMakeLists.txt
bhatsach: Done

bhatsach:
> I can't provide a better proposal since I haven't tried it yet but I am not convinced that adding a CMake option for this is the best path forward.

* What are the downsides to adding the cmake option?
* The new cmake option could potentially be used to make all sanitizer specific changes going forward, like skipping some tests or setting up some prerequisites when library path/LD_PRELOAD is changed.
* Do you think unconditionally linking pthread is better?

dirk-thomas:
> What are the downsides to adding the cmake option?

You need to specifically enable that CMake option as well as you need to document its existence.

> Do you think unconditionally linking pthread is better?

Either a package does need `pthread` or it doesn't. If the sanitizer itself brings in that dependency it should be specified by the sanitizer invocation.

bhatsach:
> > What are the downsides to adding the cmake option?
> 
> You need to specifically enable that CMake option as well as you need to document its existence.

Agree that the new CMake option should be made known. However this is true for any cmake option introduced, including the other new CMake arg in this file, INSTALL_EXAMPLES

> 
> > Do you think unconditionally linking pthread is better?
> 
> Either a package does need `pthread` or it doesn't. If the sanitizer itself brings in that dependency it should be specified by the sanitizer invocation.
If we specify the new cmake option during sanitizer invocation, then the package itself can determine how it would like to work with the sanitizer, either by adding pthread or not overriding LD_PRELOAD or so on. Without the cmake option, we will need to add this pthread/LD_PRELOAD logic in the caller (like the address sanitizer job) and that can get messy when dealing with multiple requirements for multiple packages. One way to think about this is that the new cmake option is how the sanitizer brings in or requests for the required dependencies

bhatsach:
> > > What are the downsides to adding the cmake option?
> > 
> > 
> > You need to specifically enable that CMake option as well as you need to document its existence.
> 
> Agree that the new CMake option should be made known. However this is true for any cmake option introduced, including the other new CMake arg in this file, INSTALL_EXAMPLES
> 
> > > Do you think unconditionally linking pthread is better?
> > 
> > 
> > Either a package does need `pthread` or it doesn't. If the sanitizer itself brings in that dependency it should be specified by the sanitizer invocation.

If we specify the new cmake option during sanitizer invocation, then the package itself can determine how it would like to work with the sanitizer, either by adding pthread or not overriding LD_PRELOAD or so on. Without the cmake option, we will need to add this pthread/LD_PRELOAD logic in the caller (like the address sanitizer job) and that can get messy when dealing with multiple requirements for multiple packages. One way to think about this is that the new cmake option is how the sanitizer brings in or requests for the required dependencies

bhatsach:
Additionally having the sanitizer based logic in the package itself might be better to avoid duplication of such logic in the calling packages (CI jobs is just one of the places where this package is built)

dirk-thomas:
> Additionally having the sanitizer based logic in the package itself might be better to avoid duplication of such logic in the calling packages

Isn't implying that potentially each and every package needs this kind of logic the very essence of undesired duplication?

bhatsach:
> > Additionally having the sanitizer based logic in the package itself might be better to avoid duplication of such logic in the calling packages
> 
> Isn't implying that potentially each and every package needs this kind of logic the very essence of undesired duplication?

Please let me know a good way forward here :). What I really meant with the above statement is that instead of having multiple callers have the pthread logic like:
Address sanitizer CI job: colcon build .... -fsanitizer  -pthread...
Thread sanitizer CI job: colcon build .... -fsanitizer  -pthread...
Another place of asan based build invocation: colcon build .... -fsanitizer  -pthread...
we can have the logic to introduce pthread only in the fastrtps package

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1293/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1263/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1276/)

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1295/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1265/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1278/)

bhatsach:
Gentoo wiki pthread suggestion: https://wiki.gentoo.org/wiki/AddressSanitizer/Problems#pthread_linking_issues

nuclearsandwich:
> Gentoo wiki pthread suggestion: https://wiki.gentoo.org/wiki/AddressSanitizer/Problems#pthread_linking_issues

I asked @bhatsach to drop this link while we were discussing the sanitizer jobs we're aiming to run on the ROS 2 CI farm earlier. 

The relevent text from the linked article:

> Some configure scripts test the need of the `-lpthread` compiler flag by testing whether the function `pthread_create()` is available without it. The Address Sanitizer library (libasan) provides this function. However it does not have a full pthread implementation, more advanced pthread functions aren't provided. Therefore the compilation fails.
> 
> A workaround is to add the -lpthread compiler flag manually, the overlay provides a package.env file which does this for some known packages with this problem.

One question I have is if there's a reason not to explicitly add -lpthreads unconditionally on Linux as we've done for [several](https://github.com/osrf/osrf_testing_tools_cpp/pull/21) [packages](https://github.com/ament/googletest/pull/3/) in ROS 2 already. Always adding the flag on Linux removes @dirk-thomas objection to new CMake options and also resolves the issue when using sanitizers that break the CMake heuristic.

bhatsach:
+1 for adding -lpthread unconditionally on Linux. @dirk-thomas and @richiware, what do you think about the same?

richiware:
@MiguelCompany has reminded me that we always do `find_package(Threads REQUIRED)` and later we use `${CMAKE_THREAD_LIBS_INIT}` in `target_link_libraries`. This should add `-lpthread` in Linux. Is it not working for you?

bhatsach:
> @MiguelCompany has reminded me that we always do `find_package(Threads REQUIRED)` and later we use `${CMAKE_THREAD_LIBS_INIT}` in `target_link_libraries`. This should add `-lpthread` in Linux. Is it not working for you?

Ideally `${CMAKE_THREAD_LIBS_INIT}` should add -lpthread, however when sanitizers are linked, CMAKE_THREAD_LIBS_INIT is either empty or points to the sanitizer's own flavor of pthread which is not a complete implementation.

bhatsach:
> > @MiguelCompany has reminded me that we always do `find_package(Threads REQUIRED)` and later we use `${CMAKE_THREAD_LIBS_INIT}` in `target_link_libraries`. This should add `-lpthread` in Linux. Is it not working for you?
> 
> Ideally `${CMAKE_THREAD_LIBS_INIT}` should add -lpthread, however when sanitizers are linked, CMAKE_THREAD_LIBS_INIT either does not contain pthread or points to the sanitizer's own flavor of pthread which is not a complete implementation.

bhatsach:
> > > @MiguelCompany has reminded me that we always do `find_package(Threads REQUIRED)` and later we use `${CMAKE_THREAD_LIBS_INIT}` in `target_link_libraries`. This should add `-lpthread` in Linux. Is it not working for you?
> > 
> > 
> > 

Ideally `${CMAKE_THREAD_LIBS_INIT}` should add -lpthread, however when sanitizers are linked, CMAKE_THREAD_LIBS_INIT either does not contain pthread or points to the sanitizer's own flavor of pthread which is not a complete implementation.

richiware:
I was reviewing `FindThreads.cmake` file that comes with CMake. It is prepared to detect if the thread implementation is inside the libc library. This is the reason `CMAKE_THREAD_LIBS_INIT` is empty when `-fsanitize=address` is set. But I don't know if there will be another situation where the thread implementation is in libc, and setting always pthread can conflict with how `FindThreads.cmake` works.

@dirk-thomas Knowing that using `-fsanitize=address` requires of `-lpthread`. Is not more suitable to add this into the asan mixin? Maybe another packages can benefice of this.

nuclearsandwich:
> It is prepared to detect if the thread implementation is inside the libc library. This is the reason `CMAKE_THREAD_LIBS_INIT` is empty when `-fsanitize=address` is set. But I don't know if there will be another situation where the thread implementation is in libc, and setting always pthread can conflict with how `FindThreads.cmake` works.

There's [a recent change](https://gitlab.kitware.com/cmake/cmake/commit/e9a1ddc594de6e6251bf06d732775dae2cabe4c8) in CMake upstream's FindThreads.cmake that handles this case explicitly by using a different (presumably absent in asan) function for the heuristic. One possibility is vendoring this version of FindThreads.cmake with a guard to print a warning when the required CMake version is at 3.14 or newer as this change is part of the 3.14 release candidate.

bhatsach:
Closing this in favor or https://github.com/eProsima/Fast-RTPS/pull/438 wherein @nuclearsandwich has implemented what was discussed in the above comment. We are vendoring FindThreads.cmake from CMake 3.14-rc2 which fixes the need of pthread library

PR #434, opened 2019-02-26
Made Subscriber history resource limits dependent on history depth [4782]

MiguelCompany:
This makes the behavior on the allocation resource limits consistent on the subscriber and publisher history.

For any entity with a `KEEP_LAST` history policy, its history depth is taken into consideration for the resource limits.

### DIFF
--- a/src/cpp/subscriber/SubscriberHistory.cpp
+++ b/src/cpp/subscriber/SubscriberHistory.cpp
@@ -42,7 +42,17 @@ SubscriberHistory::SubscriberHistory(
         const HistoryQosPolicy& history,
         const ResourceLimitsQosPolicy& resource,
         MemoryManagementPolicy_t mempolicy)
-    : ReaderHistory(HistoryAttributes(mempolicy, payloadMaxSize,resource.allocated_samples,resource.max_samples + 1))
+    : ReaderHistory(HistoryAttributes(mempolicy, payloadMaxSize,
+                history.kind == KEEP_ALL_HISTORY_QOS ?
+                        resource.allocated_samples :
+                        simpl->getAttributes().topic.getTopicKind() == NO_KEY ?
+                            std::min(resource.allocated_samples, history.depth) :
+                            std::min(resource.allocated_samples, history.depth * resource.max_instances),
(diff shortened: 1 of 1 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1298/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1268/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1281/)

src/cpp/subscriber/SubscriberHistory.cpp
```diff
@@ -42,7 +42,17 @@ SubscriberHistory::SubscriberHistory(
...
         MemoryManagementPolicy_t mempolicy)
-    : ReaderHistory(HistoryAttributes(mempolicy, payloadMaxSize,resource.allocated_samples,resource.max_samples + 1))
+    : ReaderHistory(HistoryAttributes(mempolicy, payloadMaxSize,
+                history.kind == KEEP_ALL_HISTORY_QOS ?
+                        resource.allocated_samples :
+                        simpl->getAttributes().topic.getTopicKind() == NO_KEY ?
+                            std::min(resource.allocated_samples, history.depth) :
+                            std::min(resource.allocated_samples, history.depth * resource.max_instances),
+                history.kind == KEEP_ALL_HISTORY_QOS ?
+                        resource.max_samples :
```
richiware: Reader side always needs one more `CacheChange_t`. In case the history is full, before add a new sample it needs one more `CacheChange_t` to deserialize the sample there. Maybe is good idea to add a new blackboxtest that check this scenario as we already have for the writer side.

MiguelCompany: I don't see this on `PublisherHistory`. It seems the extra `CacheChange_t` is already added on `CacheChangePool` [here](https://github.com/eProsima/Fast-RTPS/blob/5e2bebd47536abd82e9e098aa2c68b990ae4327f/src/cpp/rtps/history/CacheChangePool.cpp#L61)

PR #438, opened 2019-02-27
Vendor FindThreads.cmake from CMake 3.14-rc2. [4790]

nuclearsandwich:
In order to resolve linking issues that appear when running Fast-RTPS with ASan [1] due to the asan library providing the symbol whose absense is being used as a test for the need to link pthreads.

This CMake pull request [2] addresses the issue by using a different function as the heuristic and has been accepted upstream as part of the 3.14-rc2 release.

This commit drops the version of FindThreads.cmake from that release into the local CMake modules directory where it should take precedence over the system-provided one.

This file can be removed when the cmake_minimum_required version is 3.14 or newer as the system version should be sufficient.

It would be nice to modify this module to print a warning when it is no longer needed. I tried to find a means to check the minimum version rather than the current cmake version but didn't come up with a way to do so. Issuing a warning a warning on the current cmake version doesn't make sense because it would nag users on rolling systems to make a change incompatible with an older stable system.

[1]: https://wiki.gentoo.org/wiki/AddressSanitizer/Problems#pthread_linking_issues
[2]: https://gitlab.kitware.com/cmake/cmake/merge_requests/2629

Fixes eProsima/Fast-RTPS#430

### DIFF
--- a/cmake/modules/CheckForPthreads.c
+++ b/cmake/modules/CheckForPthreads.c
@@ -1,0 +1,15 @@
+#include <pthread.h>
+
+void* start_routine(void* args)
+{
+  return args;
+}
+
+int main(void)
+{
+  /* This is a compile and link test, no code to actually run things. */
+  pthread_t thread;
+  pthread_create(&thread, 0, start_routine, 0);
+  pthread_join(thread, 0);
+  return 0;
+}
--- a/cmake/modules/FindThreads.cmake
+++ b/cmake/modules/FindThreads.cmake
@@ -1,0 +1,220 @@
+# Copied from https://gitlab.kitware.com/cmake/cmake/raw/v3.14.0-rc2/Modules/FindThreads.cmake to resolve
+# Linking issues when using sanitizers.
+# See https://wiki.gentoo.org/wiki/AddressSanitizer/Problems#pthread_linking_issues and
+# https://gitlab.kitware.com/cmake/cmake/merge_requests/2629 for details.
+# When the minimum version of CMake for this project is at least 3.14 this file can likely be removed.
+
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+
+#[=======================================================================[.rst:
+FindThreads
+-----------
+
+This module determines the thread library of the system.
+
+The following variables are set
+
+::
+
+  CMAKE_THREAD_LIBS_INIT     - the thread library
+  CMAKE_USE_WIN32_THREADS_INIT - using WIN32 threads?
+  CMAKE_USE_PTHREADS_INIT    - are we using pthreads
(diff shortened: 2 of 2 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1302/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1272/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1285/)

nuclearsandwich:
I'm super weirded out by the fact that the mac build passed. Although I suppose it must be because the specific block wasn't used.

The other references to `CMAKE_CURRENT_LIST_DIR` are
```
81:        set(_threads_src ${CMAKE_CURRENT_LIST_DIR}/CheckForPthreads.c)
84:        configure_file(${CMAKE_CURRENT_LIST_DIR}/CheckForPthreads.c "${_threads_src}" COPYONLY)
```

I didn't vendor that file and was able to build locally without it. But I haven't decided if it makes more sense to vendor them or not.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1303/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1273/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1286/)

bhatsach:
I can confirm that this change is working well with the address sanitizer use case and there is no need to pass any pthread flag in the asan mixin now.

richiware:
> The other references to `CMAKE_CURRENT_LIST_DIR` are
> 
> ```
> 81:        set(_threads_src ${CMAKE_CURRENT_LIST_DIR}/CheckForPthreads.c)
> 84:        configure_file(${CMAKE_CURRENT_LIST_DIR}/CheckForPthreads.c "${_threads_src}" COPYONLY)
> ```
> 
> I didn't vendor that file and was able to build locally without it. But I haven't decided if it makes more sense to vendor them or not.

As far as I understand, that piece of block is executed when failed to find pthread symbols in header files and the library libpthread. In this case it tries to check if the compiler has a `-pthread` flag. I don't know which systems or compilers need this, but I think it is better to vendor the file to avoid a future error.

nuclearsandwich:
> I think it is better to vendor the file to avoid a future error.

Added in 7276aeb.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1309/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1279/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1292/)

PR #442, opened 2019-03-08
Remove erroneous two-way fnmatch for DDS Security [4963]

ruffsl:
This fixes:  https://github.com/eProsima/Fast-RTPS/issues/441

The query string and the pattern string should never be swapped when evaluating matching criteria, as this unavoidably results in pattern expansion using user input data, weeking the security permissions enforced, and deviates from the standard set by the default DDS Security Access Control plugin. I.e. remote participants can connect to topics that evaluate to matching expressions for string literal topic names that occur in the remote participant's own granted permission set. E.g. participant `A` may publish to topic name `foo*` even though `A` was only ever granted to `foobar`, allowing for side channel communications while retaining the original permission set.

This subtle issue was uncovered during my development of a formal verification framework for the default DDS Security Access Control plugins and procedurally generated permission files.

### DIFF
--- a/src/cpp/utils/StringMatching.cpp
+++ b/src/cpp/utils/StringMatching.cpp
@@ -97,8 +97,6 @@ bool StringMatching::matchString(const char* str1, const char* str2)
 {
 	if(fnmatch(str1,str2,FNM_NOESCAPE)==0)
 		return true;
-	if(fnmatch(str2,str1,FNM_NOESCAPE)==0)
-		return true;
 	return false;
 }
 


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1319/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1289/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1302/)

MiguelCompany:
Hi @ruffsl 

Some issues with this:
1. I agree that strings should not be swapped when evaluating matching criteria, but the modified method is also used on non-secure partition matching, where the remote endpoint may legally request a partition of `foo*`. That is the reason why the `PubSubAsReliableHelloworldPartitions` blackbox tests are failing. I would add a new method that doesn't swap the strings and use that one when matching criteria. For the name ot the method I would propose `matchStringOneWay`
1. There are three versions of the function you modified: one for `__cplusplus_winrt`, one for `_WIN32` and one for the rest of cases. If a change is made, it should be done for all versions. That is the reason why the tests are passing on Windows.
1. Unit tests should be added for the new method

ruffsl:
> the modified method is also used on non-secure partition matching, where the remote endpoint may legally request a partition of foo*.

I'm still not sure why the middleware should be expected to swap query string and the pattern string and try both permutations, interpret query string as the template and the template as the string data. Shouldn't the call site in the code know which argument should be used for what purpose?

MiguelCompany:
Closing in favor of #1379

PR #445, opened 2019-03-08
Optimize StatefulWriter.check_acked_status [4897]

araccp:
Do a sequence number range check on each change in history rather than an == comparison.

Ran a comparison test, using vtune, on ros1bridge implementation with and without optimized code on a low-latency environment.

With Optimization:
Elapsed time: 386.000s
CPU Time: 459.260s
CPU Time in optimized path: 3.328s

Without Optimization:
Elapsed Time: 403.621s
CPU Time: 591.480
CPU Time in optimized path: 302.246

### DIFF
--- a/src/cpp/rtps/writer/StatefulWriter.cpp
+++ b/src/cpp/rtps/writer/StatefulWriter.cpp
@@ -795,18 +795,24 @@ void StatefulWriter::check_acked_status()
         // Inform of samples acked.
         if(mp_listener != nullptr)
         {
-            for(SequenceNumber_t current_seq = get_seq_num_min(); current_seq <= min_low_mark; ++current_seq)
+            std::vector<CacheChange_t*> all_acked_changes;
+            SequenceNumber_t min_sequence = get_seq_num_min();
+            for (auto cit = mp_history->changesBegin(); cit != mp_history->changesEnd(); ++cit)
             {
-                std::vector<CacheChange_t*>::iterator history_end = mp_history->changesEnd();
-                std::vector<CacheChange_t*>::iterator cit = std::find_if(mp_history->changesBegin(), history_end,
-                    [current_seq](const CacheChange_t* change)
-                    {
-                        return change->sequenceNumber == current_seq;
-                    });
-                if(cit != history_end)
+                // History is order by sequence number
+                // We can break if the analyzed CacheChange is already greater than 'min_low_mark'
+                if ((*cit)->sequenceNumber > min_low_mark)
                 {
-                    mp_listener->onWriterChangeReceivedByAll(this, *cit);
+                    break;
                 }
+                else if ((*cit)->sequenceNumber >= min_sequence)
+                {
+                    all_acked_changes.push_back(*cit);
+                }
+            }
+            for(auto cit = all_acked_changes.begin(); cit != all_acked_changes.end(); ++cit)
+            {
+                mp_listener->onWriterChangeReceivedByAll(this, *cit);
             }
         }
 


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1322/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1292/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1305/)

src/cpp/rtps/writer/StatefulWriter.cpp
```diff
@@ -768,12 +768,12 @@ void StatefulWriter::check_acked_status()
...
-            for(SequenceNumber_t current_seq = get_seq_num_min(); current_seq <= min_low_mark; ++current_seq)
             {
-                for(std::vector<CacheChange_t*>::iterator cit = mp_history->changesBegin();
-                        cit != mp_history->changesEnd(); ++cit)
+                std::lock_guard<std::recursive_mutex> guard(*mp_history->getMutex());
+                SequenceNumber_t min_sequence = get_seq_num_min();
+                for (auto cit = mp_history->changesBegin(); cit != mp_history->changesEnd(); ++cit)
                 {
-                    if((*cit)->sequenceNumber == current_seq)
+                    if ((*cit)->sequenceNumber >= min_sequence || (*cit)->sequenceNumber <= min_low_mark)
```
richiware: Cannot be a `||`. It should be an `&&`. Another optimization suggested by @MiguelCompany: 

```suggestion
                    // History is order by sequence number.
                    // We can break if the analyzed CacheChange is already greater than 'min_low_mark'
                    if ((*cit)->sequenceNumber > min_low_mark)
                    {
                        break;
                    }
                    else
```

src/cpp/rtps/writer/StatefulWriter.cpp
```diff
@@ -768,12 +768,12 @@ void StatefulWriter::check_acked_status()
         if(mp_listener != nullptr)
         {
             std::vector<CacheChange_t*> all_acked_changes;
-            for(SequenceNumber_t current_seq = get_seq_num_min(); current_seq <= min_low_mark; ++current_seq)
             {
-                for(std::vector<CacheChange_t*>::iterator cit = mp_history->changesBegin();
-                        cit != mp_history->changesEnd(); ++cit)
+                std::lock_guard<std::recursive_mutex> guard(*mp_history->getMutex());
```
richiware: This mutex is not needed here. Writer and its history share the same mutex.
```suggestion
```

src/cpp/rtps/writer/StatefulWriter.cpp
araccp: Apologies as I had reversed the logic but forgot the operator.  If the history is sorted by sequence, then the suggested change is great.

MiguelCompany: > If the history is sorted by sequence, then the suggested change is great.

Yes, the history is sorted by sequence. Please apply the suggestions so the CI will pass and this could be merged.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1334/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1304/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1318/)

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1336/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1306/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1320/)

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1342/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1312/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1326/)

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1343/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1313/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1327/)

araccp:
I have made the changes advised but these tests stream appear to be giving rather inconsistent results.  Technically, all three tests have passed but never all at once.  After some time of investigating, it seems to me that I am either unlucky or have exposed bugs; I am believe it's both.  I would really appreciate any additional insight or information regarding these failures that I'm getting.

The following two are not a concern but could be improved in the test code:
1) Linux has one failure - BlackBox_PreallocMem.AsyncPubSubAsReliableData300kb.  My investigation has found that there is a small, but existent, chance to fail when the fastrtps stack works as expected.  Ensuring that the correct number of packets are dropped for the test would take a bit more complexity in test_UDPv4Transport::packet_should_drop; otherwise, adjusting the ASSERT to accept a range instead of the set number is another way to reduce flakiness.
2) Windows BlackBox_PreallocMem.TCP_TLS is failing due to what appears to be an intermittent error when calling asio::ip::tcp::socket::send.  The errno is not printed in RTCPMessageManager::sendMessage() which would be nice. I'm suggesting this is intermittent because it seems to happen on another test I have seen in other windows tests (1322).  The windows tests have since passed, only furthering my assumption.

SEGFAULTS:
1) Windows has a segfault which occurred one 2 tests: BlackBox_ReallocMem.TCPMaxInitialPeer_P0_5_P4.  This is more concerning because there is a semblance of consistency with this failure and the result is a crash.  Looking at the logs, it looks like it's crashing on cleanup as the crash occurs after the main bulk of work is completed. (after discovery) I need a backtrace for the core dump in order to know more.
2) Mac tests, 3 out of 4 result in being "unstable".  All 3 are failing on the same test with different conditions: BlackBox_PreallocMem.TCP_TLS, BlackBox_DynMem.TCP_TLS, BlackBox_ReallocMem.TCP_TLS.  Looks like all these tests fail at the end (upon completion) leading me to suggest that there is an issue with cleanup.  Stacktraces are valuable for looking deeper into these segfaults.

richiware:
Those errors are not related to your changes. We've updated our TCP transport with TLS support and it is a bit unstable. We are working on it. As long as our CI is stable, we will merge your changes. Thanks.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1515/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1479/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1490/)

MiguelCompany:
@araccp We are reducing the number of temporary allocations, and this PR includes the addition of a std::vector.

I have just created PR #515, which should improve performance. Would you mind to test that PR on the same environment specified on the description of this PR?

araccp:
@MiguelCompany I'll have some time today to run a comparison.

MiguelCompany:
Closing this, as #515 is now part of release v1.8.0. 

@araccp Feel free to check the timings with the new release and comment here.

PR #436, opened 2019-02-27
Readers: Resource limits on the number of matched writers [4654]

MiguelCompany:
This PR:
* Adds `matched_publisher_allocation` to `SubscriberAttributes`
* Adds `matched_writers_allocation` to `ReaderAttributes`
* Adds a dependency on [foonathan_memory](https://github.com/foonathan/memory)
* Refactors the collections on `RTPSReader`, `StatelessReader` and `StatefulReader` that depend on the number of matched writers to be resouce limited by `matched_writers_allocation`.
* Refactors `WriterProxy` to allow for its reuse and having a pool on `StatefulReader`
* Refactors the collections on `WriterProxy` that depend on the resource limits of the writer's history to be resource limited.

PR #462, opened 2019-03-25
Remote locators management [4779]

MiguelCompany:
Note: draft pull request as it is built on top of #436

This PR introduces new participant attributes to allow the configuration of:
* The maximum number of unicast locators per remote entity
* The maximum number of multicast locators per remote entity

It has also changed the algorithm to select the remote locators to use when the same data is to be sent to several entities, avoiding the usage of temporary collections and using a new dedicated data structure for the selection algorithm.

# Design
Two different tasks should be performed when managing remote locators:
- Transforming local addresses to localhost
- Selecting the locators to be used when sending the same data to several entities

Right now, these tasks are performed together inside ShrinkLocatorLists, but it makes more sense to perform the first one when receiving the information about the remote entity, leaving the second one to the moment data is sent.

## Receiving locators from remote entities
When information about a participant, reader or writer is received during the discovery process, their locators are added to two `LocatorList_t` fields. `LocatorList_t` holds its values on an internal `std::vector<Locator_t>`, so it may cause allocations whenever its data are modified.

A new struct `RemoteLocatorList` with two `ResourceLimitedVector<Locator_t>` has been created, that is constructed with the maximum number of unicast and multicast locators allowed. Space for that amount of locators will be preallocated when the struct is created, so no reallocations will be triggered later on. This new struct is used inside `XxxProxyData` structures. Before adding a received locator to this struct, method `transform_remote_locator` of `NetworkFactory` is called to perform address translation to localhost if required, and also discard locators for a transport not supported by the participant.

## Adding specific remote locators (initial peers)
Initial peers are only used on built-in PDP writers and cannot be changed after the participant has been created. This means that `fixed_locators_` field can remain a `LocatorList_t` as it will only be modified during the creation of the participant.

## Selecting the locators to use
Summary of algorithm:
- Entries are added/removed with add_entry/remove_entry when matched/unmatched.
- When data is to be sent:
  - A reference to this object is passed to the message group
  - For each submessage:
    - A call to reset is performed
    - A call to enable is performed per desired destination
    - If state_has_changed() returns true:
      - the message group is flushed
      - selection_start is called
      - for each transport:
        - transport_starts is called
        - transport handles the selection state of each entry
        - select may be called
    - Submessage is added to the message group

## Using the selected locators
For optimal traversal of selected locators, the locator selector has a ResourceLimitedVector with the indexes of selected entries, and each entry has a ResourceLimitedVector with the indexes of the selected locators. A traversal method `for_each` receiving a function allows for an action to be called on each selected locator.


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1368/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1336/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1347/)

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1449/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1415/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1426/)

MiguelCompany:
Closing this in favor of #479

PR #495, opened 2019-04-16
Fix for put_time on CentOS 7

### LINKED ISSUE #494: Fast-RTPS Fails to Compile on CentOS 7
It looks like Fast-RTPS fails to compile on CentOS 7 due to a call to `std::put_time()` inside `src/cpp/log/Log.cpp`. This is because `put_time` wasn't implemented until Gcc 5.0 according to [this bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=54354#c9). Unfortunately, CentOS 7 ships w/ Gcc 4.8.5 and upgrading may not be an option for all users.

I realize that `put_time` is part of C++-11, however, CentOS / RHEL are very popular OS's that may be worth supporting. Is there any chance there is an alternative implementation that you will accept to get around this issue?

This particular method appears to be pretty innocuous. I will take a look at an alternative implementation and open a PR. Thanks

jml1795:
This should fix #494 but is mostly provided as an example workaround. Not sure this is the implementation you will want to use. Hope this helps

### DIFF
--- a/src/cpp/log/Log.cpp
+++ b/src/cpp/log/Log.cpp
@@ -203,7 +203,29 @@ void Log::GetTimestamp(std::string &timestamp)
     localtime_s(&timeinfo, &now_c);
     stream << std::put_time(&timeinfo, "%F %T") << "." << std::setw(3) << std::setfill('0') << ms << " ";
 #else
-    stream << std::put_time(localtime(&now_c), "%F %T") << "." << std::setw(3) << std::setfill('0') << ms << " ";
+   struct tm* timeinfo = localtime(&now_c);
+   stream << timeinfo->tm_year+1900 << "-";
+   if (timeinfo->tm_mon+1 < 10) {
+      stream << "0";
+   }
+   stream << timeinfo->tm_mon+1 << "-";
+   if (timeinfo->tm_mday < 10) {
+      stream << "0";
+   }
+   stream << timeinfo->tm_mday << " ";
+   if (timeinfo->tm_hour < 10) {
+      stream << "0";
+   }
+   stream << timeinfo->tm_hour << ":";
+   if (timeinfo->tm_min < 10) {
+      stream << "0";
+   }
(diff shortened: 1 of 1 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1548/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1511/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1522/)

jml1795:
@richiware I force pushed a fix for the extra `<<` (the cause of the failed CI job). I see `2b47e... into a8691...` in the logs which is the commit I force-pushed over. Can you rerun manually?

Unfortunately, I also hit a snag w/ an include of `codecvt`, which also wasn't implemented in the version of gcc that ships w/ CentOS 7. I am working on a patch that uses boost if so configured, but not sure if this is desirable for upstream. Let me know if you would like it. Similar to [this](https://github.com/bakwc/JamSpell/pull/13/files)

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1549/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1512/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1523/)

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1557/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1520/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1531/)

Issue #473, opened 2019-04-01
Fast-RTPS not terminating cleanly [5681]


guillaumeautran:
The problem is expressed with ROS2 and Fast-RTPS. When the last ROS2 node is deleted and the ROS2 system properly terminated, a Fast-RTPS ChannelResource thread lingers behind blocked on a network receive call preventing proper termination of the stack. 

Stacktrace of the surviving thread:
```
Thread 5 (Thread 0x7f0d03fff700 (LWP 5605)):
#0  0x00007f0d3018a94d in recvmsg () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007f0d0f769f9f in asio::detail::socket_ops::recvfrom (ec=..., addrlen=<synthetic pointer>, addr=0x7f0d03ffe270, flags=0, count=1, bufs=0x7f0d03ffe1d0, s=18)
    at /usr/include/asio/detail/impl/socket_ops.ipp:933
#2  asio::detail::socket_ops::sync_recvfrom (ec=..., addrlen=<synthetic pointer>, addr=0x7f0d03ffe270, flags=0, count=1, bufs=0x7f0d03ffe1d0, state=<optimized out>, s=18)
    at /usr/include/asio/detail/impl/socket_ops.ipp:956
#3  asio::detail::reactive_socket_service<asio::ip::udp>::receive_from<asio::mutable_buffers_1> (flags=0, this=<optimized out>, impl=..., impl=..., ec=..., 
    sender_endpoint=..., buffers=...) at /usr/include/asio/detail/reactive_socket_service.hpp:284
#4  asio::datagram_socket_service<asio::ip::udp>::receive_from<asio::mutable_buffers_1> (this=<optimized out>, ec=..., flags=0, sender_endpoint=..., buffers=..., impl=...)
    at /usr/include/asio/datagram_socket_service.hpp:395
#5  asio::basic_datagram_socket<asio::ip::udp, asio::datagram_socket_service<asio::ip::udp> >::receive_from<asio::mutable_buffers_1> (sender_endpoint=..., buffers=..., 
    this=0x1001c) at /usr/include/asio/basic_datagram_socket.hpp:790
#6  eprosima::fastrtps::rtps::UDPTransportInterface::Receive (this=this@entry=0x21b9400, pChannelResource=pChannelResource@entry=0x21b21e0, receiveBuffer=0x2203210 "", 
[...]

PR #499, opened 2019-04-22
WIP: Fast-RTPS not terminating cleanly

guillaumeautran:
Convert the transport UDP receiver into a non-blocking receiver to permit proper thread termination.

issue #473


guillaumeautran:
I don't think this is the correct path for this kind of change. I would much rather see an implementation parallel to the TCP non-blocking implementation.

PR #512, opened 2019-04-30
Use Gradle Wrapper for Java Projects

### LINKED ISSUE #511: Use Gradle Wrapper instead of system Gradle for Java
Most Java project use the Gradle Wrapper in order to perform builds. The wrapper is a small downloader provided by gradle, and helps ensure that the same gradle version is used for all builds. In addition, it makes it so gradle doesn't need to be installed on the target system. It is their recommendation to use it to make builds easier and more repeatable. 

Would you be willing to accept a PR that moves to that? It's just 3 small files, and would help make the Java portion of the build much easier.

ThadHouse:
Makes build deterministic, works across platforms, and does not require manually installing Gradle.

Tested on both Windows and Linux.

Closes #511


LuisGP:
This PR must be transferred to eProsima/Fast-RTPS-Gen.

MiguelCompany:
This was closed in favor of eProsima/Fast-RTPS-Gen#9

PR #524, opened 2019-05-13
Win installer -now OPENSSL's ssl libraries are needed. [5385]

richiware:


MiguelCompany:
Changes not related with code. Merging.

PR #532, opened 2019-05-16
Fix for volatile reliable [5425]

MiguelCompany:
Found an issue (ros2/ros2#703) with late joining volatile readers on volatile writers. 

**Reproduction steps:**
1. Reliable volatile asynchronous publisher is created, and a call to write is performed.
2. A matching reliable volatile subscriber is created.
3. A new call to write is performed.

**Expected result:**
A call to the subscription listener is made shortly after the write call.

**Observed result:**
The call to the listener is made after a time that matches the heartbeat period.

**Explanation:**
1. The subscriber may receive the initial heartbeat before discovering the publisher (thus ignoring it)
2. The publisher has not received the pre-emptive ACKNACK, so it hasn't repeated the initial heartbeat.
3. When the second write call is made, the publisher sends a DATA message for sequence number 2, followed by a GAP to indicate that it will not receive sequence number 1.
4. (bug) After receiving the GAP, the subscriber was not checking it could notify the user of sequence number 2.

### DIFF
--- a/test/blackbox/BlackboxTestsVolatile.cpp
+++ b/test/blackbox/BlackboxTestsVolatile.cpp
@@ -239,3 +239,46 @@ BLACKBOXTEST(BlackBox, ReqRepVolatileHelloworldRequesterCheckWriteParams)
     requester.send(1);
 }
 
+// Test created to check bug #5423, github ros2/ros2 #703
+BLACKBOXTEST(BlackBox, AsyncVolatileSubBetweenPubs)
+{
+	PubSubReader<HelloWorldType> reader(TEST_TOPIC_NAME);
+	PubSubWriter<HelloWorldType> writer(TEST_TOPIC_NAME);
+
+	writer.history_kind(eprosima::fastrtps::KEEP_ALL_HISTORY_QOS).
+		reliability(eprosima::fastrtps::RELIABLE_RELIABILITY_QOS).
+		durability_kind(eprosima::fastrtps::VOLATILE_DURABILITY_QOS).
+		resource_limits_allocated_samples(9).
+		resource_limits_max_samples(9).
+		asynchronously(eprosima::fastrtps::ASYNCHRONOUS_PUBLISH_MODE).
+		heartbeat_period_seconds(3600).
+		init();
+
+	ASSERT_TRUE(writer.isInitialized());
+
+	HelloWorld hello;
+	hello.index(1);
+	hello.message("HelloWorld 1");
+
+	writer.send_sample(hello);
+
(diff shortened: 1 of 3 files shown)


dirk-thomas:
I can confirm that this patch fixes the regression which made the time source tests in ROS 2 fail, see ros2/build_cop#196. Running a full set of builds now to check for any other side effects: https://github.com/ros2/ros2/issues/703#issuecomment-493071206

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1747/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1706/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1718/)

MiguelCompany:
Wrong config on CI. Rebuilding

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1748/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1707/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1719/)

dirk-thomas:
The full builds of ROS 2 look very good. All the regressions seem to be resolved. :+1: to merge this fix.

Thank you for creating it that quickly!

test/blackbox/BlackboxTestsVolatile.cpp
```diff
@@ -239,3 +239,46 @@ BLACKBOXTEST(BlackBox, ReqRepVolatileHelloworldRequesterCheckWriteParams)
...
+	PubSubWriter<HelloWorldType> writer(TEST_TOPIC_NAME);
+
+	writer.history_kind(eprosima::fastrtps::KEEP_ALL_HISTORY_QOS).
+		reliability(eprosima::fastrtps::RELIABLE_RELIABILITY_QOS).
+		durability_kind(eprosima::fastrtps::VOLATILE_DURABILITY_QOS).
+		resource_limits_allocated_samples(9).
+		resource_limits_max_samples(9).
+		asynchronously(eprosima::fastrtps::ASYNCHRONOUS_PUBLISH_MODE).
+		heartbeat_period_seconds(3600).
+		init();
```
raquelalvarezbanos: Cool, I didn't realize we could initialize readers and writers this way in the tests. I'll use it!

PR #548, opened 2019-05-29
Install config files in correct location

hansfn:
According to https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure the config files should be *not* placed under /usr/share (which is what DATA_INSTALL_DIR gives you). Putting it one location under /usr/share makes it impossible to install for multiple libs (or archs).

The provided patch is only tested under Linux and might need extension to work under Windows. See link above for Windows only search path. The Unix search paths should work for Windows too.

PS! This patch uses LIB_INSTALL_DIR/PROJECTNAME/cmake and not LIB_INSTALL_DIR/cmake/PROJECTNAME, which I think is more common on Linux, in the hope that it will work under Windows too.

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -401,7 +401,7 @@ if((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER)
     endif()
 
     install(FILES ${PROJECT_SOURCE_DIR}/cmake/packaging/windows/${PROJECT_NAME}-config.cmake
-        DESTINATION ${DATA_INSTALL_DIR}/${PROJECT_NAME}/cmake
+        DESTINATION ${LIB_INSTALL_DIR}/${PROJECT_NAME}/cmake${MSVCARCH_DIR_EXTENSION_EXT}
         COMPONENT cmake
         )
 elseif(NOT EPROSIMA_INSTALLER)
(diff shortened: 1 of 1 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1786/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1745/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1757/)

JLBuenoLopez-eProsima:
Thanks @hansfn for your contribution. However after reading CMake documentation, I have found no reference to the prohibition to install in `usr/share`. Also, as the fix does not contemplate every supported platform, it is not possible to consider it for merging. Consequently, I am going to proceed and close this PR. Please, feel free to reopen at any time, with a fix working in every platform and more in detail explanation.

hansfn:
Just for the record. You are right that the documentation I pointed to does not support my claim - sorry about that. I guess I misread
```
<prefix>/(lib/<arch>|lib|share)/<name>*/(cmake|CMake)/
```
which definitely allows `/usr/share/${PROJECT_NAME}/cmake`.

I will not reopen, however, for other readers finding this issue: What I tried to solve was building for both 32 and 64 bit and then using `/usr/share/...` breaks.

PR #572, opened 2019-06-24
Stateful writer destruction race condition [5746]

EduPonz:
On destroying the stateful writer, StatefulWriter::perform_nack_supression sometimes gets called after deleting mp_periodicHB, which causes a Segmentation fault


raquelalvarezbanos:
Runnint tests manually:

* On Mac: all tests passed

PR #575, opened 2019-06-26
Test/duration t [5760]

laura-eprosima:


raquelalvarezbanos:
@laura-eprosima Due to some changes that were merged into master recently, this branch no longer builds. Can you have a look and fix it, please?

PR #580, opened 2019-06-29
Update sqlite from 3.22 to 3.28

Vociferix:
See title. The updated amalgamation files were retrieved from https://www.sqlite.org/download.html on June 28, 2019.


MiguelCompany:
We already have sqlite 3.32.3 on master, so this PR is obsolete.

PR #589, opened 2019-07-05
Update also the fascdr installation scripts in thirdparty

MiguelBarro:
commit bc434bd changes in installation scripts are linked to similar ones in thirdparty/fastcdr. fastcdr must reference commit cb513de.


richiware:
CI not run. Not necessary for this PR.

PR #595, opened 2019-07-10
Fixed repos file. [5935]

richiware:


fastrtps.repos
```diff
@@ -3,3 +3,7 @@ repositories:
         type: git
         url: https://github.com/eProsima/Fast-CDR.git
         version: v1.0.10
+    fastrtps:
```
MiguelCompany: Shouldn't we put fastrtps at the beginning?

richiware: At first place, we list dependencies first and at last the product.

PR #598, opened 2019-07-10
Fixed not sending GAP in several cases. [5952]

richiware:


src/cpp/rtps/writer/ReaderProxy.cpp
```diff
@@ -455,6 +455,19 @@ ReaderProxy::ChangeConstIterator ReaderProxy::find_change(const SequenceNumber_t
         : it->getSequenceNumber() == seq_num ? it : end;
 }
 
+bool ReaderProxy::are_there_gaps()
+{
+    bool are_there_gaps = false;
+
+    if (changes_for_reader_.size() > 0 &&
```
MiguelCompany: We could just return the condition of this if

richiware: Done

raquelalvarezbanos:
Merged, as failed test is unrelated

PR #639, opened 2019-07-26
Change unknown writers default behaviour [6121]

### LINKED ISSUE #638: A RTPS reader object will receive any message before matched to a writer [6115]
After creation of a local reader, the reader object is left to accept data from any writer: https://github.com/eProsima/Fast-RTPS/blob/master/src/cpp/rtps/reader/RTPSReader.cpp#L50

When a Data message arrives, the MessageReceiver will try to match it with any reader which wants it: https://github.com/eProsima/Fast-RTPS/blob/master/src/cpp/rtps/messages/MessageReceiver.cpp#L459

But, because the StatelessReader has not been matched to a writer just yet, it will accept just about anything: https://github.com/eProsima/Fast-RTPS/blob/master/src/cpp/rtps/reader/StatelessReader.cpp#L448-L464

This cause a random message to be added to the ReaderHistory for a reader not yet matched with a writer. With the ROS2 stack, this miss-placed message triggers an abort in the typesupport deserialization routine causing the application to terminate abnormally.

Reproduction scenario is to have a bunch of StatelessWriters sending some data at a high frequency. Then, create a new StatelessReader that is supposed to match one and only one of the writers (or none at all). Depending on the timing, the StatelessReader will be receiving message from a random writer during the split second it take for the reader / writer match to happen.

MiguelCompany:
This should fix #638 by changing the default value of m_acceptMessagesFromUnkownWriters on RTPSReader.

### DIFF
--- a/examples/C++/RTPSTest_as_socket/TestReaderSocket.cpp
+++ b/examples/C++/RTPSTest_as_socket/TestReaderSocket.cpp
@@ -69,6 +69,7 @@ bool TestReaderSocket::init(std::string ip, uint32_t port)
 	loc.port = static_cast<uint16_t>(port);
 	ratt.endpoint.multicastLocatorList.push_back(loc);
 	mp_reader = RTPSDomain::createRTPSReader(mp_participant,ratt,mp_history,&m_listener);
+    mp_reader->enableMessagesFromUnkownWriters(true);
 	if(mp_reader == nullptr)
 		return false;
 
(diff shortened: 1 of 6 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/161/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/146/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/160/)

raquelalvarezbanos:
Discovery server tests passed with these changes. Fast-RTPS test failures are the usual

PR #675, opened 2019-08-27
Fixed documentation about installation from sources. [6195]

richiware:


franava:
Under some Linux distro, you might need to "sudo" the following line:
_$ cmake --build . --target install_
since /usr/local is not always accessible from all users

raquelalvarezbanos:
Merging, as this only changes the README.

Issue #684, opened 2019-08-29
Excessive memory usage with lots of reliable topics


guillaumeautran:
On systems with a large number of reliable topics (500 for example), the number of CDRMessage_t structure created at once is directly proportional to the number of reliable topics (in the 1.7.2 release, that number was at least 3 times the number of topics). This does not scale for large number of participant and topics and needs to be improved. 

A viable improvement would be to create a CDRMessage_t pool shared across the system that allocate and services message structures dynamically.

(PR coming for information only as it is based on 1.7.2)

MiguelCompany:
I have addressed this using `thread_local` storage in commit 21b06568b8fbf1c3cbb421b41a63210609b302f4. As there may be systems with limited support for that keyword (namely QNX), it will need to be improved (refactored?) to make it more portable.

guillaumeautran:
Ok. I will try to re-evaluate the memory usage of the stack as soon as this change makes it in master.

MiguelCompany:
Hi @guillaumeautran 

I just closed your PR #685 in favor of my PR #763. I hope it solves your issue.

PR #685, opened 2019-08-29
Deprecate RTPSMessageGroup_t in favor of using the CDRMessagePool ins…

guillaumeautran:
…tead

Remove the early allocation of the CDRMessage_t contained in RTPSMessageGroup_t and replace it with the usage of the CDRMessagePool.
The savings are linearly proportional to the number of Statefull readers / writers. Those use to preallocate CDRMessage_t buffers for the ACKNACK and HEARTBEAT packets resulting in a significant amount of RAM used.
Instead, these will now leverage the CDRMessagePool and allocate a buffer from there on demand only, saving a large amount of memory for lots of statefull endpoints.

Issue: #684


MiguelCompany:
Closing this in favor of #763

PR #719, opened 2019-09-17
Fix history depth and history resources in throughput tests [6357]

EduPonz:
In case KEEP_LAST, set the history depth to at least the burst demand
In case KEEP_ALL, the the topic resource limit max samples to at least the burst demand

### DIFF
--- a/test/performance/ThroughputPublisher.cpp
+++ b/test/performance/ThroughputPublisher.cpp
@@ -38,20 +38,29 @@ using namespace eprosima::fastrtps;
 using namespace eprosima::fastrtps::rtps;
 using namespace eprosima::fastrtps::types;
 
-ThroughputPublisher::DataPubListener::DataPubListener(ThroughputPublisher& up):m_up(up){}
-ThroughputPublisher::DataPubListener::~DataPubListener(){}
+// *******************************************************************************************
+// ************************************ DATA PUB LISTENER ************************************
+// *******************************************************************************************
+ThroughputPublisher::DataPubListener::DataPubListener(ThroughputPublisher& up)
+    :m_up(up)
+{
+}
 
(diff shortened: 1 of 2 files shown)


test/performance/ThroughputPublisher.cpp
```diff
@@ -322,23 +373,42 @@ void ThroughputPublisher::run(uint32_t test_time, uint32_t recovery_time_ms, int
...
             command.m_size = sit->first;
             command.m_demand = *dit;
-            //cout << "SEND COMMAND "<< command.m_command << endl;
+
+            // Check history resources depending on the history kind and demand
+            if (pubAttr.topic.historyQos.kind == KEEP_LAST_HISTORY_QOS)
+            {
+                // Ensure that the history depth is at least the demand
+                if (pubAttr.topic.historyQos.depth < 0 ||
+                    static_cast<uint32_t>(pubAttr.topic.historyQos.depth) < command.m_demand)
```
raquelalvarezbanos: Not sure about this. What will happen in the case that `historyQos.depth` (not sure if this is possible, as we could be checking for this in the code) is negative when we cast to to `uint32_t`? I would re-write the above if as:

`if (pubAttr.topic.historyQos.depth < static_cast<int32_t>(command.m_demand)`

I believe only this condition is enough. The same applies for the `KEEP_ALL` case below.

raquelalvarezbanos: OK, as discussed, that would never happen as the first condition will be checked first.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/432/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/391/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/430/)

PR #723, opened 2019-09-19
Refs #6341 FastCDR version that fixes cmake paths. [6377]

LuisGP:
Changes the version of FastCDR needed by fastrtps.repos to fix generated cmake paths.


richiware:
Not need to run CI

PR #739, opened 2019-09-25
Flag to avoid builtin multicast [6450]

MiguelCompany:
This PR addresses ros2/rmw_fastrtps#315 by allowing the user to disable multicast traffic for builtin endpoints.


MiguelCompany:
Not waiting for Mac, as we will do more testing later

PR #753, opened 2019-10-02
Refs #6500 BUILD_DOCUMENTATION now gets the current version being gen… [6500]

LuisGP:
…erated.


raquelalvarezbanos:
No need to wait for the tests, as this only changes our cmake.

PR #754, opened 2019-10-02
BUILD_DOCUMENTATION version [6502]

LuisGP:
Download the current `PROJECT_VERSION` tag of the documentation instead of `latest`.


raquelalvarezbanos:
Aborted the tests, as this change doesn't have any effect.

PR #755, opened 2019-10-02
BUILD_DOCUMENTATION version - 1.9.x [6503]

LuisGP:
Download the current `PROJECT_VERSION` tag of the documentation instead of `latest`.


raquelalvarezbanos:
No need to wait for the test, as this only changes cmakelists.

PR #810, opened 2019-10-24
shared memory transport, sony primitive information and proposal.

fujitatomoya:
This is the proposal information from Sony about shared memory transport and shared memory only data sharing. including background and our basic design and consideration, that would be better to be shared with.


fujitatomoya:
@adolfomarver @MiguelCompany 

could you take a look at this?

PR #851, opened 2019-11-06
Only call find_package for TinyXML2 if thirdparty options are off

dljsjr:
FindTinyXML2.cmake's current behavior is to always call `find_package` even if the options for `THIRDPARTY` or `TINYXML2_FROM_SOURCE` aren't set. So on systems with a TinyXML2 installation this means that those options are effectively ignored as find_package will pick up the system level install.

This PR does a simple check to only call find_package if those options aren't set without modifying any of the fallback code.


MiguelCompany:
@dljsjr Thanks for the contribution!

PR #889, opened 2019-11-26
Improve sending GAP [1.9.x]

richiware:
Improvement sending only GAP for sequence number greater/equal than history minimum sequence number. In other case, force to send a piggyback heartbeat.


src/cpp/rtps/writer/StatefulWriter.cpp
```diff
@@ -621,7 +622,15 @@ void StatefulWriter::send_any_unsent_changes()
                             }
                             else
                             {
-                                notRelevantChanges.add_sequence_number(seq_num, remoteReader);
+                                if (seq_num >= get_seq_num_min())
```
MiguelCompany: As this is inside a loop, and we have the mutex taken, it may be a good idea to keep the result of `get_seq_num_min` on a local variable

richiware: Done

PR #973, opened 2020-01-23
Fixed error sending more than 256 fragments <1.9.x> [7403]

richiware:
The async thread was not awaken to send the next unsent fragments after sending the first 256. This PR fixes this.

PR #974, opened 2020-01-23
Fixed error sending more than 256 fragments <master> [7404]

richiware:
The async thread was not awaken to send the next unsent fragments after sending the first 256. This PR fixes this.

Refactored as a port of #973

### DIFF
--- a/include/fastdds/rtps/common/CDRMessage_t.h
+++ b/include/fastdds/rtps/common/CDRMessage_t.h
@@ -47,22 +47,12 @@ namespace rtps {
  * @brief Structure CDRMessage_t, contains a serialized message.
  * @ingroup COMMON_MODULE
  */
-struct RTPS_DllAPI CDRMessage_t
+struct RTPS_DllAPI CDRMessage_t final
 {
-    //! Default constructor
+    // TODO(Miguel C): Deprecate when not used in mocks
     CDRMessage_t()
-        : wraps(false)
+        : CDRMessage_t(RTPSMESSAGE_DEFAULT_SIZE)
     {
-        pos = 0;
-        length = 0;
-        buffer = (octet*) malloc(RTPSMESSAGE_DEFAULT_SIZE);
-        max_size = RTPSMESSAGE_DEFAULT_SIZE;
-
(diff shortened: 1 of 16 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1255/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1042/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1252/)

include/fastdds/rtps/common/CDRMessage_t.h
```diff
@@ -196,6 +205,8 @@ struct RTPS_DllAPI CDRMessage_t final
     uint32_t pos;
     //!Max size of the message.
     uint32_t max_size;
+    //!Size allocated on buffer. May be higher than max_size.
+    uint32_t reserved_size;
```
IkerLuengo: I can only see `max_size` and `reserved_size` differing after `reserve()`, and in that case it seems that `max_size` can end up being larger than `reserverd_size`, not the contrary

Even when the `realloc` fails, we set the `max_size` to the attempted size leaving the `reserved_size` with the old (lower) value. 

What is the intention keeping a separate `max_size`?

MiguelCompany: This is about calling `reserve()` first with a big size and then with a shorter size. In that case we dont't want the reallocation, but we want the shorter size being assigned to `max_size`

IkerLuengo: It makes sense

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1489/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1260/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1487/)

PR #944, opened 2020-01-08
Fixing wait_until when time is in the past <1.9.x> [7210]

MiguelCompany:
When `TimedConditionVariable::wait_until` receives a `max_blocking_time` in the past, the calculations lead to a negative number on the timespec struct. This leads to EINVAL being returned from pthread_cond_timed_wait.

PR #1000, opened 2020-02-04
Fix TimedConditionVariable hanging forever when now() > max_blocking_time

ivanpauno:
I'm targetting the PR to `1.9.x` just because it's the branch where I was working on.
The same probably applies to master.

The original code did:
```cpp
std::chrono::nanoseconds nsecs = max_blocking_time - std::chrono::steady_clock::now();
```
overflows when max_blocking_time has already passed.

I've carefully checked that that was the cause of the hang I was experimenting (I printed everything and checked values, it was overflowing).

[Here](https://gist.github.com/ivanpauno/960bc58b1ab506e9ad7f5b8e43a3c3ea) backtraces of such a hung (thread 1 and thread 2 are the problem).

Extra comment:
The return type of `wait_until` when a predicate is not passed should be `std::cv_status`, and not `bool` (see [here](https://en.cppreference.com/w/cpp/thread/condition_variable_any/wait_until)). I'm not fixing that here, and returning `false` when timeout happens (same was happening before as return value of `pthread_cond_timedwait` is not `0` when timout happens).

Bottom line: Why not just using `std::condition_variable_any`? Is because a compiler not supporting it? I see that that's actually being used in not linux systems.

### DIFF
--- a/include/fastrtps/utils/TimedConditionVariable.hpp
+++ b/include/fastrtps/utils/TimedConditionVariable.hpp
@@ -117,7 +117,11 @@ class TimedConditionVariable
                 std::function<bool()> predicate)
         {
             bool ret_value = true;
-            std::chrono::nanoseconds nsecs = max_blocking_time - std::chrono::steady_clock::now();
+            auto now = std::chrono::steady_clock::now();
+            if (now > max_blocking_time) {
+                return predicate();
+            }
+            std::chrono::nanoseconds nsecs = max_blocking_time - now;
             struct timespec max_wait = { 0, 0 };
             clock_gettime(CLOCK_REALTIME, &max_wait);
             nsecs = nsecs + std::chrono::nanoseconds(max_wait.tv_nsec);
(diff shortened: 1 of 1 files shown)


MiguelCompany:
We already have #944 for this. It has a different approach, as it leaves the decission on how to timeout to the native pthread function.

Regarding why we have this, we found out that certain implementations of `std::condition_variable_any` have an `std::mutex` inside that is taken without timeout. This can lead to the call to wait_until to be non-deterministic.

MiguelCompany:
> Regarding why we have this, we found out that certain implementations of `std::condition_variable_any` have an `std::mutex` inside that is taken without timeout. This can lead to the call to wait_until to be non-deterministic.

Now that I think about it, we should only use this when building with `STRICT_REALTIME`

ivanpauno:
Didn't see the other PR, which looks better. Can't that be pushed?
This might be the cause of many hangs we have in ros2 CI (I can't confirm that, as the error I was trying to fix was other).

ivanpauno:
Closing this, as I finally realized my fix didn't solve the issue.
Returning when `now > max_blocking_time` without releasing the mutex is just wrong.

I also tried https://github.com/eProsima/Fast-RTPS/pull/944, it doesn't work because it's not including  `<condition_variable>` when `std::condition_variable_any` is used. Correcting that, it worked perfectly for me.
But if I edit the code to use the custom implementation of `TimedConditionVariable` (with the #944 fix included), it hungs.
As far as I've tried, it seems that passing past point times is still a problem.

I don't have time for looking at the bug, and I'm happy if #944 gets merged as it solves the problem for ROS2 (we don't use `STRICT_REALTIME`). But just for let you know, the custom implementation of `TimedConditionVariable` still have problems.
I can provide detailed steps of how to reproduce the issue if you want. It's not easy to reproduce, as I'm using many custom ros2 branches and I have to run a test more than 100 times to make it hung.

When using `std::condition_variable_any`, it doesn't have, even if I run the problematic test 2000 times.

Issue #1016, opened 2020-02-14
1.9.4 deadlocking between StatefulWriter and PDP [7614]


wtsnyder:
## Expected Behavior
PDP mutex should never be locked before endpoint mutex.

## Current Behavior
FastRTPS can deadlock when StatefulReader::change_received is called because it does not lock the Endpoint mutex but can lock the PDP mutex.  Other code paths lock these mutexes in the opposite order (see helgrind trace below)

## Fix
StatefulReader::change_received should lock the Endpoint mutex like other public functions in StatefulReader

Diff
```
--- a/src/cpp/rtps/reader/StatefulReader.cpp
+++ b/src/cpp/rtps/reader/StatefulReader.cpp
@@ -696,6 +696,7 @@ bool StatefulReader::change_received(
         CacheChange_t* a_change,
         WriterProxy* prox)
 {
+    std::lock_guard<RecursiveTimedMutex> guard(mp_mutex);
     //First look for WriterProxy in case is not provided
     if (prox == nullptr)
     {
```

## System information

- Fast-RTPS version: 1.9.4 (with #996 fix applied)
- OS: Ubuntu 16.04
- Network interfaces: eth0 (192.168.1.x), lo (127.0.0.1)
- ROS2: n/a

## Additional context
- using simple endpoint discovery protocol
- program creates 2 participants within a single application
- new intraprocess delivery appears to be being used

## Additional resources
Valgrind --tool=helgrind capture of deadlock

```
==21147== ----------------------------------------------------------------
==21147== 
==21147== Thread #4: lock order "0x8539C38 before 0x8292400" violated
==21147== 
==21147== Observed (incorrect) order is: acquisition of lock at 0x8292400
==21147==    at 0x4C321BC: ??? (in /usr/lib/valgrind/vgpreload_helgrind-amd64-linux.so)
==21147==    by 0x5EDEC2D: __gthread_mutex_lock(pthread_mutex_t*) (gthr-default.h:748)
==21147==    by 0x5EDECB7: __gthread_recursive_mutex_lock(pthread_mutex_t*) (gthr-default.h:810)
==21147==    by 0x5EE5AFB: std::recursive_mutex::lock() (mutex:176)
==21147==    by 0x5EE7DE9: std::lock_guard<std::recursive_mutex>::lock_guard(std::recursive_mutex&) (mutex:386)
[...]

PR #1017, opened 2020-02-18
Unlock PDP mutex before callbacks [7653]

IkerLuengo:
There is no need for the PDP mutex to be held once the callback is called. This will reduce the likelihood of interlocking during the client's operations.

This relates to #1016

### DIFF
--- a/src/cpp/rtps/builtin/discovery/participant/PDP.cpp
+++ b/src/cpp/rtps/builtin/discovery/participant/PDP.cpp
@@ -557,7 +557,7 @@ bool PDP::removeReaderProxyData(
         const GUID_t& reader_guid)
 {
     logInfo(RTPS_PDP, "Removing reader proxy data " << reader_guid);
-    std::lock_guard<std::recursive_mutex> guardPDP(*this->mp_mutex);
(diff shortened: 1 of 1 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1425/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1198/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1423/)

wtsnyder:
I could be wrong but is this change sufficient to resolve this issue?  

PDP uses a `std::recursive_mutex` so calling `guardPDP.unlock()` will only reduce the level of ownership by 1 if the mutex has been locked previously by this thread.  It may be true that these methods are never called from other methods locking the same mutex (I didn't dig that deep) but in that case a `std::mutex` would be more appropriate.

MiguelCompany:
Closing this as it is not the correct solution. 

@IkerLuengo We should state in the documentation that certain operations should not be called inside the discovery callbacks.

Issue #1042, opened 2020-03-05
Publisher write call blocks or fails when a new subscriber is started later [7833]


aquad:
## Expected Behavior

The publisher's write call should not block excessively or fail when a new subscriber is started.

## Current Behavior

I have a publisher regularly writing e.g. at 100Hz, with reliable QoS, keep-last, depth 1.

Some minutes later, I start a subscriber to it. The publisher's write call then blocks for many milliseconds, sometimes over 100ms, resulting in failure. This only happens once, when the subscriber is started, after which it returns to normal.

## Steps to Reproduce

I have a simple ROS2 test publisher that keeps statistics of how long each publish call takes, and prints out when it's high.
https://github.com/aquad/test_pub_blocking_time

1. Create publisher running at 1000Hz:
   `ros2 run test_pub_blocking_time test_pub_blocking_time --rate 1000`
2. Wait 1 minute.
3. Create subscriber: `ros2 topic hz /test_topic`

With the above wait, it blocks for ~15ms. Waiting longer increases the blocking time.

## System information

- Fast-RTPS version: 1.8.0
- OS: Ubuntu 18.04
- Network interfaces: lo (127.0.0.1)
- ROS2: Dashing

## Diagnosis

Within `PublisherImpl::create_new_change_with_params()`, the mp_writer mutex is try-locked, which blocks, waiting on another thread. If it reaches 100ms, it will fail and return.

It is waiting on the mutex lock at `StatefulWriter::send_any_unsent_changes()`. Most of the time in this function is spent on handling a very large `std::set<SequenceNumber_t>` in notRelevantChanges (e.g.  size over 100000), around `RTPSMessageGroup::add_gap()`,

The set is created at `StatefulWriter::send_any_unsent_changes()`:
```
remoteReader->for_each_unsent_change(max_sequence, unsent_change_process);
```
The remoteReader (ReaderProxy) stores a `SequenceNumber_t changes_low_mark_` member, and it has to iterate from there to max_sequence (can be hundreds of thousands), adding elements to the std::set.  When you create a new subscriber, it results in a new ReaderProxy, with `changes_low_mark_ = 0`, and thus a giant std::set.

[...]

PR #1045, opened 2020-03-06
Simulate initial acknack on ReaderProxy::start() [7838]

EduPonz:
This PR tackles the #1042 issue.

With it, as part of starting a `ReaderProxy`,  a call to `acked_changes_set(SequenceNumber_t())` is performed to simulate an initial acknack from the reader. This has the effect of setting the ReaderProxy's low mark to the writer's first available change, which avoids the writer need to iterate through all the changes 0 to low mark the first time it sends data to the reader.

This is OK to do, since an initial heartbeat is sent to the reader notifying the writer's state right after discovery, which already tells the reader that the first change is low mark, so there is no need of sending any gap, nor requesting changes older than that.

### DIFF
--- a/src/cpp/rtps/writer/ReaderProxy.cpp
+++ b/src/cpp/rtps/writer/ReaderProxy.cpp
@@ -102,6 +102,7 @@ void ReaderProxy::start(
     expects_inline_qos_ = reader_attributes.m_expectsInlineQos;
     is_reliable_ = reader_attributes.m_qos.m_reliability.kind != BEST_EFFORT_RELIABILITY_QOS;
     disable_positive_acks_ = reader_attributes.disable_positive_acks();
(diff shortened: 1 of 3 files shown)


EduPonz:
Comment updated in 
49b6ed1

EduPonz:
> Change the comment on `src/cpp/rtps/writer/ReaderProxy.cpp` (lines 292-293) to reflect that the special case is now used on late joiners too.

Comment updated in
49b6ed1

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1521/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1293/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1519/)

PR #1064, opened 2020-03-12
Fixing unsent_change_added_to_history with no readers [7927]

MiguelCompany:
I've seen some of the liveliness tests passing correctly but after a very long time (more than 50 seconds). I checked that 50 seconds is what the DATA(p) announcement period is set to on those tests. All of them were manual_by_topic and non-intraprocess.

The problem was that liveliness was not asserted when matched_readers_ was empty.

I have also seen that the call to `encrypt_cachechange` was suffering from the same problem on `StatelessWriter`. May have been the reason for the failures on `SimpleCommunicationSecureBestEffort`.

This PR fixes both problems.

### DIFF
--- a/src/cpp/rtps/writer/StatefulWriter.cpp
+++ b/src/cpp/rtps/writer/StatefulWriter.cpp
@@ -209,6 +209,14 @@ void StatefulWriter::unsent_change_added_to_history(
 {
     std::lock_guard<RecursiveTimedMutex> guard(mp_mutex);
 
+    if (liveliness_lease_duration_ < c_TimeInfinite)
+    {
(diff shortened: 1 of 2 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1570/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1340/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1568/)

Issue #1076, opened 2020-03-18
CMake Error: Couldn't compile thirdparty/boost with current configuration [7996]


nbaldy:
On compilation with current master (e4907e61325cec47419424a4ac5ce21c96bd51ab) 3rdParty/Boost errors happen when attempting to run "colcon build" or "cmake .." (tried with both manual and  colcon installation)

Note that 1e403b47d9bfe4dff3455d55ba7638c09919e4f2 builds with warnings (and is what I will be using) - I just wanted to point out this issue. Sorry if I missed a step, I'm still very new at everything

## Expected Behavior
Successful compilation

## Current Behavior
```
The following error chain:
-- Setting build type to 'Release' as none was specified.
-- The C compiler identification is GNU 5.4.0
-- The CXX compiler identification is GNU 5.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring Fast RTPS
-- Version: 1.10.0
-- To change the version modify the file configure.ac
-- Performing Test SUPPORTS_CXX14
-- Performing Test SUPPORTS_CXX14 - Success
-- Performing Test SUPPORTS_CXX1Y
-- Performing Test SUPPORTS_CXX1Y - Success
-- Performing Test SUPPORTS_CXX11
-- Performing Test SUPPORTS_CXX11 - Success
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of unsigned short
-- Check size of unsigned short - done
-- Using unsigned short
-- Check if the system is big endian - little endian
-- fastcdr library found...
-- Found tinyxml2: /usr/lib/x86_64-linux-gnu/libtinyxml2.so  
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_kill
-- Looking for pthread_kill - not found
[...]

PR #1078, opened 2020-03-19
CMake Error: compiling thirdparty/boost in Ubuntu16 <master> [8000]

adolfomarver:
This PR solves the github issue #1076.


MiguelCompany:
Not waiting for tests, as we have checked locally on a docker that #1076 is fixed.

PR #1095, opened 2020-03-24
Remove unnecesary error log msg when opening a SHM port <master> [8030]

adolfomarver:
When trying to open a SHM port (read exclusive), the open_port function logs an error if the port is already opened. This should be avoided because FastRTPS automatically tries another port number until a free port is found. Therefore, errors must not be logged to the user, errors in open_port are reported through exceptions that, in this case, are catched and filtered by the upper layer.


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/1640/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1406/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1636/)

PR #1165, opened 2020-04-21
Change register type return type [8227]

MiguelCompany:


include/fastdds/dds/topic/TypeSupport.hpp
```diff
@@ -84,7 +88,7 @@ class TypeSupport : public std::shared_ptr<fastdds::dds::TopicDataType>
     {
     }
 
-    RTPS_DllAPI virtual bool register_type(
+    RTPS_DllAPI virtual ReturnCode_t register_type(
             DomainParticipant* participant,
             std::string type_name) const;
```
IkerLuengo, line 93: Set empty string as default value for the name. That way it is easier if we want to use the type's name:
`my_type.register_type(my_participant)`
Instead of having to pass `my_type.get_type_name()` or `""` or even `std::string()` too

MiguelCompany: I don't like defalt values. Will add another method instead.

Shall I also change the places where the method is called?

IkerLuengo: OK to the extra method.
I wouldn't bother on tests, but maybe on examples, to show the usage to users.
But I will not complain if they are changed on tests too, so as you wish,

MiguelCompany: done

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2000/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1732/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/1994/)

Issue #1120, opened 2020-04-01
Very slow message delivery in some cases with Fast-RTPS 1.9.x [8914]


clalancette:
## Expected Behavior
There is an example ROS 2 workload at https://gist.github.com/clalancette/cdd5082838c38da94923ed8ba16d1342 .  When compiling and running that workload (exact steps below), we expect that a trigger on one end of the pipeline will result in a pointcloud being published at the other end of the pipeline in < 500 ms.

## Current Behavior
It can take anywhere from 500 ms up to 30 seconds for the pipeline to complete.  Sometimes, the pipeline never completes at all (this happens more as more load is being placed on the machine).

## Steps to Reproduce
1.  Setup machine for ROS 2: https://index.ros.org/doc/ros2/Installation/Foxy/Linux-Development-Setup/
1.  mkdir -p pc_pipe_full_ws/src
1.  cd pc_pipe_full_ws
1.  wget https://gist.githubusercontent.com/clalancette/cdd5082838c38da94923ed8ba16d1342/raw/8670ceb1e201822f0d5d56a52f45982c5c3739b1/pc_pipe_full.repos
1.  vcs import src < pc_pipe_full.repos
1.  colcon build  # This will take a while

Terminal 1:
1.  source pc_pipe_full_ws/install/setup.bash
1.  ros2 launch pc_pipe bringup.launch.xml start_gazebo:=true viz:=true

Terminal 2:
1.  source pc_pipe_full_ws/install/setup.bash
1.  ros2 topic echo --qos-reliability reliable --no-arr /mycamera/points2

Terminal 3:
1.  source pc_pipe_full_ws/install/setup.bash
1.  ros2 topic pub -1 mycamera/image_trigger std_msgs/msg/Empty

When that last command is run, the pipeline is kicked once.  A bunch of timing information is written to /tmp/pc_pipe_times.csv.  If you download https://gist.github.com/clalancette/dfaee6948105bd1f781bbf151dc60a3c, that script does some basic analysis on the output.  You can run it like: `python3 parse_pc_pipe_times.py /tmp/pc_pipe_times.csv`

## System information

- Fast-RTPS version: 1.9.x branch, hash e72b1c3c380affd177bf427072c0fe74b9d874d2
- OS: Ubuntu 18.04
- Network interfaces: lo, eth0 (whatever Fast-RTPS chooses as default)
- ROS2: Foxy pre-release.

## Additional context

The pipeline goes like this:

```
[...]

Issue #1107, opened 2020-03-26
High CPU usage when big message are published on arm32 [8995]


lukicdarkoo:
We compiled ROS2 Eloquent with Fast RTPS middleware implementation on arm32 (Raspberry Pi Zero). In general, it works quite good, but we are experiencing low FPS when we send images locally (publisher and subscriber onboard). We profiled our node and noticed that a lot of images are dropped - the images are acquired and sent (`publisher->publish()`) at 10Hz but the subscriber (`ros2 topic hz image_raw`) receives them at 3 - 4Hz. 

## Expected Behavior
Since both nodes, publisher and subscriber, are running locally I would expect it to easily achieve 10fps. The images are around 1MB (640x480x3).

## Current Behavior
Achieved fps is around 3 - 4 and the CPU is at ~100%

## Steps to Reproduce
1. Publish raw images (640x480) at a high rate
2. Subscribe to the topic on the same device

## System information
- Fast-RTPS version: [tag] release/eloquent/fastrtps/1.9.3-2
- OS: Raspbian Buster
- Network interfaces: lo (127.0.0.1)
- ROS2: Eloquent

I assume that the messages are dropped because the device cannot handle the workload. But since the high CPU usage is caused by Fast RTPS I was wondering if it is possible to reduce it?

EduPonz:
Hi @lukicdarkoo ,

Unfortunately we don't have the hardware to verify your issue, but we did some improvements that can and should affect your case. It'd be nice if you could try with master branch. Furthermore, if both publisher and subscriber are in the same machine, you could benefit from [Fast DDS Shared Memory Transport](https://fast-rtps.docs.eprosima.com/en/latest/fastdds/transport/shared_memory/shared_memory.html). If you have any problems setting it up just let us know and we'll be happy to guide you.

Issue #1062, opened 2020-03-11
Message delays when system under load [7920]


jacobperron:
If we send reasonably large data (~1MB) once every few seconds, there are sometimes delays in reading the messages locally when the machine is under load. I've noticed that these delays come in bursts.

## Expected Behavior

Messages should arrive consistently.

## Current Behavior

Messages sometimes take 30-40 seconds to arrive, and typically these delays affect consecutive messages in bursts.

## Steps to Reproduce

1. Source a ROS 2 installation
1. Compile the modified ROS 2 talker/listener demos [here](https://github.com/ros2/demos/tree/jacob/repro_comms_delay).

        mkdir -p /tmp/repro_ws/src
        cd /tmp/repro_ws/src
        git clone https://github.com/ros2/demos.git -b jacob/repro_comms_delay
        cd ..
        colcon build --packages-up-to demo_nodes_cpp

2. Run the talker

        source /tmp/repro_ws/install/setup.bash
        ros2 run demo_nodes_cpp talker

3. Run the listener

        source /tmp/repro_ws/install/setup.bash
        ros2 run demo_nodes_cpp listener

4. Induce load on the machine

        stress --cpu 32

The ROS publisher will send an image message every 3 seconds, and the listener will output the time difference between the time stamp in the message and the time it received the message.

After some time (about a minute), we start see large delays in the arrival of messages to the listener.

## System information

- Fast-RTPS version: 1.9.x
- OS: Ubuntu 18.04
- Network interfaces: lo (127.0.0.1), there's also an ethernet  interface (enp0s25), but I presume it shouldn't be involved since communication is limited to one host.
- ROS 2: latest (Foxy development)

## Additional context

Sending a message every 1 second makes the problem even worse sometimes messages never arrive (even though we are using reliability set to reliable for the publisher and subscription).

I'm not able to reproduce this delayed behavior with other supported RMW implementations (Connext or CycloneDDS).

MiguelCompany:
@jacobperron We have been able to reproduce this problem. We have also checked that using synchronous publishing the problem does not happen. You can check how to enable synchronous publishing [here](https://github.com/ros2/rmw_fastrtps#advance-usage)

Regarding other RMW implementations, we know that CycloneDDS uses synchronous publishing by default, see ros2/rmw_cyclonedds#89. And, IIRC, Connext detects if both participants are in the same host, and will use shared memory in that case.

dirk-thomas:
[...]

PR #1209, opened 2020-05-11
Writer improvements <2.0.x> [8382]

MiguelCompany:
This PR alleviates #1062, #1107 and #1120 with several improvements on the writers.

* Several loops reduced on fragment management, on b2b528c, 071febe and c1151a8
* Changed algorithm on `StatefulWriter::check_acked_status` from `O(n*log(n))` to `O(n)` on 4340614
* Removed loop on `StatelessWriter::try_remove_change` on adcce30
* Implicit flow controller added to both writers on 3f7fef2 and efb08fc

### DIFF
--- a/src/cpp/rtps/writer/StatelessWriter.cpp
+++ b/src/cpp/rtps/writer/StatelessWriter.cpp
@@ -418,10 +418,14 @@ void StatelessWriter::send_all_unsent_changes()
 {
     //TODO(Mcc) Separate sending for asynchronous writers
 
+    static constexpr uint32_t implicit_flow_controller_size = RTPSMessageGroup::get_max_fragment_payload_size();
+
     NetworkFactory& network = mp_RTPSParticipant->network_factory();
     RTPSMessageGroup group(mp_RTPSParticipant, this, *this);
     bool remote_destinations = locator_selector_.selected_size() > 0 || !fixed_locators_.empty();
     bool bHasListener = mp_listener != nullptr;
+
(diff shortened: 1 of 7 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2098/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1855/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2093/)

src/cpp/rtps/writer/StatelessWriter.cpp
```diff
@@ -53,11 +53,7 @@ static bool add_change_to_rtps_group(
         {
             for (uint32_t frag = 1; frag <= n_fragments; frag++)
             {
-                if (group.add_data_frag(*change, frag, inline_qos))
-                {
-                    reader_change->markFragmentsAsSent(frag);
```
richiware, line 58: Why now is not needed to mark fragment as sent?

MiguelCompany: This method is called when there are no flow controllers, and the whole change is sent (i.e. all fragments). As this is the StatelessWriter, if a fragment is lost, the whole change will be lost, so we don't care to mark each fragment as sent, as we are always sending all of them, and the whole change will be marked as sent.

PR #1222, opened 2020-05-21
Compliant behavior of ENTITY_FACTORY and enable() <2.0.x> [8471]

MiguelCompany:


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2122/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1913/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2117/)

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -1676,31 +1701,44 @@ bool DomainParticipantImpl::can_qos_be_updated(
...
+void DomainParticipantImpl::create_instance_handle(
+        fastrtps::rtps::InstanceHandle_t& handle)
+{
+    using eprosima::fastrtps::rtps::octet;
+
+    ++next_instance_id_;
+    handle = guid_;
+    handle.value[15] = 0x01; // Vendor specific;
+    handle.value[14] = static_cast<octet>(next_instance_id_ & 0xFF);
+    handle.value[13] = static_cast<octet>( (next_instance_id_ >> 8) & 0xFF);
```
IkerLuengo, line 1741:
```suggestion
    handle.value[13] = static_cast<octet>((next_instance_id_ >> 8) & 0xFF);
```

src/cpp/fastdds/publisher/Publisher.cpp
```diff
@@ -61,7 +61,9 @@ ReturnCode_t Publisher::enable()
     }
 
     enable_ = true;
```
IkerLuengo, line 63: This is not needed anymore

src/cpp/fastdds/subscriber/Subscriber.cpp
```diff
@@ -55,7 +55,9 @@ ReturnCode_t Subscriber::enable()
     }
 
     enable_ = true;
```
IkerLuengo, line 57: This is not needed anymore

src/cpp/fastdds/publisher/DataWriter.cpp
```diff
@@ -48,6 +48,24 @@ DataWriter::~DataWriter()
...
+    {
+        return ReturnCode_t::RETCODE_OK;
+    }
+
+    if (false == impl_->get_publisher()->is_enabled())
+    {
+        return ReturnCode_t::RETCODE_PRECONDITION_NOT_MET;
+    }
+
+    enable_ = true;
```
IkerLuengo, line 63: This should not be necessary, we are setting the value after enabling the implementation.

src/cpp/fastdds/subscriber/DataReader.cpp
```diff
@@ -55,6 +55,24 @@ DataReader::~DataReader()
...
+    {
+        return ReturnCode_t::RETCODE_OK;
+    }
+
+    if (false == impl_->get_subscriber()->is_enabled())
+    {
+        return ReturnCode_t::RETCODE_PRECONDITION_NOT_MET;
+    }
+
+    enable_ = true;
```
IkerLuengo, line 70: This should not be necessary, we are setting the value after enabling the implementation.

src/cpp/fastdds/domain/DomainParticipant.cpp
```diff
@@ -59,8 +59,10 @@ ReturnCode_t DomainParticipant::enable()
         return ReturnCode_t::RETCODE_OK;
     }
 
-    Entity::enable();
-    return impl_->enable();
+    enable_ = true;
```
IkerLuengo, line 62: This should not be necessary, we are setting the value after enabling the implementation.

src/cpp/fastdds/publisher/Publisher.cpp
MiguelCompany: Yes, it is. Inside `impl_->enable()`, we may call the `enable()` method on previously created writers, which will fail if the publisher is not enabled.

src/cpp/fastdds/subscriber/Subscriber.cpp
MiguelCompany: Yes, it is. Inside `impl_->enable()`, we may call the `enable()` method on previously created readers, which will fail if the subscriber is not enabled.

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
MiguelCompany: And, most important, on the next line we should shift by 16, not 8!

src/cpp/fastdds/domain/DomainParticipant.cpp
MiguelCompany: Yes, it is. Inside `impl_->enable()`, we may call the `enable()` method on previously created entities, which will fail if the participant is not enabled.

src/cpp/fastdds/domain/DomainParticipantFactory.cpp
```diff
@@ -247,15 +213,13 @@ DomainParticipant* DomainParticipantFactory::create_participant(
 
     if (factory_qos_.entity_factory().autoenable_created_entities)
     {
-        dom_part->enable();
+        if (ReturnCode_t::RETCODE_OK != dom_part->enable())
+        {
+            delete_participant(dom_part);
+            return nullptr;
+        }
```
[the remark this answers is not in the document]
MiguelCompany: Which behavior should we choose? If `autoenable_created_entities` is true, it means that the user wants it enabled.

Should the user check that the returned entity is enabled? Should a `nullptr` be returned?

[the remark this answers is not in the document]
IkerLuengo: According to our conversation, it looks like returning `nullptr` is what agrees with the standard most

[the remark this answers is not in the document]
MiguelCompany: Done on e4cd956

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2123/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1914/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2118/)

PR #1233, opened 2020-05-27
Add versioning documents [8531]

MiguelCompany:


VERSIONING.md
```diff
@@ -0,0 +1,9 @@
+Versioning policy declaration
+=============================
+
+Starting on v2.0.0, the version numbers for this library will follow [Semantic Versioning](https://semver.org/).
```
EduPonz, line 4:
```suggestion
Starting on v2.0.0, the version numbers for this library will adhere to the versioning policies defined by [Semantic Versioning](https://semver.org/).
```

VERSIONING.md
```diff
@@ -0,0 +1,9 @@
...
+
+This means that API breaks should only happen between major version changes.
+If an ABI break is required, it should be done between minor versions and be clearly stated on the release notes.
```
EduPonz, line 7:
```suggestion
If an ABI break is required, it should be done between minor versions, and it should be clearly stated on the release notes.
```

EduPonz, line 7: I'm not sure whether we need these

MiguelCompany: I prefer to be explicit, especially regarding ABI changes, as semver does not seem to be clear.

PR #1239, opened 2020-05-29
ThreadSanitizer: data race on socket close/receive_from.

fujitatomoya:
fix https://github.com/ros2/rmw_fastrtps/issues/391,

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/include/fastdds/rtps/transport/UDPChannelResource.h
+++ b/include/fastdds/rtps/transport/UDPChannelResource.h
@@ -169,6 +169,7 @@ protected:
 private:
 
     TransportReceiverInterface* message_receiver_; //Associated Readers/Writers inside of MessageReceiver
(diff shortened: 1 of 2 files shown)


MiguelCompany:
As you can see [here](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1928/console) all tests are timing out when closing.

I am almost sure that this race is not harmful, so I would prefer a patch that avoids thread sanitizer to complain.

fujitatomoya:
@MiguelCompany 

I am taking this back, since I believe,

- this is false positive.
- seems to be not a problem in fastrtps.
- even if there is data race, fastrtps is not the right place to fix but boost.

PR #1240, opened 2020-05-29
Fix examples files installation <2.0.x> [8566]

adolfomarver:
- Fix segfault in DDSTypeLookupService when no XML file is found.
- Install examples XML and certs files.


MiguelCompany:
As examples are not checked by our CI, and I have checked locally that the changes here work, I am letting this in

PR #1253, opened 2020-06-09
remove link libs rt for Android NDK compiling.

jayhou:
-lrt ld params will lead Android NDK cross-compiling failed. 

Signed-off-by: jayhou


cmake/modules/FindThirdpartyBoost.cmake
```diff
@@ -15,6 +15,8 @@ find_package(Threads REQUIRED)
 
 if(WIN32 OR APPLE)
     set(THIRDPARTY_BOOST_LINK_LIBS ${CMAKE_THREAD_LIBS_INIT})
+elseif(ANDROID)
+    set(THIRDPARTY_BOOST_LINK_LIBS ${CMAKE_THREAD_LIBS_INIT})
```
MiguelCompany, line 19:
```suggestion
if(WIN32 OR APPLE OR ANDROID)
    set(THIRDPARTY_BOOST_LINK_LIBS ${CMAKE_THREAD_LIBS_INIT})
```

PR #1259, opened 2020-06-16
Call listener on incompatible QoS [8707]

IkerLuengo:
- When requested/offered QoS is incompatible with the one offered/requested by the current entity, the corresponding listener is called.
- Also the incompatibleQos status of the DataReader and DataWriter are updated
- Returning from the listener, the total_count_change is reset
- Retrieving the incompatibleQos status with get_requested_incompatible_qos_status and get_offered_incompatible_qos_status resets the total_count_change
- Unit testing for the QoS compatibility logic + incompatibleQos status updates
- Blackbox testing for a series of combinations of incompatibilities and compatibilities, both using listener callbacks and getters.

Note: the current listener calling logic has been maintained. That is, the listener callback is called whenever there is a listener for the entity, regardless of the status filtering mask. Correcting the listener calling logic will be done in a separate task, not only for incompatible QoS, but for all the rest of the callbacks as well.


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2202/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/1991/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2196/)

MiguelCompany:
@IkerLuengo Please solve warnings on Windows and Mac

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -680,6 +680,21 @@ void DataWriterImpl::InnerDataWriterListener::onWriterMatched(
...
+        fastdds::dds::QosPolicyId_t qos_id)
+{
+    OfferedIncompatibleQosStatus& status = data_writer_->update_offered_incompatible_qos(qos_id);
+    if (data_writer_->listener_ != nullptr)
+    {
+        data_writer_->listener_->on_offered_incompatible_qos(data_writer_->user_datawriter_, status);
+        status.total_count_change = 0u;
+    }
+
+    data_writer_->publisher_->publisher_listener_.on_offered_incompatible_qos(data_writer_->user_datawriter_, status);
```
MiguelCompany, line 694: Should this be always called?

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -420,6 +420,20 @@ void DataReaderImpl::InnerDataReaderListener::on_liveliness_changed(
...
+        fastdds::dds::QosPolicyId_t qos_id)
+{
+    RequestedIncompatibleQosStatus& status = data_reader_->update_requested_incompatible_qos(qos_id);
+    if (data_reader_->listener_ != nullptr)
+    {
+        data_reader_->listener_->on_requested_incompatible_qos(data_reader_->user_datareader_, status);
+        status.total_count_change = 0u;
+    }
+
+    data_reader_->subscriber_->subscriber_listener_.on_requested_incompatible_qos(data_reader_->user_datareader_, status);
```
MiguelCompany, line 434: Should this be always called?

test/unittest/rtps/discovery/EdpTests.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+TEST_F(EdpTests, IncompatibleType)
+{
+    set_incompatible_type();
+    EXPECT_FALSE(edp->validMatching(wdata, rdata));
+    EXPECT_TRUE(edp->writer_incompatible_qos_policies(wdata->guid()).none());
+
+    EXPECT_FALSE(edp->validMatching(rdata, wdata));
+    EXPECT_TRUE(edp->reader_incompatible_qos_policies(rdata->guid()).none());
+}
```
MiguelCompany, line 197:
```suggestion
TEST_F(EdpTests, CompleteCompatibility)
{
    check_expectations(true);
}

TEST_F(EdpTests, IncompatibleTopic)
{
    set_incompatible_topic();
    check_expectations(false);
}

TEST_F(EdpTests, IncompatibleTopicKind)
{
    set_incompatible_topic_kind();
    check_expectations(false);
}

TEST_F(EdpTests, IncompatibleType)
{
    set_incompatible_type();
    check_expectations(false);
}
```

test/unittest/rtps/discovery/EdpTests.cpp
```diff
@@ -0,0 +1,476 @@
...
+
+TEST_F(EdpTests, CheckDurabilityCompatibility)
```
MiguelCompany, line 199: On this and all the following tests, I would follow a pattern of:
* Defining a list of tuples with offered value, requested value, expected result. For the policies depending on an enumeration, I would put all possible combinations of the values.
* Traverse the list assigning the values and calling `check_expectations(expected_result, policy_id)`

We can make an exception on the partitions one, which I would put before this test.

test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -640,20 +640,19 @@ TEST_P(PubSubBasic, PubSubIncompatibleQosListeners)
     EXPECT_EQ(incompatible_durability_reader.last_incompatible_qos(), eprosima::fastdds::dds::DURABILITY_QOS_POLICY_ID);
     EXPECT_FALSE(incompatible_durability_reader.is_matched());
 
-    // A compatible Reader on the same Topic
-    // Should match and never call incompatible QoS callbacks
+    // A compatible Reader on another Topic
+    // Should not match but never call incompatible QoS callbacks
     // Total count of incompatible QoS occurrences and latest incompatible QoS ID stay the same
-    PubSubReader<HelloWorldType> compatible_reader(TEST_TOPIC_NAME);
+    PubSubReader<HelloWorldType> compatible_reader("IncompatibleTopic");
```
MiguelCompany: Use a define or constexpr for the incompatible topic name, and use it here and below.

test/blackbox/common/DDSBlackboxTestsListeners.cpp
```diff
@@ -0,0 +1,298 @@
+// Copyright 2019 Proyectos y Sistemas de Mantenimiento SL (eProsima).
```
MiguelCompany, line 1:
```suggestion
// Copyright 2020 Proyectos y Sistemas de Mantenimiento SL (eProsima).
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -932,13 +932,18 @@ fastrtps::TopicAttributes DataWriterImpl::get_topic_attributes(
...
-    offered_incompatible_qos_status_.last_policy_id = policy_id;
-
+    for (fastrtps::rtps::octet id = 0; id < NEXT_QOS_POLICY_ID; ++id)
+    {
+        if (incompatible_policies.test(id))
+        {
+            ++offered_incompatible_qos_status_.policies[static_cast<QosPolicyId_t>(id)].count;
+            offered_incompatible_qos_status_.last_policy_id = static_cast<QosPolicyId_t>(id);
+        }
+    }
```
MiguelCompany: I think we should:
* Start the loop on the first policy, not on the invalid one
* Increase the total counts every time we increase the individual count

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -729,13 +729,18 @@ TypeSupport DataReaderImpl::type()
...
-    requested_incompatible_qos_status_.last_policy_id = policy_id;
-
+    for (fastrtps::rtps::octet id = 0; id < NEXT_QOS_POLICY_ID; ++id)
+    {
+        if (incompatible_policies.test(id))
+        {
+            ++requested_incompatible_qos_status_.policies[static_cast<QosPolicyId_t>(id)].count;
+            requested_incompatible_qos_status_.last_policy_id = static_cast<QosPolicyId_t>(id);
+        }
+    }
```
MiguelCompany: I think we should:
* Start the loop on the first policy, not on the invalid one
* Increase the total counts every time we increase the individual count

include/fastdds/rtps/builtin/discovery/endpoint/EDP.h
```diff
@@ -350,14 +356,14 @@ class EDP
...
-        fastdds::dds::QosPolicyId_t requested_incompatible_qos;
+        fastdds::dds::PolicyMask requested_incompatible_qos;
     };
     std::map<GUID_t, reader_status> reader_status_;
 
     struct writer_status
     {
         fastdds::dds::PublicationMatchedStatus publication_matched;
-        fastdds::dds::QosPolicyId_t offered_incompatible_qos;
+        fastdds::dds::PolicyMask offered_incompatible_qos;
```
MiguelCompany: Looking at this again, I think it is a better solution to keep the data structures as they were, and extend the `validMatching` methods.

I mean, create two new methods having the policy mask as an output parameter, and making the old versions call the new one with the mask as a local variable

```
bool valid_matching(
        const WriterProxyData* wdata,
        const ReaderProxyData* rdata,
        PolicyMask& offered_incompatible_qos);

bool valid_matching(
        const ReaderProxyData* rdata,
        const WriterProxyData* wdata,
        PolicyMask& requested_incompatible_qos);
```

This way we save memory, as well as improve multi-threading behavior.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
IkerLuengo: No, that was part of the listener logic correction I had prepared, but ended up removing. We should have at least an 'else' clause, also on the rest of the listeners. Unfortunately that doesn't correct the logic, but should not do any harm and makes it better.

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
IkerLuengo: Same as before

test/blackbox/api/dds-pim/PubSubReader.hpp
```diff
@@ -549,6 +551,28 @@ class PubSubReader
         return current_received_count_;
     }
 
+    PubSubReader& deactivate_status_listener(
+            eprosima::fastdds::dds::StatusMask mask)
+    {
+        eprosima::fastdds::dds::StatusMask tmp = status_mask_;
+        status_mask_ &= mask;
+        status_mask_ ^= tmp;
```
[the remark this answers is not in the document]
IkerLuengo: me dumb o_O

test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -686,7 +686,148 @@ TEST_P(PubSubBasic, PubSubIncompatibleOfferedQosListeners)
...
+    ASSERT_TRUE(writer.isInitialized());
+
+    // A Reader on the same Topic but with incompatible QoS
+    // Should not match and increase the incompatible QoS occurrences on both Writer and Reader
+    PubSubReader<HelloWorldType> incompatible_reliability_reader(TEST_TOPIC_NAME);
+    incompatible_reliability_reader.reliability(eprosima::fastrtps::RELIABLE_RELIABILITY_QOS)
+                  .deactivate_status_listener(eprosima::fastdds::dds::StatusMask::requested_incompatible_qos()).init();
+    ASSERT_TRUE(incompatible_reliability_reader.isInitialized());
+
+    std::this_thread::sleep_for(std::chrono::seconds(2));
```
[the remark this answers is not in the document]
IkerLuengo: Even better, we can wait for incompatible qos :)

test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -686,7 +686,148 @@ TEST_P(PubSubBasic, PubSubIncompatibleOfferedQosListeners)
...
+    // Should not match and increase incompatible QoS occurrences on both Writer and Reader
+    // Total count of incompatible QoS occurrences in Writer increments, and updates the latest incompatible QoS ID,
+    // but old Reader stays the same
+    PubSubReader<HelloWorldType> incompatible_durability_reader(TEST_TOPIC_NAME);
+    incompatible_durability_reader.reliability(eprosima::fastrtps::BEST_EFFORT_RELIABILITY_QOS)
+            .durability_kind(eprosima::fastrtps::TRANSIENT_LOCAL_DURABILITY_QOS)
+            .deactivate_status_listener(eprosima::fastdds::dds::StatusMask::requested_incompatible_qos()).init();
+    ASSERT_TRUE(incompatible_durability_reader.isInitialized());
+
+    std::this_thread::sleep_for(std::chrono::seconds(2));
```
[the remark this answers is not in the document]
IkerLuengo: same as above

src/cpp/fastdds/publisher/DataWriterImpl.cpp
IkerLuengo: If I'm understanding the standard correctly, the total counts refer to the number of peers with an incompatible QoS, not to the number of incompatible QoS found in total:
```
Total cumulative number of times the concerned DataWriter
discovered a DataReader for the same Topic with a requested QoS that
is incompatible with that offered by the DataWriter.
```
i.e., if we only have 1 peer, the total count should not be greater than one.

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
IkerLuengo: same as above

src/cpp/fastdds/publisher/DataWriterImpl.cpp
MiguelCompany: You're right.

Starting the loop on the first policy still applies, though.

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
MiguelCompany: You're right.

Starting the loop on the first policy still applies, though.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
IkerLuengo: Also added the filtering part, as the unit testing was already testing that

test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -686,7 +686,148 @@ TEST_P(PubSubBasic, PubSubIncompatibleOfferedQosListeners)
...
+    EXPECT_EQ(incompatible_durability_reader.last_incompatible_qos(), eprosima::fastdds::dds::INVALID_QOS_POLICY_ID);
+
+    rstatus = compatible_reader.get_incompatible_qos_status();
+    EXPECT_EQ(rstatus.total_count, 0u);
+    EXPECT_EQ(rstatus.total_count_change, 0u);
+    EXPECT_EQ(rstatus.last_policy_id, eprosima::fastdds::dds::INVALID_QOS_POLICY_ID);
+
+    //No listeners were called
+    EXPECT_EQ(compatible_reader.times_incompatible_qos(), 0u);
+    EXPECT_EQ(compatible_reader.last_incompatible_qos(), eprosima::fastdds::dds::INVALID_QOS_POLICY_ID);
```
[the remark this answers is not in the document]
IkerLuengo: Not sure if I understood correctly, but tried something on these lines. See if that's what you meant.

test/unittest/rtps/discovery/EdpTests.cpp
```diff
@@ -0,0 +1,476 @@
...
+    {
+        rdata->topicKind(TopicKind_t::WITH_KEY);
+    }
+
+    void set_incompatible_type()
+    {
+        rdata->typeName("AnotherTypeName");
+    }
+
+
```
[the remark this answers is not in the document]
IkerLuengo: Had to modify them a little to adapt to the new design

include/fastdds/rtps/builtin/discovery/endpoint/EDP.h
IkerLuengo: As per our conversation, I also added a mask to return the cause of failure to match.

Just one thing more, the mask could potentially return several causes for not matching. But I don't think all of them are even interesting. For example, having a different Topic name already disqualifies for all the rest. Do we want to support returning several active bits on the mask? If so, which ones?

Aborting the matching algorithm as soon as one cause is found may save some time, but I guess the only 'normal' reason is to have different Topic names, and all the rest could be considered 'probably a configuration error', and could be aggregated. What do you think?

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2219/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2024/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2213/)

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -676,22 +676,27 @@ void DataWriterImpl::InnerDataWriterListener::onWriterMatched(
...
 }
 
 void DataWriterImpl::InnerDataWriterListener::on_offered_incompatible_qos(
         RTPSWriter* /*writer*/,
         fastdds::dds::PolicyMask qos)
 {
     OfferedIncompatibleQosStatus& status = data_writer_->update_offered_incompatible_qos(qos);
-    if (data_writer_->listener_ != nullptr)
+    if (data_writer_->listener_ != nullptr  &&
+            (data_writer_->user_datawriter_->get_status_mask() & StatusMask::offered_incompatible_qos()).any())
```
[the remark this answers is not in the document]
IkerLuengo: Have to modify StatusMask for that (it does not provide bit positions for each status, only a complete formed StatusMask with a single active bit), but is probably worth it. Will do

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -676,22 +676,27 @@ void DataWriterImpl::InnerDataWriterListener::onWriterMatched(
...
     {
         data_writer_->listener_->on_offered_incompatible_qos(data_writer_->user_datawriter_, status);
         status.total_count_change = 0u;
     }
-
-    data_writer_->publisher_->publisher_listener_.on_offered_incompatible_qos(data_writer_->user_datawriter_, status);
+    else
+    {
+        data_writer_->publisher_->publisher_listener_.on_offered_incompatible_qos(data_writer_->user_datawriter_, status);
```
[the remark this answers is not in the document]
IkerLuengo: This is the inner listener of the Publisher. It is this callback the one that should check the user's mask (and the existence of the user's listener)

Which currently is not doing, because of the wrong implementation of the listener logic. That was also corrected before I reverted that commit. The thing is it won't be completely correct if we don't correct the DomainParticipant's inner listener too. Don't know if it is worth changing all that on this single status, having all the rest wrong. At least, they are consistently wrong now.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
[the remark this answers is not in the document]
IkerLuengo: Didn't have to modify anything, there is a 'is_active' method that takes a fully formed StatusMask as argument.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2222/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2027/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2216/)

MiguelCompany:
Linter errors are due to uncrustify misconfiguration.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2228/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2034/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2222/)

PR #1274, opened 2020-07-02
Fixed behavior of get_liveliness_changed_status [8812]

MiguelCompany:
It was detected that when no listener is set to a DataReader, `xxx_count_change` fields of `LivelinessChangedStatus` where always being read as 0.

A Blackbox test is added to reproduce the problem.

DataReaderImpl now has a copy of its `LivelinessChangedStatus` independent of the one in RTPSReader.

Status mask is checked before calling `on_liveliness_changed`.

### DIFF
--- a/src/cpp/fastdds/subscriber/DataReaderImpl.cpp
+++ b/src/cpp/fastdds/subscriber/DataReaderImpl.cpp
@@ -415,13 +415,19 @@ void DataReaderImpl::InnerDataReaderListener::on_liveliness_changed(
         RTPSReader* /*reader*/,
         const fastrtps::LivelinessChangedStatus& status)
 {
-    if (data_reader_->listener_ != nullptr)
+    data_reader_->update_liveliness_status(status);
+    DataReader* user_reader = data_reader_->user_datareader_;
+
+    if ( (data_reader_->listener_ != nullptr) &&
+            (user_reader->get_status_mask().is_active(StatusMask::liveliness_changed())))
     {
-        data_reader_->listener_->on_liveliness_changed(data_reader_->user_datareader_, status);
+        LivelinessChangedStatus reader_status;
+        data_reader_->get_liveliness_changed_status(reader_status);
+        data_reader_->listener_->on_liveliness_changed(user_reader, reader_status);
     }
     else
     {
-        data_reader_->subscriber_->subscriber_listener_.on_liveliness_changed(data_reader_->user_datareader_, status);
+        data_reader_->subscriber_->subscriber_listener_.on_liveliness_changed(user_reader, status);
     }
 }
 
@@ -658,20 +664,19 @@ const DataReaderListener* DataReaderImpl::get_listener() const
  */
 
 ReturnCode_t DataReaderImpl::get_liveliness_changed_status(
-        LivelinessChangedStatus& status) const
+        LivelinessChangedStatus& status)
 {
     if (reader_ == nullptr)
     {
         return ReturnCode_t::RETCODE_NOT_ENABLED;
     }
 
-    std::unique_lock<RecursiveTimedMutex> lock(reader_->getMutex());
+    std::lock_guard<RecursiveTimedMutex> lock(reader_->getMutex());
 
-    status = reader_->liveliness_changed_status_;
+    status = liveliness_changed_status_;
+    liveliness_changed_status_.alive_count_change = 0u;
+    liveliness_changed_status_.not_alive_count_change = 0u;
 
-    reader_->liveliness_changed_status_.alive_count_change = 0u;
-    reader_->liveliness_changed_status_.not_alive_count_change = 0u;
-    // TODO add callback call subscriber_->subscriber_listener_->on_liveliness_changed
     return ReturnCode_t::RETCODE_OK;
 }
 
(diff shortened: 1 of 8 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2279/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2080/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2272/)

src/cpp/fastdds/subscriber/SubscriberImpl.cpp
```diff
@@ -471,9 +471,13 @@ void SubscriberImpl::SubscriberReaderListener::on_liveliness_changed(
         DataReader* reader,
         const fastrtps::LivelinessChangedStatus& status)
 {
+    (void)status;
+
     if (subscriber_->listener_ != nullptr)
```
IkerLuengo, line 476: If we are now checking the status mask on the DataReader, I'd add the check here too.

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -415,13 +415,19 @@ void DataReaderImpl::InnerDataReaderListener::on_liveliness_changed(
...
     {
-        data_reader_->listener_->on_liveliness_changed(data_reader_->user_datareader_, status);
+        LivelinessChangedStatus reader_status;
+        data_reader_->get_liveliness_changed_status(reader_status);
+        data_reader_->listener_->on_liveliness_changed(user_reader, reader_status);
     }
     else
     {
-        data_reader_->subscriber_->subscriber_listener_.on_liveliness_changed(data_reader_->user_datareader_, status);
+        data_reader_->subscriber_->subscriber_listener_.on_liveliness_changed(user_reader, status);
```
IkerLuengo, line 430: I'd pass the internal updated liveliness_changed_status_ to the parents. I know they are not using it, and reading from the `reader` instead, but somehow it feels wrong sending the status that is not managed at DDS level.

src/cpp/fastdds/subscriber/SubscriberImpl.cpp
```diff
@@ -473,11 +476,23 @@ void SubscriberImpl::SubscriberReaderListener::on_liveliness_changed(
 {
     (void)status;
 
-    if (subscriber_->listener_ != nullptr)
+    SubscriberListener* listener = subscriber_->listener_;
+    if (!subscriber_->user_subscriber_->get_status_mask().is_active(StatusMask::liveliness_changed()))
```
IkerLuengo, line 480: Should try the participant also if the listener is not set, even when the mask is active. 
```suggestion
    if (listener == nullptr || !subscriber_->user_subscriber_->get_status_mask().is_active(StatusMask::liveliness_changed()))
```

src/cpp/fastdds/subscriber/SubscriberImpl.cpp
```diff
@@ -473,11 +476,23 @@ void SubscriberImpl::SubscriberReaderListener::on_liveliness_changed(
...
+    {
+        auto participant = subscriber_->get_participant();
+        auto part_listener = const_cast<DomainParticipantListener*>(participant->get_listener());
+        listener = static_cast<SubscriberListener*>(part_listener);
+        if (!participant->get_status_mask().is_active(StatusMask::liveliness_changed()))
+        {
+            listener = nullptr;
+        }
+    }
```
IkerLuengo: I'd like the callback handling to be the same in all entities, which is not right now:
- The DataReader checks its user listener and mask, if it cannot process the callback, forwards to the Subscriber's inner listener `SubscriberReaderListener`
- The Subscriber checks its user listener and mask, if it cannot process the callback, the Subscriber itself checks the Participant's listener and mask and calls the callback of the Subscriber of Participant accordingly.

It would be beautiful if all entities forward the callback to the parent entity's inner listener, and each Entity checks its own listener and mask; or the other way round, that al the checks are done in the lowest entity (DataReader in this case).

I see drawbacks for both solutions: the former adds new callbacks to the parent entity's inner listener that may have little to do with that entity's work, the latter makes the lower entities access all the parent hierarchy to the top.

Not saying that this must be addressed on this PR (wouldn't reject it because of this), but I think it is worth a though, mostly to decide how we want to implement the listener logic correction (task 8717)

MiguelCompany: I would have a `DataReaderListener* get_reader_listener_for(StatusMask status)` on all implementation entities. Each one will check if they have a listener set and the received mask active, and bubble up to their parent when it is not the case. With that, we could have the call to the listener on a single place (the bottom inner listener).

But, as you say, that is worth a separate PR.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2286/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2086/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2278/)

PR #1283, opened 2020-07-07
Update colcon installations instructions

EduPonz:
Create a `fastdds_ws` root directory for the colcon workspace

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ pip install -U colcon-common-extensions vcstool
 Download the repos file that will be used to download Fast RTPS and its dependencies:
 
 ```bash
+$ mkdir fastdds_ws
+$ cd fastdds_ws
 $ wget https://raw.githubusercontent.com/eProsima/Fast-DDS/master/fastrtps.repos
 $ mkdir src
 $ vcs import src < fastrtps.repos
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -100,6 +100,8 @@ pip install -U colcon-common-extensions vcstool
 Download the repos file that will be used to download Fast RTPS and its dependencies:
 
 ```bash
+$ mkdir fastdds_ws
+$ cd fastdds_ws
 $ wget https://raw.githubusercontent.com/eProsima/Fast-DDS/master/fastrtps.repos
```
jwillemsen, line 105: Why not rename fastrtps.repos to fastdds.repos and update its content to also clone Fast-DDS and not fastrtps

EduPonz: I've modified `fastrtps.repos` to point to Fast-DDS, again good catch!

However, we have decided not to rename the file yet. The process of changing product name is a complicated one and we have to be very careful with the implications. I'm afraid such a small change will break buildings for many people and in that sense we consider it similar to an API break, which we only allow when changing version major, in accordance with our [versioning policy](https://github.com/eProsima/Fast-DDS/blob/master/VERSIONING.md).

jwillemsen: Yes, renaming the file could break things, always hard to do a project rename

EduPonz: I'll mark suggestion as resolved then

PR #1293, opened 2020-07-10
Fixes topic name

LuisGP:
Signed-off-by: LuisGasco <luisgasco@eprosima.com>


EduPonz:
I've marked the PR with `no-test` label since examples are not run in CI

Issue #1342, opened 2020-08-11
Equivalent to idl module keyword in Dynamic Data Types 


mdnasr:
FastDDS generator support `module` keyword in IDL. The generated code is wrapped in C++ namespace. Are there any way to do the same thing using Dynamic Data Types?

## Steps to Reproduce

To clarify, Most ROS IDL use `module` keyword in their IDL

```
module sensor_msgs {
  module msg {
    struct Image {
      uint32 height;
      uint32 width;
    };
  };
};
```

Topic data type of such IDL will be `sensor_msgs::msg::Image` using `myType.getName();`
I want to create equivalent type to this using Dynamic Data Types in C++ or XML, but I couldn't find any API to replicate `module` keyword.

## System information

- Fast-RTPS version: 1.10.x
- OS: Ubuntu 16.04

LuisGP:
Hi @mdnasr,

The `module` concept isn't part of the standard and we haven't implemented it.
You can define the Dynamic Type name to something similar in a C way in its builder: `sensor_msgs__msg__Image`.

What is your use case?

Best regards.

EduPonz:
I'm closing this issue as the question has been solved. Please feel free to comment or reopen at any time.

PR #1345, opened 2020-08-17
Support colon in type name for dynamic data types 

mdnasr:
To solve #1342
This PR add support for colon ":" to data type name, to make it compatible with `module` keyword in FastDDS generator and works perfectly with ROS messages.

For example, The following XML dynamic data types will work now.

```
<types>
  <type>
    <struct name="sensor_msgs::msg::Image">
      <member name="height" type="uint32"/>
      <member name="width" type="uint32"/>
      ...
    </struct>
  </type>
</types>

```

### DIFF
--- a/src/cpp/dynamic-types/TypeDescriptor.cpp
+++ b/src/cpp/dynamic-types/TypeDescriptor.cpp
@@ -253,7 +253,7 @@ bool TypeDescriptor::is_type_name_consistent(
     if (!sName.empty() > 0 && std::isalpha(sName[0]))
     {
         // All characters must be letters, numbers, underscore, or colon.
-        for (uint32_t i = 1; i < sName.length(); ++i)
+        for (size_t i = 1; i < sName.length(); ++i)
         {
             if (!std::isalnum(sName[i]) && sName[i] != 95 && sName[i] != 58)
             {
--- a/src/cpp/dynamic-types/TypeDescriptor.cpp
+++ b/src/cpp/dynamic-types/TypeDescriptor.cpp
@@ -250,7 +250,7 @@ bool TypeDescriptor::is_type_name_consistent(
         const std::string& sName) const
 {
     // The first letter must start with a letter ( uppercase or lowercase )
-    if (!sName.empty() > 0 && std::isalpha(sName[0]))
+    if (!sName.empty() && std::isalpha(sName[0]))
     {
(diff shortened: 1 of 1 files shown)


src/cpp/dynamic-types/TypeDescriptor.cpp
```diff
@@ -252,10 +252,10 @@ bool TypeDescriptor::is_type_name_consistent(
     // The first letter must start with a letter ( uppercase or lowercase )
     if (sName.length() > 0 && std::isalpha(sName[0]))
```
jwillemsen, line 253:
``` suggestion
    if (!sName.empty() > 0 && std::isalpha(sName[0]))
```

jwillemsen: std::isalpha takes current C locale into account, that doesn't seem correct at this point?

src/cpp/dynamic-types/TypeDescriptor.cpp
```diff
@@ -252,10 +252,10 @@ bool TypeDescriptor::is_type_name_consistent(
...
     {
-        // All characters must be letters, numbers or underscore.
+        // All characters must be letters, numbers, underscore, or colon.
         for (uint32_t i = 1; i < sName.length(); ++i)
```
jwillemsen, line 256:
``` suggestion
        for (size_t i = 1; i < sName.length(); ++i)
```

src/cpp/dynamic-types/TypeDescriptor.cpp
```diff
@@ -252,10 +252,10 @@ bool TypeDescriptor::is_type_name_consistent(
...
         {
-            if (!std::isalnum(sName[i]) && sName[i] != 95)
+            if (!std::isalnum(sName[i]) && sName[i] != 95 && sName[i] != 58)
             {
```
jwillemsen, line 259: Also std::isalnum takes current C locale into account https://en.cppreference.com/w/cpp/string/byte/isalnum

MiguelCompany (changes requested):
Changes look good to me, but could you please fix DCO and apply our [uncrustify config](https://github.com/eProsima/cpp-style/blob/master/uncrustify.cfg) to the whole file?

MiguelCompany:
Closing in favor of #1831

PR #1351, opened 2020-08-28
Feature/gen examples [9179]

jparisu:
**Update the examples from FastRTPS and FastDDS.**

The tool Fast-DDS-GEN / Fast-RTPS-GEN is being used to modify the types that every example uses.
- The types used in the fastrtps examples has been updated with Fast-RTPS-GEN v1.0.4
- The types used in the fastdds examples has been updated with Fast-DDS-GEN v2.0.0

The examples have been compiled and execute to verify their good behavior (or at least the same previous behavior, some issues has already been raised in order to fix some examples).

Changes committed:

- Every example has it's own .idl file where the type is specified. The types used for the examples have been updated using the tool get auto-generated code.

- The Dynamic example does not need type files or an .idl file. This file has been renamed to .idl.example in case the user wants to review it.

- In some cases the old Types were used. Instead, those files has been deleted and replaced with PubSubTypes files.

- The names of some executables have been modified to uniform the way they are built. Every executable is called as the Example Directory it belongs. This has been achieved by changing the executable name in the CMakeLists.txt files.

- Some example's code has been modified in order to amend inconsistency between the old Type names and the new ones.


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2496/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2310/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2503/)

jwillemsen:
Why put the generated files into the archive, why not just run the IDL compiler as part of compiling an example or test, any DDS user would also need the IDL compiler for his own code. That way the IDL compiler code is always fully tested with a change.

jparisu:
> Why put the generated files into the archive, why not just run the IDL compiler as part of compiling an example or test, any DDS user would also need the IDL compiler for his own code. That way the IDL compiler code is always fully tested with a change.

The Fast-DDS-GEN library is a different library than Fast-DDS and it is not required in the Fast-DDS compilation.
This way, we can provide any user with runnable examples, avoiding the necessity of compiling and executing the Fast-DDS-GEN library, what would needlessly overhead the compilation of the library.
It could be some Fast-DDS users that use dynamic typing, or just demo users that may not want the Fast-DDS-GEN library to be compiled.

jwillemsen:
Compiling Fast-DDS-Gen should be easy, maybe for a very very first experiment someone could just want to only execute an example but very likelye in his second attempt he will modify some IDL and need the IDL compiler. Maybe it should be simplified to also use Fast-DDS-Gen.

If you want to keep this 100% separate, is there a simple way to just run the IDL compiler on all examples/tests, we reported some ideas for improvements to the IDL generated code but how do we test that easily on all examples/tests?

imontesino (approved):
LGTM

EduPonz:
Since the changes solely affect the examples, which are not tested in the CI, I'm going to mark this PR as `no-test`. Please manually test the examples before approving.

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2502/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2316/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2509/)

PR #1435, opened 2020-09-30
Fix enable separate sending in subscriptions writer [9477]

EduPonz:
The subcriptions writer must have separate sending enable for the DS 2.0 to work properly.

Signed-off-by: EduPonz <eduardoponz@eprosima.com>


MiguelBarro (approved):
🤦

PR #1439, opened 2020-10-01
Ackfunctor and History refactoring [9443] 

MiguelBarro:

PR #1445, opened 2020-10-02
Server DDB: Check sample id on ack [9496]

EduPonz:
When an acknowledgement is notified to the database, check if the change has not been updated before marking it as acknowledge.

Marked as `no-test` until #1439 is merged


jparisu (approved):
LGTM

PR #1450, opened 2020-10-02
Servers receive all clients info [9500]

EduPonz:
This PR adds for each server a virtual reader and writer in a virtual topic so that the virtual entities match with all topics. Furthermore, info about the virtual entities is not sent to the clients.

Merge after #1442


EduPonz:
Closed in favor of #1473

PR #1474, opened 2020-10-08
governance rules order <master> [9583]

IkerLuengo:
According to the standard, governance rules must be evaluated in order, and the first matching rule is applied.
We are using a map to store the rules before applying them, therefore, losing the original order.
Besides, the map object is never searched by key.
Changed the map for a vector.

### DIFF
--- a/src/cpp/security/accesscontrol/Permissions.cpp
+++ b/src/cpp/security/accesscontrol/Permissions.cpp
@@ -695,18 +695,13 @@ static bool check_subject_name(
                     for (auto topic_rule : rule.topic_rules)
                     {
                         std::string topic_expression = topic_rule.topic_expression;
-                        EndpointSecurityAttributes reader_attributes;
-                        EndpointSecurityAttributes writer_attributes;
+                        EndpointSecurityAttributes security_attributes;
                         PluginEndpointSecurityAttributes plugin_attributes;
 
-                        reader_attributes.is_discovery_protected = topic_rule.enable_discovery_protection;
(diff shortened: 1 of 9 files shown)


src/cpp/security/accesscontrol/Permissions.cpp
```diff
@@ -672,9 +672,9 @@ static bool check_subject_name(const IdentityHandle& ih, AccessPermissionsHandle
                         reader_attributes.plugin_endpoint_attributes = plugin_attributes.mask();
                         writer_attributes.plugin_endpoint_attributes = plugin_attributes.mask();
 
-                        ah->governance_reader_topic_rules_.insert(std::pair<std::string, EndpointSecurityAttributes>(
+                        ah->governance_reader_topic_rules_.push_back(std::pair<std::string, EndpointSecurityAttributes>(
                                     topic_expression, std::move(reader_attributes)));
-                        ah->governance_writer_topic_rules_.insert(std::pair<std::string, EndpointSecurityAttributes>(
+                        ah->governance_writer_topic_rules_.push_back(std::pair<std::string, EndpointSecurityAttributes>(
                                     std::move(topic_expression), std::move(writer_attributes)));
```
MiguelCompany, line 741: It seems to me that `reader_attributes` and `writer_attributes` will always have the same values. Perhaps we could have a single collection `governance_topic_rules` ?

MiguelCompany (changes requested):

MiguelCompany (approved):
LGTM

PR #1490, opened 2020-10-15
Keep pending security messages directed for each local endpoint [9621]

IkerLuengo:
`VolatileSecureMessage`s received for local endpoints that still didn't discover the remote endpoint are stored to be processed when the local endpoint discovers the remote one. The `SecurityManager` keeps them on a map keyed by the remote endpoint GUID.

But if the remote endpoint matches with multiple local endpoints, multiple `VolatileSecureMessage` will be received, one per match. However, all of them share the same origin GUID, so only a single copy will be stored on the map. Once the first local endpoint discovers the remote and completes the security processing, it will remove the message from the map of pendings, leaving the rest of the local endpoints without the required information to complete the matching.

Instead of using the remote GUID as key, we should be using the combination of the remote GUID and the GUID of the local endpoint to which the message was sent. Then each local endpoint will be able to find its copy of the message.

### DIFF
--- a/src/cpp/rtps/security/SecurityManager.cpp
+++ b/src/cpp/rtps/security/SecurityManager.cpp
@@ -198,7 +198,7 @@ bool SecurityManager::init(
             }
 
             if (!( logging_plugin_->set_log_options(log_options, exception) &&
-                    logging_plugin_->enable_logging(exception) ))
+                    logging_plugin_->enable_logging(exception)))
             {
                 return init_logging_fail(exception);
             }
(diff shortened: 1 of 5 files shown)


MiguelCompany (changes requested):
Would be nice to have a blackbox test with several secure endpoints on the same participant

MiguelCompany (approved):
LGTM

MiguelCompany (approved):
LGTM

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/2911/)
* Linux aarch64[Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux%20aarch64/198/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2663/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/2922/)

PR #1547, opened 2020-11-05
[Discovery Server] Removed log categrory filter and log verbosity in fast-discovery-server tool

rsanchez15:


EduPonz (approved):
LGTM

MiguelCompany:
I think we are safe merging this without testing, since it will correctly build (as it is only removing code), and the code removed will not affect test results.

PR #1553, opened 2020-11-06
Fix mismatching tag in package.xml

ThijsRay:
Building this package with ROS fails.

Error message:
```
Error(s) in package '/ros2_foxy/src/eProsima/Fast-DDS/package.xml':
The manifest contains invalid XML:
mismatched tag: line 20, column 22
```

This pull request fixes this error by changing `dev` to `depend` in `package.xml`.

This fixes the master branch, but this is also fixed in the 2.0.x branch with #1554

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -17,7 +17,7 @@
   <buildtool_depend>cmake</buildtool_depend>
 
   <depend>fastcdr</depend>
-  <depend>libssl-dev</dev>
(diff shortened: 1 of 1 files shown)


ThijsRay:
This change should also be cherry-picked to the 2.0.x branch.

MiguelCompany (approved):

PR #1555, opened 2020-11-06
[Discovery Server] Features for Backup Mode <2.0.x>

jparisu:
This PR contains the different features needed to backport the backup mode for Discovery-Server.
These features are:
- Persistence schema refactor (Before in PR #1528 )
- New ParticipantType (Before in PR #1521 )
- Hotfixes for Discovery-Server (Before in PR #1545 )
- Backup implementation (Before in PR #1550 )

PR #1533, opened 2020-10-30
[Discovery Server] Backup mode to Discovery-Server [9841]

jparisu:
Do not merge before PR #1323 
Do not merge before PR #1545 

This changes implement the Backup mode to the Discovery-Server.

PR #1558, opened 2020-11-06
Resolve unused variable warning

brawner:
#1555 introduced this unused variable. It doesn' look like the code needs to refer to the pair itself, so I thought it best to just remove the iterator variable.

Signed-off-by: Stephen Brawner <brawner@gmail.com>

### DIFF
--- a/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
+++ b/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
@@ -2322,7 +2322,7 @@ bool DiscoveryDataBase::from_json(
             }
 
             // Add Participant
-            auto wit = writers_.insert(std::make_pair(guid_aux, dei));
+            writers_.insert(std::make_pair(guid_aux, dei));
 
             // Extra configurations for writers
             // Add writer to writers_by_topic. This will create the topic if necessary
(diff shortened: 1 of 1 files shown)


src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -2322,7 +2322,7 @@ bool DiscoveryDataBase::from_json(
             }
 
             // Add Participant
-            auto wit = writers_.insert(std::make_pair(guid_aux, dei));
+            writers_.insert(std::make_pair(guid_aux, dei));
```
EduPonz, line 2325: That's a tricky one, as the variable is indeed used in line 2346, which is a `logInfo` message that is not compiled unless `INTERNAL_DEBUG` is set or `CMAKE_BUILD_TYPE` is set to `Debug`. Our CI builds with `INTERNAL_DEBUG` to verify that all `logInfo` are correct, and thus we don't get the warning. The best is probably to remove the part of the log that uses it.

EduPonz (changes requested):

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
brawner: What do you think about just putting in a `static_cast<void>`, like my most recent commit? It suppresses the warning, but still works for the log?

EduPonz (approved):

EduPonz:
I've added this to #1533 . Thanks @brawner for the catch!

PR #1571, opened 2020-11-20
Update entityId on big endian system

advancenOO:
The byte c[3] will always be dropped when creating entityId, but it could be the only byte that contains information in the variable idnum on big endian system, especially when idnum is created by calling setEntityID(uint8_t id).

### DIFF
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
@@ -492,16 +492,9 @@ bool RTPSParticipantImpl::preprocess_endpoint_attributes(
             idnum = IdCounter;
         }
 
-        octet* c = reinterpret_cast<octet*>(&idnum);
-#if __BIG_ENDIAN__
-        entId.value[2] = c[1];
-        entId.value[1] = c[2];
-        entId.value[0] = c[3];
-#else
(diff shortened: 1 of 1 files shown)


MiguelCompany (changes requested):
I think the code was wrong in the first place. It should be better to be semantically explicit. Something like the following

```
entId.value[2] = idnum & 0xFF;
entId.value[1] = (idnum >> 8) & 0xFF;
entId.value[0] = (idnum >> 16) & 0xFF;
```

We could then also remove the hideous `reinterpret_cast`

MiguelCompany (approved):
LGTM with green CI!

richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/3138/)
* Linux aarch64[Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux%20aarch64/356/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2801/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/3170/)

MiguelCompany:
Known flaky failures. Going in.

@advancenOO thanks for the contribution!

PR #1627, opened 2020-12-06
Only run tests with non-local interfaces if such an interface exists

roehling:
The now guarded code assumes that the interface list is never empty, and will throw an exception on `interfaces.at(0).name` otherwise.

I used `interfaces.size() > 0` for consistency with `open_and_close_two_multicast_transports_with_whitelist`. `!interfaces.empty()` would probably be more efficient, although I doubt that it makes a noticeable difference here.

Also, starting with GTest 1.10, you could call `GTEST_SKIP()` to make it explicit that the test is not actually run.

### DIFF
--- a/test/unittest/transport/UDPv4Tests.cpp
+++ b/test/unittest/transport/UDPv4Tests.cpp
@@ -463,57 +463,60 @@ TEST_F(UDPv4Tests, send_and_receive_between_allowed_sockets_using_unicast)
     std::vector<IPFinder::info_IP> interfaces;
     GetIP4s(interfaces);
 
-    for (const auto& interface : interfaces)
+    if (interfaces.size() > 0)
     {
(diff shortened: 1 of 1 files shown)


richiware:
Build status:
* Linux [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux/3260/)
* Linux aarch64[Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Linux%20aarch64/471/)
* Mac [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Mac/2912/)
* Windows [Build Status](http://jenkins.eprosima.com:8080/job/FastRTPS%20Manual%20Windows/3292/)

MiguelCompany (approved):
LGTM

PR #1636, opened 2020-12-11
Fix path on boost qnxnto.hpp

### LINKED ISSUE #1635: Wrong posix_features.hpp path for QNX Platform in Boost configuration
## Expected Behavior
Thirdparty boost should be compiled properly for QNX finding the proper path for posix_features.hpp.

## Current Behavior
The compilation fails due to the following wrong path in `boost/config/platform/qnxnto.hpp`:
`#include <boost/config/posix_features.hpp>
`
It must be fixed to:
`#include <boost/config/detail/posix_features.hpp>
`
This is the same path used for example in the linux.hpp file.

MiguelCompany:
This should fix #1635.

I omit CI, as it will not check QNX related changes.

### DIFF
--- a/thirdparty/boost/include/boost/config/platform/qnxnto.hpp
+++ b/thirdparty/boost/include/boost/config/platform/qnxnto.hpp
@@ -10,7 +10,7 @@
 #define BOOST_PLATFORM "QNX"
 
 #define BOOST_HAS_UNISTD_H
-#include <boost/config/posix_features.hpp>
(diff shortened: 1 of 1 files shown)


IkerLuengo (changes requested):
Correct linter errors on the modified file

MiguelCompany:
> Correct linter errors on the modified file

We don't apply code style to third-parties, even when they are part of our repo

IkerLuengo (approved):
The you are good to go!

PR #1715, opened 2021-01-28
Refs #10416: use constexpr BOUND_UNLIMITED [10416]

JLBuenoLopez-eProsima:
Instead of using a hardcoded zero, use a constexpr which can also be used in other parts of the library when building a bound unlimited quantity.


MiguelCompany (approved):
LGTM

PR #1780, opened 2021-02-23
[10646] Fix various CMake issues

MiguelBarro:
## Fix Fast-DDS build on raspbian buster
`FindAtomic.cmake` module relies on CMake module `CheckLibraryExists` to check if **libatomic.so** is present in the system.
This module implementation just checks that specific APIs in the given library can be successfully linked in a dummy source file.
If the module build reports any other error unrelated to the linker ability to find the APIs the wrongly module assumes the APIs are not available.
Raspbian buster g++ version arises an error (related with API signature redefinition) that gcc considers a warning. Thus in this OS the checks must be done using the C compiler. Note that in MSVC and Clang conversely the C++ compiler must be used to avoid unrelated errors.
## Update CMake config.h and windows warnings
Now if c++20 or c++17 standard is enforced a corresponding preprocessor macro is defined (as in C++14 and C++11 were).
A new windows warning is promoted to match clang warning level.

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -375,10 +375,11 @@ elseif(NOT EPROSIMA_INSTALLER)
     # make visual studio more gcc and clang like. Propagated to dependent targets.
     if(MSVC)
         # C4101 'identifier' : unreferenced local variable 
-        # C4555 expression has no effect; expected expression with side-effect 
         # C4189 'identifier' : local variable is initialized but not referenced
+        # C4555  expression has no effect; expected expression with side-effect
+        # C4715: 'Test': not all control paths return a value
         # C5038 data member 'member1' will be initialized after data member 'member2'
-        target_compile_options(${PROJECT_NAME} PUBLIC /w34101 /w34555 /w34189 /w35038)
+        target_compile_options(${PROJECT_NAME} PUBLIC /w34101 /w34555 /w34189 /w35038 /w34101)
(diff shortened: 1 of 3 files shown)


src/cpp/CMakeLists.txt
```diff
@@ -375,10 +375,11 @@ elseif(NOT EPROSIMA_INSTALLER)
     # make visual studio more gcc and clang like. Propagated to dependent targets.
     if(MSVC)
         # C4101 'identifier' : unreferenced local variable 
-        # C4555 expression has no effect; expected expression with side-effect 
         # C4189 'identifier' : local variable is initialized but not referenced
+        # C4555  expression has no effect; expected expression with side-effect
+        # C4715: 'Test': not all control paths return a value
         # C5038 data member 'member1' will be initialized after data member 'member2'
-        target_compile_options(${PROJECT_NAME} PUBLIC /w34101 /w34555 /w34189 /w35038)
+        target_compile_options(${PROJECT_NAME} PUBLIC /w34101 /w34555 /w34189 /w35038 /w34101)
```
EduPonz, line 382:
```suggestion
        target_compile_options(${PROJECT_NAME} PUBLIC /w34101 /w34555 /w34189 /w35038 /w34715)
```

EduPonz (changes requested):

EduPonz (approved):
LGTM

EduPonz:

PR #1784, opened 2021-02-24
clean reader history on performance tests 

IkerLuengo:
This PR:
* Makes sure that the reader history is empty when a new test is recycling the same reader
* Considers data-sharing overrides as a lost sample


MiguelCompany (approved):
LGTM!

Issue #1766, opened 2021-02-17
flock does not work under QNX on files created in /dev/shmem under QNX [10619]


asobhy-qnx:
https://github.com/eProsima/Fast-DDS/blob/a5a7d162e483ddc30f8f45a73ed0ad4b0ad974f7/src/cpp/rtps/transport/shared_mem/RobustExclusiveLock.hpp#L192

flock returns "The filesystem doesn't support the operation" due to that the file descriptor represents a file created under /dev/shmem

Can the robust lock file be created under another path? what are the consequences of creating a file in /dev/shmem and not locking it?

asobhy-qnx:
What is the purpose of flock in this implementation?

IkerLuengo:
The only purpose of the file we are locking is to act as an interprocess exclusion lock that is automatically freed when the creator process dies unexpectedly. These lock files always have a companion file where the data is actually shared among the processes, and both are created simultaneously by the same process. The lock file is used to detect when the creator process dies, so that we can garbage-collect and remove both.

The reason we are using flock is that it will be automatically freed when the process dies. You may have another suggestion that works for QNX.

asobhy-qnx:
You mentioned above, "These lock files always have a companion file where the data is actually shared among the processes", which got me a bit confused. The companion file you mean by that the shared memory? or a regular file?

So here is my understanding:

a process creates a shared memory then it creates a file tagged with a segment number and locks it. The purpose of the file is to tell during clean up if there has been shared memory created or not and if yes then it cleans it up, correct?

 Do we have to lock the file? In QNX we can create regular files under /dev/shmem but we cannot lock them because they are not regular files. We have two options here:

1- if the locking is necessary then we can create the file in any other path other than /dev/shmem then we can lock it just fine.
2- create the file under /dev/shmem but do not lock it.

Which one do you think would be the better option?

Thank you

liutiegang:
Could someone please help on this issue? This issue can reproduce on QNX7.0 armV8 target.

liutiegang:
I use a test demo to run, got the following log:

182885.178017408 Acquiring LOCK_EX on fd=3, fname='/tmp/fastrtps_d61cc318b6a158ff_el'
flock(fd=3, fname='/tmp/fastrtps_d61cc318b6a158ff_el', LOCK_EX): Function not implemented

it means QNX7 didn't implement the LOCK_EX of flock function?

Who can help on it?

Thanks.

IkerLuengo:
[...]

PR #1788, opened 2021-02-25
[10647] Provide an alternative path for robustlock files under QNX <master>

IkerLuengo:
This is a port of #1766 to master

### DIFF
--- a/src/cpp/utils/shared_memory/SharedDir.hpp
+++ b/src/cpp/utils/shared_memory/SharedDir.hpp
@@ -18,6 +18,12 @@
 #include <boostconfig.hpp>
 #include <boost/interprocess/detail/shared_dir_helpers.hpp>
 
+#ifdef __QNXNTO__
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <errno.h>
+#endif  // __QNXNTO__
+
 namespace eprosima {
 namespace fastdds {
 namespace rtps {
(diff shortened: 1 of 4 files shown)


MiguelCompany (changes requested):
Change [this line](https://github.com/eProsima/Fast-DDS/blob/0195e51ab7ef9fe2d7039acbcf66dd1c03e3fc9d/src/cpp/rtps/DataSharing/DataSharingPayloadPool.hpp#L26) to include `SharedDir.hpp` instead of `RobustExclusiveLock.hpp`

MiguelCompany (dismissed):

MiguelCompany:
@asobhy-qnx Would you be so kind to check if this works on QNX?

asobhy-qnx:
Sure

asobhy-qnx:
got this error:
[RTPS_MSG_OUT Error] error: couldn't generate segment_name -> Function init
 I have not looked at the source yet will let you know when I figure out what is wrong

src/cpp/utils/shared_memory/SharedDir.hpp
```diff
@@ -67,6 +73,37 @@ class SharedDir
...
+            {
+                std::string err("get_file_path() ");
+                err = err + strerror(errno);
+                throw std::runtime_error(err);
+            }
+        }
+        else
+        {
+            // directory exists do nothing
+        }
```
asobhy-qnx, line 101: this function does not return anything for when ```__QNXNTO__``` is defined.
add the following line:

```
return (std::string(defaultdir) + "/" + filename) ;
```

asobhy-qnx (changes requested):

src/cpp/utils/shared_memory/SharedDir.hpp
MiguelCompany: @asobhy-qnx Thanks for the suggestion! I missed that on my review 🤦 

I have added the necessary change on 061a563. Please check that it works.

MiguelCompany:
@richiprosima Please check style

src/cpp/utils/shared_memory/SharedDir.hpp
asobhy-qnx: Works, thank you.

MiguelCompany (approved):

PR #1764, opened 2021-02-16
Add more verbosity to discovery CLI [10611]

EduPonz:
This PR adds more verbosity to the stdout of the `fastdds discovery` command. Example:

```
### Server is running ###
  Server ID:          0
  Server GUID prefix: 44.53.00.5f.45.50.52.4f.53.49.4d.41
  Server Address:     UDPv4:[127.0.0.1]:11811
  Participant Type    SERVER
### Server shut down ###
```

This way, creating an XML file for CLIENTS connecting to the SERVER is easier, since the prefix is clearly stated. For doing this, a `<<` operator has been added for the `DiscoveryProtocol` enum. Furthermore, the `<<` operator for the `GuidPrefix` has been enhanced, so that the bytes are always showed as two digits, adding a preceding 0 if necessary.

It is important to note the the default GUID prefix for the server has changed. The following schema is applied now:

* `44.53.<server_id>.5f.45.50.52.4f.53.49.4d.41`, which translated into ASCII as `DS<id_in_hex>_EPROSIMA`.

Finally, the participant type property has been added to the `SIMPLE` participant's DATA(p)s

PR #1799, opened 2021-03-02
List all server locators [10725]

EduPonz:
With this PR, `fastdds discovery` lists all the locators specified for the server.

Signed-off-by: EduPonz <eduardoponz@eprosima.com>

### DIFF
--- a/tools/fds/server.cpp
+++ b/tools/fds/server.cpp
@@ -212,7 +212,6 @@ int main (
     // Create the server
     int return_value = 0;
     Participant* pServer = Domain::createParticipant(*att, nullptr);
-    att.reset();
 
     if ( nullptr == pServer )
     {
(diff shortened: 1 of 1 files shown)


tools/fds/server.cpp
```diff
@@ -228,10 +227,20 @@ int main (
...
-        cout << "  Participant Type    " << rtps.builtin.discovery_config.discoveryProtocol << endl;
+        cout << "  Server Addresses:   ";
+        for (auto locator_it = att->rtps.builtin.metatrafficUnicastLocatorList.begin();
+                locator_it != att->rtps.builtin.metatrafficUnicastLocatorList.end();)
+        {
+            cout << *locator_it;
+            if (++locator_it != att->rtps.builtin.metatrafficUnicastLocatorList.end())
+            {
+                cout << std::endl << "                      ";
+            }
```
JLBuenoLopez-eProsima, line 241:
```suggestion
        for (auto locator : att->rtps.builtin.metatrafficUnicastLocatorList)
        {
            cout << locator << std::endl << "                      ";
        }
```

I suggest using the ranged for. Also, LocatorList class have operator `<<` overloaded so we can also take advantage of this.

JLBuenoLopez-eProsima (changes requested):

JLBuenoLopez-eProsima:
Please, check also the following Discovery Server v2.0 tests that are failing and it is probably due to the changes implemented in this PR:

* 22: environment variable setup
* 23: fast discovery server tool

tools/fds/server.cpp
EduPonz: I actually was doing that in the beginning and then at some point switched to the iterators to see if that solved another problem I was having. I'll go back to range for 👌🏼

EduPonz: After internal discussion, we have decided to leave the iterator to be able to do different things for the last element

EduPonz:
This tests should be failing since #1764 . PR eProsima/Discovery-Server#25 was issued to deal with that change

JLBuenoLopez-eProsima (approved):
LGTM

EduPonz:
@richiprosima please test discovery-server

PR #1803, opened 2021-03-04
[10700] Add FASTDDS_STATISTICS CMake option

JLBuenoLopez-eProsima:
New Statistics tool requires a CMake option that enables the building of the specific code implemented for this tool to prevent taxing the performance.

Here the [related documentation PR](https://github.com/eProsima/Fast-DDS-docs/pull/241).

### DIFF
--- a/versions.md
+++ b/versions.md
@@ -1,7 +1,7 @@
 Forthcoming
 -----------
 
-* New Fast DDS Statistics module (implies ABI break)
+* New Fast DDS Statistics module
 
 Version 2.2.0
 -------------
--- a/versions.md
+++ b/versions.md
@@ -1,7 +1,7 @@
 Forthcoming
 -----------
 
-* New Fast DDS Instrumentation module (implies ABI break)
+* New Fast DDS Statistics module (implies ABI break)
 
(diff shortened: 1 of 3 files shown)


versions.md
```diff
@@ -1,6 +1,8 @@
 Forthcoming
 -----------
 
+* New Fast DDS Statistics module (implies ABI break)
```
EduPonz, line 4: I don't think this is necessarily true. Not on the DDS API at least

EduPonz (changes requested):

versions.md
MiguelCompany: But we are going to add new methods on the RTPS entities, so the ABI break will be there anyways

EduPonz: I agree, I just won't write it here until they happen. This PR does not entail an ABI break, so I won't state it yet.

JLBuenoLopez-eProsima: The statement saying that `implies ABI break` has been removed in 19ab1b3d01118aff59fdfc5e958b494b2659a729

MiguelCompany (approved):

PR #1804, opened 2021-03-04
Expose get_unread_count on DataReader [10756]

MiguelCompany:
The old `Subscriber` API has a `get_unread_count` method. This PR adds the equivalent on DataReader along with checks for it on the DataReader unit tests.


EduPonz (approved):
LGTM with green CI

EduPonz (approved):

Issue #33, opened 2016-03-31
publisher history with KEEP_ALL_HISTORY_QOS never removes samples


liamstask:
I noticed a problem in `PublisherHistory.cpp` similar to the one updated in https://github.com/eProsima/Fast-RTPS/commit/4ce3e29fe0371c00905074d2a06a74be595fd3ec - `m_isHistoryFull` is never cleared to false when samples are removed.

I made corresponding changes to PublisherHistory as in 4ce3e29fe0371c009050 to clear `m_isHistoryFull` when removing changes, but when `m_historyQos.kind == KEEP_ALL_HISTORY_QOS` I never see the changes get removed, even once they have been delivered to the single matched reader.

The sample code is very similar to [this gist](https://gist.github.com/liamstask/b1c47a93800be9795697756dcfd00fec) (from #31) but the publisher history kind is changed to KEEP_ALL_HISTORY_QOS.

liamstask:
also, this expectation is based on the following excerpt from the spec: "On the publishing side, the Service will attempt to keep all samples (representing each value written) of each instance of data (identified by its key) managed by the DataWriter until they can be delivered to all subscribers."

The problem is that when `m_historyQos.kind == KEEP_ALL_HISTORY_QOS`, a publisher is considered full when `m_changes.size() == m_resourceLimitsQos.max_samples`. At that point, if no samples are ever removed, it’s no longer possible to write new samples through that publisher.

maybe there's another approach?

liamstask:
thanks - will give this a shot!

PR #1807, opened 2021-03-05
Receive const pointers on delete_xxx methods [10764]

MiguelCompany:


JLBuenoLopez-eProsima (approved):
LGTM.

CI results seem to be already known flaky tests. Discovery server test \#33 is the first time that fails and I am unable to reproduce it locally, so it seems to also be flaky.

Issue #1819, opened 2021-03-10
Segmentation fault in RTPSDomainImpl::find_local_reader [10842]


AtsushiMorinaga:
# Segmentation fault in RTPSDomainImpl::find_local_reader

Callstack: 

```
Stack trace (most recent call last) in thread 29:
#10   Object "[0xffffffffffffffff]", at 0xffffffffffffffff, in 
#9    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7f4a124874ce, in clone
#8    Object "/lib/x86_64-linux-gnu/libpthread.so.0", at 0x7f4a128bafa2, in 
#7    Object "/usr/lib/x86_64-linux-gnu/libstdc++.so.6", at 0x7f4a127a7b2e, in 
#6    Object "/usr/local/lib/libfastrtps.so.0", at 0x7f4a12b64a26, in eprosima::fastrtps::rtps::ResourceEvent::event_service()
#5    Object "/usr/local/lib/libfastrtps.so.0", at 0x7f4a12b6452b, in eprosima::fastrtps::rtps::ResourceEvent::do_timer_actions()
#4    Object "/usr/local/lib/libfastrtps.so.0", at 0x7f4a12b65a3f, in eprosima::fastrtps::rtps::TimedEventImpl::trigger(std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long, std::ratio<1l, 1000000000l> > >)
#3    Object "/usr/local/lib/libfastrtps.so.0", at 0x7f4a12b7ca64, in 
#2    Object "/usr/local/lib/libfastrtps.so.0", at 0x7f4a12b72bc3, in eprosima::fastrtps::rtps::StatefulWriter::intraprocess_heartbeat(eprosima::fastrtps::rtps::ReaderProxy*, bool)
#1    Object "/usr/local/lib/libfastrtps.so.0", at 0x7f4a12bff2cb, in eprosima::fastrtps::rtps::RTPSDomainImpl::find_local_reader(eprosima::fastrtps::rtps::GUID_t const&)
#0    Object "/usr/local/lib/libfastrtps.so.0", at 0x7f4a12bf4fb7, in eprosima::fastrtps::rtps::RTPSParticipantImpl::find_local_reader(eprosima::fastrtps::rtps::GUID_t const&)
[...]

PR #1823, opened 2021-03-10
Move RTPSDomain mutex lock to callers of find_local_participant [10843]

MiguelCompany:
This prevents the race detected on #1819 by ensuring that the `RTPSDomain` mutex is taken when a new entry is added to `RTPSParticipant::m_allReaderList` or `RTPSParticipant::m_allWriterList`

### DIFF
--- a/src/cpp/rtps/RTPSDomain.cpp
+++ b/src/cpp/rtps/RTPSDomain.cpp
@@ -434,12 +434,13 @@ RTPSParticipantImpl* RTPSDomainImpl::find_local_participant(
 RTPSReader* RTPSDomainImpl::find_local_reader(
         const GUID_t& reader_guid)
 {
-    std::lock_guard<std::mutex> guard(RTPSDomain::m_mutex);
+    std::unique_lock<std::mutex> lock(RTPSDomain::m_mutex);
     for (const RTPSDomain::t_p_RTPSParticipant& participant : RTPSDomain::m_RTPSParticipants)
     {
         if (participant.second->getGuid().guidPrefix == reader_guid.guidPrefix)
         {
             // Participant found, forward the query
+            lock.unlock();
             return participant.second->find_local_reader(reader_guid);
         }
     }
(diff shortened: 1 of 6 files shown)


fujitatomoya (commented):
as far as i see, this makes sense to me.

IkerLuengo (dismissed):
:+1:

src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -434,12 +434,13 @@ RTPSParticipantImpl* RTPSDomainImpl::find_local_participant(
...
         const GUID_t& reader_guid)
 {
-    std::lock_guard<std::mutex> guard(RTPSDomain::m_mutex);
+    std::unique_lock<std::mutex> lock(RTPSDomain::m_mutex);
     for (const RTPSDomain::t_p_RTPSParticipant& participant : RTPSDomain::m_RTPSParticipants)
     {
         if (participant.second->getGuid().guidPrefix == reader_guid.guidPrefix)
         {
             // Participant found, forward the query
+            lock.unlock();
```
VladimirVassiliev, line 443: This lock cannot be released here, as it is protecting m_RTPSParticipants and the participant reference.

src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -450,12 +451,13 @@ RTPSReader* RTPSDomainImpl::find_local_reader(
...
         const GUID_t& writer_guid)
 {
-    std::lock_guard<std::mutex> guard(RTPSDomain::m_mutex);
+    std::unique_lock<std::mutex> lock(RTPSDomain::m_mutex);
     for (const RTPSDomain::t_p_RTPSParticipant& participant : RTPSDomain::m_RTPSParticipants)
     {
         if (participant.second->getGuid().guidPrefix == writer_guid.guidPrefix)
         {
             // Participant found, forward the query
+            lock.unlock();
```
JLBuenoLopez-eProsima, line 460: The same issue also applies here.

JLBuenoLopez-eProsima (changes requested):
There are several build warnings. Some of them I think are due to not being rebased with the latest changes in `master` (for `2.5.0` the test types were updated). However, some of the warnings seems to be introduced by this PR. I suggest rebasing to `master`.

MiguelCompany:
@JLBuenoLopez-eProsima This PR is not correct. What we should do is have another mutex on RTPSDomainParticipantImpl protecting the endpoint collections.

That would be easy to do. The difficult part comes in designing a regression test.

JLBuenoLopez-eProsima:
Close in favor #2506

PR #1826, opened 2021-03-11
Avoid using a custom atomic type for shared pointer data.

mp3guy:
Running [MSAN](https://github.com/google/sanitizers/wiki/MemorySanitizer) on a trivial example results in an instant crash and cryptic stack trace from a debugger:

```
Program received signal SIGSEGV, Segmentation fault.
0x00000000020aa99f in select_compare_exchange_16 () at third-party/gcc/src/libatomic/cas_n16.c:124
124	third-party/gcc/src/libatomic/cas_n16.c: No such file or directory.
(gdb) bt
#0  0x00000000020aa99f in select_compare_exchange_16 () at third-party/gcc/src/libatomic/cas_n16.c:124
#1  0x00007ffff7de04a8 in _dl_relocate_object () from /lib64/ld-linux-x86-64.so.2
#2  0x00007ffff7dd8208 in dl_main () from /lib64/ld-linux-x86-64.so.2
#3  0x00007ffff7dee14f in _dl_sysdep_start () from /lib64/ld-linux-x86-64.so.2
#4  0x00007ffff7dd6136 in _dl_start () from /lib64/ld-linux-x86-64.so.2
#5  0x00007ffff7dd5088 in _start () from /lib64/ld-linux-x86-64.so.2
#6  0x0000000000000001 in ?? ()
#7  0x00007fffffffe0f1 in ?? ()
#8  0x0000000000000000 in ?? ()
```

We narrowed it down to the fact a custom type is being used in a `std::atomic` that is shared between processes. [This question and answer](https://www.quora.com/Can-C++11-std-atomic_*-be-used-across-processes-via-shared-memory) seems to say this is not guaranteed to work, and sure enough, this fix makes the memory sanitizer happy.

### DIFF
--- a/src/cpp/rtps/transport/shared_mem/MultiProducerConsumerRingBuffer.hpp
+++ b/src/cpp/rtps/transport/shared_mem/MultiProducerConsumerRingBuffer.hpp
@@ -88,7 +88,7 @@ public:
             auto pointer = buffer_.node_->pointer_.load(std::memory_order_relaxed);
 
             // If local read_pointer and write_pointer are equal => buffer is empty for this listener
-            if (read_p_ == pointer.write_p )
+            if (read_p_ == pointer.ptr.write_p )
             {
                 return nullptr;
             }
@@ -123,7 +123,7 @@ public:
                 // Increase the free cells => increase the global read pointer
                 auto pointer = buffer_.node_->pointer_.load(std::memory_order_relaxed);
                 while (!buffer_.node_->pointer_.compare_exchange_weak(pointer,
-                        { pointer.write_p, pointer.free_cells + 1 },
+                        { { pointer.ptr.write_p, pointer.ptr.free_cells + 1 } },
                         std::memory_order_release,
                         std::memory_order_relaxed))
                 {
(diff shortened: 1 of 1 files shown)


MiguelCompany (approved):
The changes look good to me. We have an issue with the CI infrastructure, so I will mark this as ready, and will merge after CI has been run

MiguelCompany:

MiguelCompany (changes requested):
@mp3guy Please fix [linter errors](http://jenkins.eprosima.com:8080/job/fastdds_github_uncrustify/2009/)

mp3guy:
Sure thing! That link doesn't work for me right now, is there a way I can run the linter locally?

EduPonz:
Hi @mp3guy ,

> Sure thing! That link doesn't work for me right now, is there a way I can run the linter locally?

Sure! we use uncrustify, the config file for it is hosted [here](https://github.com/eProsima/cpp-style/blob/master/uncrustify.cfg)

mp3guy:
Thanks, should be good now.

MiguelCompany (changes requested):
Please fix [build errors](http://jenkins.eprosima.com:8080/job/fastdds_linux/135/gcc/)

mp3guy:
Huh weird, I couldn't repro those build errors locally. Should be addressed now.

MiguelCompany (dismissed):
LGTM with green CI.

MiguelCompany (approved):
LGTM.

All platforms built correctly without warnings.
Failed tests are known to be flaky.

MiguelCompany:
Going in. Thank you @mp3guy !

PR #1828, opened 2021-03-12
[10715] Update idl to include the Physical Data topic and type

MiguelBarro:
Include a new topic with it's corresponding type:
```idl
struct PhysicalData
{
    @Key detail::GUID_s participant_guid;
    string host;
    string user;
    string process;
};
```


JLBuenoLopez-eProsima (approved):
LGTM

PR #1834, opened 2021-03-16
Fix asan and undefined behavior error [10916]

princeward:
Fix two error reported by AddressSanitizer and UndefinedBehaviorSanitizer.

The AddressSanitizer error is trivial.

For the UndefinedBehaviorSanitizer error, left shifting a uint64 by 64 bits is considered undefined behavior. See this [article](https://stackoverflow.com/questions/10499104/is-shifting-more-than-32-bits-of-a-uint64-t-integer-on-an-x86-machine-undefined) for reference.

### DIFF
--- a/include/fastdds/rtps/common/SerializedPayload.h
+++ b/include/fastdds/rtps/common/SerializedPayload.h
@@ -124,9 +124,9 @@ struct RTPS_DllAPI SerializedPayload_t
             }
         }
         encapsulation = serData->encapsulation;
-        if(!data || !serData->data)
+        if (length == 0)
         {
-            return false;
+            return true;
         }
         memcpy(data, serData->data, length);
         return true;
--- a/include/fastdds/rtps/common/SerializedPayload.h
+++ b/include/fastdds/rtps/common/SerializedPayload.h
@@ -124,6 +124,10 @@ struct RTPS_DllAPI SerializedPayload_t
             }
         }
         encapsulation = serData->encapsulation;
+        if(!data || !serData->data)
+        {
+            return false;
+        }
         memcpy(data, serData->data, length);
         return true;
     }
--- a/src/cpp/fastdds/publisher/qos/DataWriterQos.cpp
+++ b/src/cpp/fastdds/publisher/qos/DataWriterQos.cpp
@@ -60,9 +60,9 @@ WriterQos DataWriterQos::get_writerqos(
     {
         uint64_t id = 0;
         Host::uint48 mac_id = Host::instance().mac_id();
-        for (size_t i = 0; i < Host::mac_id_length; ++i)
+        for (size_t i = 1; i < Host::mac_id_length; ++i)
         {
-            id |= mac_id.value[i] << (64 - i);
+            id |= static_cast<uint64_t>(mac_id.value[i]) << (64 - i);
         }
         qos.data_sharing.add_domain_id(id);
     }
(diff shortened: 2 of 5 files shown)


include/fastdds/rtps/common/SerializedPayload.h
```diff
@@ -124,6 +124,10 @@ struct RTPS_DllAPI SerializedPayload_t
             }
         }
         encapsulation = serData->encapsulation;
+        if(!data || !serData->data)
+        {
+            return false;
+        }
```
MiguelCompany, line 130: Those pointers can only be `nullptr` if the length of the payload is 0. Otherwise, a bad_alloc is thrown. When the length is 0, it should return `true`, as the copy has been successful.

```suggestion
        if (length == 0)
        {
            return true;
        }
```

src/cpp/fastdds/publisher/qos/DataWriterQos.cpp
```diff
@@ -60,9 +60,9 @@ WriterQos DataWriterQos::get_writerqos(
     {
         uint64_t id = 0;
         Host::uint48 mac_id = Host::instance().mac_id();
-        for (size_t i = 0; i < Host::mac_id_length; ++i)
+        for (size_t i = 1; i < Host::mac_id_length; ++i)
```
MiguelCompany, line 63: Looking at this, I think the code was completely wrong in the first place. The intention here was to convert the 6-octet array on `mac_id` to a `uint64_t` with the lowest 16-bits having a value of 0.

```suggestion
        for (size_t i = 0; i < Host::mac_id_length; ++i)
```

src/cpp/fastdds/publisher/qos/DataWriterQos.cpp
```diff
@@ -60,9 +60,9 @@ WriterQos DataWriterQos::get_writerqos(
...
         {
-            id |= mac_id.value[i] << (64 - i);
+            id |= static_cast<uint64_t>(mac_id.value[i]) << (64 - i);
```
MiguelCompany, line 65:
```suggestion
            id |= static_cast<uint64_t>(mac_id.value[i]) << (63 - (i * 8));
```

src/cpp/fastdds/subscriber/qos/DataReaderQos.cpp
```diff
@@ -56,9 +56,9 @@ ReaderQos DataReaderQos::get_readerqos(
     {
         uint64_t id = 0;
         Host::uint48 mac_id = Host::instance().mac_id();
-        for (size_t i = 0; i < Host::mac_id_length; ++i)
+        for (size_t i = 1; i < Host::mac_id_length; ++i)
```
MiguelCompany, line 59:
```suggestion
        for (size_t i = 0; i < Host::mac_id_length; ++i)
```

src/cpp/fastdds/subscriber/qos/DataReaderQos.cpp
```diff
@@ -56,9 +56,9 @@ ReaderQos DataReaderQos::get_readerqos(
...
         {
-            id |= mac_id.value[i] << (64 - i);
+            id |= static_cast<uint64_t>(mac_id.value[i]) << (64 - i);
```
MiguelCompany, line 61:
```suggestion
            id |= static_cast<uint64_t>(mac_id.value[i]) << (63 - (i * 8));;
```

MiguelCompany (changes requested):
Aside from my comments below, please fix DCO

princeward:
I messed up with the DCO signoff. Now closing this one and move to a new PR instead: https://github.com/eProsima/Fast-DDS/pull/1845

PR #1836, opened 2021-03-17
[10715] Hotfix: remove case NONE from IDL

JLBuenoLopez-eProsima:
GCC issues a warning due to this case that does not have any data type associated.


MiguelBarro (approved):
LGTM

PR #1837, opened 2021-03-17
[10500] Upgrade persistence database to v3.0

MiguelBarro:
The new database schema solves LivesSpam and RPC issues.
The Upgrading procedure is updated.
New tests are added to check database migration.

### DIFF
--- a/src/cpp/rtps/persistence/SQLite3PersistenceService.cpp
+++ b/src/cpp/rtps/persistence/SQLite3PersistenceService.cpp
@@ -313,7 +313,7 @@ bool SQLite3PersistenceService::load_writer_from_storage(
             }
 
             // timestamp
-            change->sourceTimestamp = Time_t(sqlite3_column_double(load_writer_stmt_, 5));
+            change->sourceTimestamp.from_ns(sqlite3_column_int64(load_writer_stmt_, 5));
 
             changes.insert(changes.begin(), change);
         }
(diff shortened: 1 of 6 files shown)


src/cpp/rtps/persistence/SQLite3PersistenceService.cpp
```diff
@@ -284,6 +300,21 @@ bool SQLite3PersistenceService::load_writer_from_storage(
...
+                istringstream is(string(reinterpret_cast<const char*>(sqlite3_column_text(load_writer_stmt_, 3))));
+                auto& si = change->write_params.related_sample_identity();
+                is >> si.writer_guid();
+                // Sequence Number
+                SequenceNumber_t rsn(sqlite3_column_int64(load_writer_stmt_, 4));
+                si.sequence_number(rsn);
+            }
+
+            // timestamp
+            change->sourceTimestamp = Time_t(sqlite3_column_double(load_writer_stmt_, 5));
```
richiware, line 316: Using `double` to store the sourcetimestap sure there will be loss of precision. I recomend to store the values of `rtps::Time_t`: seconds and nanoseconds.

richiware (changes requested):

src/cpp/rtps/persistence/SQLite3PersistenceService.cpp
MiguelBarro: In order to avoid precision loss the *timestamps* are now stored as `int64_t`.

richiware (approved):

PR #1846, opened 2021-03-23
[10714] Statistic module FASTDDS_STATISTICS environment variable

JLBuenoLopez-eProsima:


include/fastdds/dds/domain/DomainParticipantFactory.hpp
```diff
@@ -242,6 +242,12 @@ class DomainParticipantFactory
     void participant_has_been_deleted(
             DomainParticipantImpl* part);
 
+    /**
+     * Load statistics profiles from environment variable
+     * @return string with the semicolon-separated list of statistics topic names to be enabled
+     */
+    const std::string& load_statistics_profiles_from_env();
```
MiguelCompany, line 249: I think this should be on a private header / source. As it is only going to be used on the creation of a participant, I propose to add it on `DomainParticipantFactory.cpp`. It could then be a file static method.

include/fastdds/statistics/dds/domain/DomainParticipant.hpp
```diff
@@ -89,6 +89,13 @@ class DomainParticipant : public eprosima::fastdds::dds::DomainParticipant
             const eprosima::fastdds::dds::DomainParticipant* domain_participant);
 };
 
+/* Environment variable to specify a semicolon-separated list of topic names that define the statistics DataWriters that
+ * the DomainParticipant will enable when created.
```
MiguelCompany, line 93: Are the statistics data writers going to be created if the participant is created disabled?

```suggestion
 * the DomainParticipant will enable when enabled.
```

src/cpp/fastdds/domain/DomainParticipantFactory.cpp
```diff
@@ -426,6 +426,18 @@ void DomainParticipantFactory::participant_has_been_deleted(
     }
 }
 
+const std::string& DomainParticipantFactory::load_statistics_profiles_from_env()
```
MiguelCompany, line 429: If my first comment is addressed this could be changed like this:

```suggestion
static void load_statistics_profiles_from_env(
        std::string& topics)
```

MiguelCompany (changes requested):

include/fastdds/statistics/dds/domain/DomainParticipant.hpp
JLBuenoLopez-eProsima: Good question! Probably it makes sense that the implementation of the statistics endpoints creation is done only when the participant is enabled.

MiguelCompany (approved):
Might be a good idea to add a test to the `statistics::DomainParticipant` tests to check the behavior of `create_participant` when there are valid and/or invalid names on the environment variable.

JLBuenoLopez-eProsima:
Rebased and changed base branch

JLBuenoLopez-eProsima:
@MiguelCompany,

Suggested test has been added in d5d8139027feee173f17aa68e3ad99ebb9853ec5

JLBuenoLopez-eProsima:
Closed in favor of #1885

PR #1752, opened 2021-02-11
Do not fail ProxyData deserialization when TypeInformation deserialization fails [10577]

MiguelCompany:
This PR adds unit tests for builtin data (de)serialization, and allows matching with endpoints when the TypeInformation has an unexpected format.

Depends on eProsima/Fast-CDR#104

PR #1847, opened 2021-03-23
Do not fail ProxyData deserialization when TypeObject deserialization fails [10960]

MiguelCompany:
Similar to #1752 for TypeObject instead of TypeInformation


JLBuenoLopez-eProsima (approved):
LGTM

MiguelCompany:
Unrelated failures

PR #1862, opened 2021-03-26
Fix heartbeat and ACKs issues [11003]

MiguelCompany:
This PR fixes two issues:

* Too many piggyback heartbeats were being sent when a SHM transport is registered
* ACKs were not correctly delivered on certain corner cases

### DIFF
--- a/src/cpp/rtps/network/ReceiverResource.cpp
+++ b/src/cpp/rtps/network/ReceiverResource.cpp
@@ -51,7 +51,7 @@ ReceiverResource::ReceiverResource(
             };
     LocatorMapsToManagedChannel = [&transport, locator](const Locator_t& locatorToCheck) -> bool
             {
-                return transport.DoInputLocatorsMatch(locator, locatorToCheck);
+                return locator.kind == locatorToCheck.kind && transport.DoInputLocatorsMatch(locator, locatorToCheck);
             };
 }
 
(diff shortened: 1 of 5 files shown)


src/cpp/rtps/network/ReceiverResource.cpp
```diff
@@ -45,12 +45,18 @@ ReceiverResource::ReceiverResource(
...
     // Implementation functions are bound to the right transport parameters
-    Cleanup = [&transport, locator]() { transport.CloseInputChannel(locator); };
+    Cleanup = [&transport, locator]()
+            {
+                transport.CloseInputChannel(locator);
+            };
     LocatorMapsToManagedChannel = [&transport, locator](const Locator_t& locatorToCheck) -> bool
-    { return transport.DoInputLocatorsMatch(locator, locatorToCheck); };
+            {
+                return locator.kind == locatorToCheck.kind && transport.DoInputLocatorsMatch(locator, locatorToCheck);
```
JLBuenoLopez-eProsima, line 54: Only a nitpick. Shared memory transport implementation of `DoInputLocatorsMatch` already checks that the locators `kind` is equal. Maybe it would be better to remove this check from this specific transport implementation so it is not checked twice.

JLBuenoLopez-eProsima (approved):
LGTM. Only a nitpick that can be disregarded.

src/cpp/rtps/network/ReceiverResource.cpp
MiguelCompany: This is not used much (only on endpoint matching), so I think it does no harm to perform the check twice.

PR #1866, opened 2021-03-26
Fix heartbeat and ACKs issues <1.8.x> [11007]

MiguelCompany:
Backport #1862 to `1.8.x`


JLBuenoLopez-eProsima (changes requested):
Please, fix code style. I suggest changing the label to `no-test` instead of `skip-ci` so the building is done.

JLBuenoLopez-eProsima (approved):
LGTM

Build errors are unrelated to this PR.

PR #1868, opened 2021-03-26
Fix C++ 20 issue with explicit use of operator

### LINKED ISSUE #1867: Infinite recursion in equality operator in ReturnCode_t using c++20
From this commit https://github.com/eProsima/Fast-DDS/commit/99f7b6dd9d67ba1edd404d63ae515dfcade38b7a there is a bug when compiling a project using c++20. The problem is caused by an infinite recursion in the == operator of ReturnCode_t.

c++20 introduces changes in the operator candidates generation and the resolution of ambiguities in them. The change that is causing this error is the generation of a rewritten candidate with the parameters reversed. [over.match#oper-3.4.4](https://timsong-cpp.github.io/cppwp/over.match#oper-3.4.4). This change also means that the expression ``` a == b ``` can be either ``` a.operator==(b)``` or ```b.operator==(a) ``` depending on which is better for the compiler. In this particular case the templated operator is better than the class member operator because there are no implicit type conversions.

https://github.com/eProsima/Fast-DDS/blob/b7e76b2012df1d4133c04f56faf664031cf266d7/include/fastrtps/types/TypesBase.h#L247-L253

You can replicate the issue with this code:
``` c++
#include <fastrtps/types/TypesBase.h>
#include <iostream>

using namespace eprosima::fastrtps::types;

int main() {

    std::cout << std::boolalpha << (ReturnCode_t(0) == ReturnCode_t::RETCODE_OK) << std::endl;
    return 0;
}
```

Compiling with ``` c++ -std=c++20 file.cpp -lfastrtps ```. With ``` -std=c++17 ``` it works fine.

Possible soluions for this issue are adding conditional compilation to the templated operators so in c++20 they are not compiled (They are not needed because of the operator generation), using the class member operator ```b.operator==(a)``` or using the call operator ``` b() == a ```.
## Expected Behavior
The comparison runs correctly.

## Current Behavior
Infinite recursion leading to a segfault.

## Steps to Reproduce

1. Copy the code snippet.
2. Compile with std=c++20 flag and run.

## System information
- Fast-RTPS version: master

MiguelCompany:
This should fix #1867

### DIFF
--- a/include/fastrtps/types/TypesBase.h
+++ b/include/fastrtps/types/TypesBase.h
@@ -249,7 +249,7 @@ operator ==(
         T a,
         const ReturnCode_t& b)
 {
-    return b == a;
+    return b.operator ==(a);
 }
 
 template<class T>
@@ -258,7 +258,7 @@ operator !=(
         T a,
         const ReturnCode_t& b)
 {
-    return b != a;
+    return b.operator !=(a);
 }
 
 // TODO Remove this alias when Fast-RTPS reaches version 2
--- a/include/fastrtps/types/TypesBase.h
+++ b/include/fastrtps/types/TypesBase.h
@@ -249,7 +249,8 @@ operator ==(
         T a,
         const ReturnCode_t& b)
 {
-    return b.operator ==(a);
+    return b.operator ==(
+        a);
 }
 
 template<class T>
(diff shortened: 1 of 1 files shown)


IkerLuengo (approved):
LGTM

MiguelBarro (approved):
LGTM

PR #1873, opened 2021-03-29
Avoid using reference after removing item on LivelinessManager [11019]

### LINKED ISSUE #1871: LivelinessManager remove writer issue [11017]
Hi, I'am working on fastrtps 2.0.x version.

Maybe I found an issue in LivelinessManager.cpp.
In function "LivelinessManager::remove_writer",  the "writer"  will pointer the last object of the vector after calling "writers_.remove(writer)" ,   maybe the removing writer is not the last,  the result is it will notify the wrong guid writer(the remove opertion makes the iterator points to the last object in the vector) in the "writers_" vector by calling "callback_".

Should we put the "writers_.remove(writer)" after calling "callback_"  ?

Thanks a lot.

MiguelCompany:
Port of #1872 to `master`

### DIFF
--- a/src/cpp/rtps/writer/LivelinessManager.cpp
+++ b/src/cpp/rtps/writer/LivelinessManager.cpp
@@ -23,9 +23,9 @@ LivelinessManager::LivelinessManager(
     , timer_(
         service,
         [this]() -> bool
-            {
-                return timer_expired();
-            },
+        {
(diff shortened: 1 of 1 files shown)


MiguelCompany:

JLBuenoLopez-eProsima (approved):
LGTM

PR #1874, opened 2021-03-29
Avoid using reference after removing item on LivelinessManager <2.2.x> [11020]

MiguelCompany:
Port #1872 to `2.2.x`


JLBuenoLopez-eProsima (commented):
LGTM

PR #1875, opened 2021-03-29
Avoid using reference after removing item on LivelinessManager <2.1.x> [11021]

MiguelCompany:
Port #1872 to `2.1.x`


JLBuenoLopez-eProsima (approved):
LGTM

PR #1881, opened 2021-03-31
Fix rpmbuild issues

mtheall:
Fix warnings/errors from rpmbuild


jsantiago-eProsima (changes requested):
Hi @mtheall, thanks for your contribution,

I have some questions regarding some of the changes since I don't have the spec file. I assume you are setting DESTDIR to the staging area for your package but only using it for the tools installation path. Are you not using the same CMAKE_INSTALL_PREFIX for the rest?

jsantiago-eProsima:
Thanks for the thorough explanation @mtheall . Changes look good. If you could you please rebase and fix the CMakeLists conflicts that would be great.

JLBuenoLopez-eProsima:
@jsantiago-eProsima, I have proceed and this is now rebased and the conflicts are fixed. Can you take a look again? Also, looking at the other header files, I suggest changing the permissions to 664 instead of 644, keeping the same criteria. What do you think? If you are OK with that change I will apply it too.

JLBuenoLopez-eProsima:

JLBuenoLopez-eProsima:
Windows failure is unrelated.

JLBuenoLopez-eProsima (approved):

PR #1891, opened 2021-04-07
Fix and expose pull mode

MiguelCompany:
This PR fixes the logic on StatefulWriter when `m_pushMode` is false and exposes it through property `fastdds.push_mode`

Documentation PR on eProsima/Fast-DDS-docs#256


JLBuenoLopez-eProsima (approved):
LGTM!

CI failures seem unrelated to this PR and these tests are known to be flaky.

PR #1909, opened 2021-04-11
Fix crash in dynamic type registration of primitive type array

YangboLong:
Don't add complete type object in type registration for plain types or EK_MINIMAL types. Otherwise, code like below would result in crash because of null TypeIdentifier pointer.
std::vector<uint32_t> lengths = {3};
DynamicType_ptr base_type = DynamicTypeBuilderFactory::get_instance()->create_int32_type();
DynamicTypeBuilder_ptr array = DynamicTypeBuilderFactory::get_instance()->create_array_builder(base_type, lengths);
array->set_name("HelloWorld");
...
Domain::registerDynamicType(mp_participant, &m_DynType);

### DIFF
--- a/src/cpp/fastrtps_deprecated/Domain.cpp
+++ b/src/cpp/fastrtps_deprecated/Domain.cpp
@@ -375,7 +375,10 @@ bool Domain::registerDynamicType(
             // Complete, just to make sure it is generated
             const TypeIdentifier* type_id_complete = typeFactory->get_type_identifier(type->getName(), true);
             const TypeObject* type_obj_complete = typeFactory->get_type_object(type->getName(), true);
-            typeFactory->add_type_object(type->getName(), type_id_complete, type_obj_complete); // Add complete
+            if (type_id_complete && type_obj_complete)
+            {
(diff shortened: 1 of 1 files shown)


YangboLong:
What is the purpose of directory fastrtps_deprecated? I do notice that Domain.cpp sits in this deprecated directory. Does it make sense at all to fix any potential issue in this deprecated directory? Is it going to be replaced as a whole very soon?

JLBuenoLopez-eProsima:
@YangboLong sorry for the long delay in answering. fastrtps_deprecated folder has the old PubSub API that was deprecated with Fast DDS v2.0.0. At some point it will be removed and we encourage every user to use the DDS compliant API instead of the old deprecated one. In fact, the examples using the old API were removed this last year and new features are only implemented in the new API.

Nevertheless, it may be interesting to have this fix even though the API will be removed at some point in the future. I am going to assign a milestone in order to review and merge it.

JLBuenoLopez-eProsima:
Thanks @YangboLong for your contribution and sorry for the late reply. I am going to proceed and close this PR because right now, efforts to improve support for XTYPES v1.3 are being made (https://github.com/eProsima/Fast-DDS/pull/3392) which supersedes this PR.

PR #1922, opened 2021-04-19
Enable/Disable statistics datawriter implementation [10807]

JLBuenoLopez-eProsima:
This PR implements `enable_statistics_datawriter` and `disable_statistics_datawriter` methods.

It also includes new tests to ensure full code line coverage.

[WIP] After the refactor (move auxiliary methods to `eprosima::fastdds::statistics::dds::DomainParticipantImpl`) some tests are failing. Also, including a new test showed a flaw in the implementation logic that has to be adressed (if the topic and type are not consistent with what is expected, `disable_statistics_datawriter` should not delete the datawriter).


src/cpp/fastdds/domain/DomainParticipantImpl.hpp
```diff
@@ -85,7 +85,7 @@ class DomainParticipantImpl
 
 public:
 
-    ReturnCode_t enable();
+    virtual ReturnCode_t enable();
```
EduPonz, line 88: I don't think `virtual` is needed here
```suggestion
    ReturnCode_t enable();
```

EduPonz (changes requested):
Partial

src/cpp/statistics/fastdds/domain/DomainParticipant.cpp
```diff
@@ -78,83 +79,6 @@ const DomainParticipant* DomainParticipant::narrow(
...
-        const std::string& topic_list)
-{
-    // Parse list and call enable_statistics_datawriter
-    std::stringstream topics(topic_list);
-    std::string topic;
-    while (std::getline(topics, topic, ';'))
-    {
-        ReturnCode_t ret = enable_statistics_datawriter(topic, STATISTICS_DATAWRITER_QOS);
-        // case RETCODE_ERROR is checked and logged in enable_statistics_datawriter.
-        // case RETCODE_INCONSISTENT_POLICY cannot happen. STATISTICS_DATAWRITER_QOS is consitent.
```
MiguelBarro, line 132: consistent

src/cpp/statistics/fastdds/domain/DomainParticipant.cpp
```diff
@@ -78,83 +79,6 @@ const DomainParticipant* DomainParticipant::narrow(
...
-    }
-}
-
-void DomainParticipant::delete_statistics_builtin_entities()
-{
-    std::vector<eprosima::fastdds::dds::DataWriter*> builtin_writers;
-    builtin_publisher_->get_datawriters(builtin_writers);
-    for (auto writer : builtin_writers)
-    {
-        std::string topic_name = writer->get_topic()->get_name();
```
MiguelBarro, line 150: The usual practice is to delegate this sort of call to the implementation class where a mutex can prevent concurrency issues.
In this case, several threads running this method may invalidate the `writer` contents and lead to a segmentation fault.

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -0,0 +1,407 @@
...
+    }
+}
+
+void DomainParticipantImpl::delete_statistics_builtin_entities()
+{
+    std::vector<eprosima::fastdds::dds::DataWriter*> builtin_writers;
+    builtin_publisher_->get_datawriters(builtin_writers);
+    for (auto writer : builtin_writers)
+    {
+        std::string topic_name = writer->get_topic()->get_name();
```
MiguelBarro, line 196: This block should be atomic to assure the `writer` variable references a valid object.

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+bool DomainParticipantImpl::check_statistics_topic_name(
+        const std::string& topic) noexcept
+{
+    if (HISTORY_LATENCY_TOPIC != topic && NETWORK_LATENCY_TOPIC != topic && PUBLICATION_THROUGHPUT_TOPIC != topic &&
+            SUBSCRIPTION_THROUGHPUT_TOPIC != topic && RTPS_SENT_TOPIC != topic && RTPS_LOST_TOPIC != topic &&
+            RESENT_DATAS_TOPIC != topic && HEARTBEAT_COUNT_TOPIC != topic && ACKNACK_COUNT_TOPIC != topic &&
+            NACKFRAG_COUNT_TOPIC != topic && GAP_COUNT_TOPIC != topic && DATA_COUNT_TOPIC != topic &&
+            PDP_PACKETS_TOPIC != topic && EDP_PACKETS_TOPIC != topic && DISCOVERY_TOPIC != topic &&
+            SAMPLE_DATAS_TOPIC != topic && PHYSICAL_DATA_TOPIC != topic)
```
MiguelBarro, line 291: I know it's linter friendly but ... 🙄

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -0,0 +1,407 @@
...
+    // delete_topic can fail if the topic is referenced by any other entity. This case could happen even if
+    // it should not. It also fails if topic is a nullptr (dynamic_cast failure).
+    if (ReturnCode_t::RETCODE_OK != delete_topic(topic))
+    {
+        return false;
+    }
+    // unregister_type failures are of no concern here. It will fail if the type is still in use (something
+    // expected) and if the type_name is empty (which is not going to happen).
+    unregister_type(type_name);
+    return true;
```
MiguelBarro, line 401: Our usual failure policy is to try to fulfill as many of the method's tasks as possible. Shouldn't we try to unregister the type even if the topic deletion fails?

src/cpp/fastdds/domain/DomainParticipantImpl.hpp
```diff
@@ -85,7 +85,7 @@ class DomainParticipantImpl
 
 public:
 
-    ReturnCode_t enable();
```
MiguelBarro, line 88: Doxygen?

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.hpp
```diff
@@ -0,0 +1,187 @@
...
+     * @param topic_name Name of the topic associated to the Statistics DataWriter
+     * @return RETCODE_UNSUPPORTED if the FASTDDS_STATISTICS CMake option has not been set,
+     * RETCODE_BAD_PARAMETER if the topic name provided does not correspond to any Statistics DataWriter,
+     * RETCODE_OK if the DataWriter has been correctly deleted or does not exist,
+     * and RETCODE_ERROR otherwise
+     */
+    ReturnCode_t disable_statistics_datawriter(
+            const std::string& topic_name);
+
+    virtual ReturnCode_t enable() override;
```
MiguelBarro, line 77: Doxygen?

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.hpp
```diff
@@ -0,0 +1,187 @@
...
+            const eprosima::fastdds::dds::DomainParticipantQos& qos,
+            eprosima::fastdds::dds::DomainParticipantListener* listen = nullptr)
+        : eprosima::fastdds::dds::DomainParticipantImpl(dp, did, qos, listen)
+        , builtin_publisher_(nullptr)
+    {
+    }
+
+    /**
+     * Auxiliary function to create the statistics builtin entities.
+     * @return true if succesfully created statistics DDS entities,
```
MiguelBarro, line 96: currently it doesn't return anything

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+    friend class eprosima::fastdds::dds::DomainParticipantFactory;
+};
+
+/* Environment variable to specify a semicolon-separated list of topic names that define the statistics DataWriters that
+ * the DomainParticipant will enable when enabled itself.
+ * The topic names must conform to the topic names aliases defined in @ref topic_names.hpp.
+ * For the variable to take any effect the CMake option FASTDDS_STATISTICS must be enabled.
+ */
+const char* const FASTDDS_STATISTICS_ENVIRONMENT_VARIABLE = "FASTDDS_STATISTICS";
```
MiguelBarro, line 178: better as a `constexpr`

MiguelBarro (changes requested):
please address comments

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.cpp
JLBuenoLopez-eProsima: This method is unified with `transform_topic_name_alias` in #1924.

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.cpp
JLBuenoLopez-eProsima: The type is being used by the topic, so if the topic deletion fails, then `unregister_type` would also fail.

src/cpp/statistics/fastdds/domain/DomainParticipant.cpp
JLBuenoLopez-eProsima: This code was already moved to the implementation class.

src/cpp/statistics/fastdds/domain/DomainParticipantImpl.cpp
JLBuenoLopez-eProsima: It is not required for this block to be atomic. `delete_statistics_builtin_entities` is a private method which is only called by `delete_participant`. I have ensured that the `mtx_participants_` mutex is locked before calling this method to prevent concurrency issues.

JLBuenoLopez-eProsima:
@MiguelBarro @EduPonz Your suggestions have been addressed in the latest commit. Thanks for reviewing!

MiguelBarro (approved):

EduPonz (approved):

PR #1931, opened 2021-04-23
Correctly update memory policy on writers and readers [11348]

MiguelCompany:
This PR fixes an assertion on writers with bounded types and `PREALLOCATED_WITH_REALLOC` policy, and also sets `PREALLOCATED` on readers with bounded types and `PREALLOCATED_WITH_REALLOC`


IkerLuengo (approved):
LGTM

PR #1941, opened 2021-04-29
HISTORY2HISTORY_LATENCY test and implementation [10789]

MiguelCompany:


MiguelBarro (changes requested):
The participant doesn't need to instrument the writers in order to notify this callback. Please remove from the mask in:
https://github.com/eProsima/Fast-DDS/blob/8def3913d848b181fe0b499a1ebeb6efd897cb86/src/cpp/statistics/rtps/StatisticsBase.cpp#L138
Add the new callback to the comments in:
https://github.com/eProsima/Fast-DDS/blob/8def3913d848b181fe0b499a1ebeb6efd897cb86/test/unittest/statistics/rtps/RTPSStatisticsTests.cpp#L519-L527
https://github.com/eProsima/Fast-DDS/blob/8def3913d848b181fe0b499a1ebeb6efd897cb86/test/unittest/statistics/rtps/RTPSStatisticsTests.cpp#L652-L657

PR #1944, opened 2021-04-30
Fixes #1484.

### LINKED ISSUE #1484: Segmentation fault caused by sequential Asio calls [9820]
## Expected Behavior
Run HelloWorldExample from Fast-RTPS together with another application that also uses Asio.

## Current Behavior
As soon as I start the HelloWorldExample with `./HelloWorldExample publisher`, the segmentation fault occurs.

## Steps to Reproduce
The error occurs alongside a simple [eCAL](https://github.com/continental/ecal) instance, which also uses Asio for communication.
1. Download and unzip the eCAL build libraries from [installecallibs.zip](https://github.com/eProsima/Fast-DDS/files/5373403/installecallibs.zip).
2. Download the modified HelloWorldExample from [helloWorld.zip](https://github.com/eProsima/Fast-DDS/files/5373422/helloWorld.zip). I've included the executable and the sources to prove that the modifications are minimal.
3. You may have to add the libraries from (1) to the path. Use `export LD_LIBRARY_PATH=pathToECALLibs:pathToFast-RTPSLibs`.
4. Check with ldd that all libraries required to run HelloWorldExample are available.
5. Execute `./HelloWorldExample publisher`.

## System information
- Fast-RTPS version: v1.8,4 with Asio v1.12.2
- OS: Ubuntu 20.04.1, QNX

## Additional context
Analyzing the core dump looks like an issue with Asio. I tried the exactly same configuration but building Fast-RTPS with Asio v1.13.0 (only one version later than status-quo) and everything works fine. In fact, I tried several versions of Asio, and all of them worked. It looks like the bug was definitely fixed in v1.13.0.

pablorcum:
Asio submodule updated to tag 1.18.1.
Added define ASIO_DISABLE_VISIBILITY.

Signed-off-by: promero <promero@mathworks.com>


MiguelCompany (approved):
LGTM. Let's see the CI results.

MiguelCompany:
@pablorcum Just one question. Will this work when Fast DDS is built on a system with a current installation of asio?
I see that the package version of `libasio-dev` on the Ubuntu apt repositories is `1.12.2` up to Groovy. On Hirsute it changes to `1.18.1`

Will `ASIO_DISABLE_VISIBILITY` make any harm on `1.12.2`, or will it just be ignored?

pablorcum:
@MiguelCompany Yes, it will work in the same way as before this commit. Updating the submodule has no impact in using an existing installation of asio, and the new define flag will be ignored.

In fact, I've just locally tested to build the project with an additional dummy define that does not exist, and did not have any impact. Therefore, old asio versions that do not support `ASIO_DISABLE_VISIBILITY` will keep working as today.

MiguelCompany:

MiguelCompany:
CI seems ok. Only known flakey failures.
Thank you @pablorcum for your contribution.

PR #1945, opened 2021-04-30
Collection of data and notification to listeners SAMPLE_DATAS [10805]

MiguelCompany:


src/cpp/rtps/writer/StatefulWriter.cpp
```diff
@@ -61,6 +61,17 @@ namespace fastrtps {
 namespace rtps {
 
 
+static void add_statistics_sent_submessage(
```
MiguelBarro, line 64: I think this method should be either:
- a static method of `RTPSWriter` to avoid a StatelessWriter redefinition of the function.
- a setter in the `CacheChange_t` struct.

src/cpp/rtps/writer/StatelessWriter.cpp
```diff
@@ -46,6 +46,18 @@ namespace fastrtps {
...
+        size_t num_locators)
+{
+    static_cast<void>(change);
+    static_cast<void>(num_locators);
+
+#ifdef FASTDDS_STATISTICS
+    change->num_sent_submessages += num_locators;
+#endif // ifdef FASTDDS_STATISTICS
+}
+
```
MiguelBarro, line 60: See related comment in `StatefulWriter`.

test/unittest/statistics/rtps/RTPSStatisticsTests.cpp
```diff
@@ -81,7 +87,11 @@ struct MockListener : IListener
             case DISCOVERED_ENTITY:
                 on_entity_discovery(data.discovery_time());
                 break;
+            case SAMPLE_DATAS:
+                on_sample_datas(data.sample_identity_count());
+                break;
             default:
+                on_unexpected_kind(kind);
```
MiguelBarro, line 94: That's a good one. This will generate a warning if gtest catches a callback with unknown discriminant...should we promote it to an error? 🤔

MiguelBarro (changes requested):

test/unittest/statistics/rtps/RTPSStatisticsTests.cpp
MiguelCompany: Combined with the expectation on the constructor, this makes the test fail if `on_unexpected_kind` is ever called. So it is already an error.

MiguelBarro: Didn't notice that 👍

src/cpp/rtps/writer/StatefulWriter.cpp
MiguelCompany: I made both `add_statistics_sent_submessage` and `send_data_or_fragments` part of RTPSWriter.

MiguelBarro (approved):
LGTM

PR #1947, opened 2021-04-30
Fix partition copy in QoS [11451]

jay-sridharan:
This PR fixes a bug where a QoS is copied from one ReaderProxyData to another but partition names are not copied if the list is empty. 

Specifically, this code path: 

```
/// src/cpp/rtps/builtin/discovery/participant/PDP.cpp   

         else
            {
                // Pool is not empty, use entry from pool
                ret_val = reader_proxies_pool_.back();
                reader_proxies_pool_.pop_back();
            }

            // Add to ParticipantProxyData
            (*pit->m_readers)[reader_guid.entityId] = ret_val;
```

### DIFF
--- a/src/cpp/fastdds/publisher/qos/WriterQos.cpp
+++ b/src/cpp/fastdds/publisher/qos/WriterQos.cpp
@@ -95,7 +95,7 @@ void WriterQos::setQos(
         m_presentation = qos.m_presentation;
         m_presentation.hasChanged = true;
     }
-    if (first_time || qos.m_partition.names().size() > 0)
+    if (first_time || qos.m_partition.names() != m_partition.names())
     {
         m_partition = qos.m_partition;
         m_partition.hasChanged = true;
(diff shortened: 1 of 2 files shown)


jay-sridharan:
@MiguelCompany

Hi Miguel, 

I'm pretty new to this project, so please forgive the newbie questions. Do I need to assign reviewers to this somehow, or does that happen automatically? And should I worry about the CI failures? Seems like quite a few other PRs have failed CI as well.

MiguelCompany:
@jay-sridharan First of all, thank you for your contribution. The process is that someone at eProsima reviews PRs, we will do it with yours soon. The CI failures seem unrelated to your changes, but checking them is also part of the review process.

MiguelCompany (dismissed):
The changes look good to me, but it would be great to have a regression test in place.
Would you mind adding one?

Also, while trying to find the root cause of the issue I've also found that [these lines](https://github.com/jay-sridharan/Fast-DDS/blob/fd802f392dc4a79034b2bca21f79947032b8c716/src/cpp/rtps/builtin/discovery/participant/PDP.cpp#L995-L1005) need a `pit.second->clear()` before the corresponding `push_back`

jay-sridharan:
Thanks for the review! I'll look into adding a regression test. Any thoughts off the top of your head as to the best way to do that?

JLBuenoLopez-eProsima:
I have rebased and apply the review suggestion. Part of the changes suggested in this PR were already introduced in #2108. I have followed the same approach in the publisher's side.

JLBuenoLopez-eProsima:

EduPonz (approved):

PR #1946, opened 2021-04-30
[10802] Collection of data and notification to listeners PDP_PACKETS

MiguelBarro:
There is a little piggyback preventing the dds layer statistics endpoints from been instrumented from the rtps layer.

PR #1948, opened 2021-05-04
[10803] Collection of data and notification to listeners EDP_PACKETS

MiguelBarro:
It must be merged after https://github.com/eProsima/Fast-DDS/pull/1946

### DIFF
--- a/include/fastdds/rtps/common/EntityId_t.hpp
+++ b/include/fastdds/rtps/common/EntityId_t.hpp
@@ -154,6 +154,24 @@ struct RTPS_DllAPI EntityId_t
 
 #endif // if !FASTDDS_IS_BIG_ENDIAN_TARGET
 
+    /*!
+     * @brief conversion to uint32_t
+     * @return uint32_t representation
+     */
+    uint32_t to_uint32() const
+    {
+        uint32_t res = *reinterpret_cast<const uint32_t*>(value);
+
+#if !FASTDDS_IS_BIG_ENDIAN_TARGET
+        res = ( res >> 24 ) |
+              (0x0000ff00 & ( res >> 8)) |
+              (0x00ff0000 & ( res << 8) ) |
+              ( res << 24 );
+#endif // if !FASTDDS_IS_BIG_ENDIAN_TARGET
+
+        return res;
+    }
+
     static EntityId_t unknown()
     {
         return EntityId_t();
--- a/include/fastdds/rtps/common/EntityId_t.hpp
+++ b/include/fastdds/rtps/common/EntityId_t.hpp
@@ -164,9 +164,9 @@ struct RTPS_DllAPI EntityId_t
 
 #if !FASTDDS_IS_BIG_ENDIAN_TARGET
         res = ( res >> 24 ) |
-              (0x0000ff00 & ( res >> 8)) |
-              (0x00ff0000 & ( res << 8) ) |
-              ( res << 24 );
+                (0x0000ff00 & ( res >> 8)) |
+                (0x00ff0000 & ( res << 8)) |
+                ( res << 24 );
 #endif // if !FASTDDS_IS_BIG_ENDIAN_TARGET
 
         return res;
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.h
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.h
@@ -295,7 +295,7 @@ public:
                 msg->length);
 
             // checkout if sender is a pdp endpoint
-            on_pdp_packet(
+            on_dp_packet(
                 sender_guid,
                 destination_locators_begin,
                 destination_locators_end);
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.h
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.h
@@ -294,8 +294,8 @@ public:
                 destination_locators_end,
                 msg->length);
 
-            // checkout if sender is a pdp endpoint
-            on_dp_packet(
+            // checkout if sender is a discovery endpoint
+            on_discovery_packet(
                 sender_guid,
                 destination_locators_begin,
                 destination_locators_end);
--- a/src/cpp/statistics/rtps/StatisticsBase.cpp
+++ b/src/cpp/statistics/rtps/StatisticsBase.cpp
@@ -377,7 +377,7 @@ void StatisticsParticipantImpl::on_edp_packet(
         const fastrtps::rtps::GUID_t& sender)
 {
     // check it's a pdp endpoint
-    switch(sender.entityId.to_uint32())
+    switch (sender.entityId.to_uint32())
     {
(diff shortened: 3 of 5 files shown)


include/fastdds/rtps/common/EntityId_t.hpp
```diff
@@ -154,6 +154,24 @@ struct RTPS_DllAPI EntityId_t
...
+        uint32_t res = *reinterpret_cast<const uint32_t*>(value);
+
+#if !FASTDDS_IS_BIG_ENDIAN_TARGET
+        res = ( res >> 24 ) |
+                (0x0000ff00 & ( res >> 8)) |
+                (0x00ff0000 & ( res << 8)) |
+                ( res << 24 );
+#endif // if !FASTDDS_IS_BIG_ENDIAN_TARGET
+
+        return res;
```
MiguelCompany, line 172: We always want the MSB first representation, so we explicitly use it. Most compilers will reduce this to one or two assembly instructions.

```suggestion
        return (static_cast<uint32_t>(value[0]) << 24) |
                (static_cast<uint32_t>(value[1]) << 16) |
                (static_cast<uint32_t>(value[2]) << 8) |
                (static_cast<uint32_t>(value[3]));
```

src/cpp/rtps/participant/RTPSParticipantImpl.h
```diff
@@ -293,6 +293,11 @@ class RTPSParticipantImpl
                 destination_locators_begin,
                 destination_locators_end,
                 msg->length);
+
+            // checkout if sender is a pdp endpoint
+            on_pdp_packet(sender_guid);
+            // checkout if sender is a edp endpoint
+            on_edp_packet(sender_guid);
```
MiguelCompany, line 300: One packet should be accounted per destination. I would follow an approach similar to the one in on_rtps_send(), passing the locator iterators and incrementing the count once per iterator.

src/cpp/statistics/rtps/StatisticsBase.cpp
```diff
@@ -341,3 +341,67 @@ void StatisticsParticipantImpl::on_entity_discovery(
...
+    for_each_listener([&data](const std::shared_ptr<IListener>& listener)
+            {
+                listener->on_statistics_data(data);
+            });
+}
+
+void StatisticsParticipantImpl::on_edp_packet(
+        const fastrtps::rtps::GUID_t& sender)
+{
+    // check it's a pdp endpoint
```
MiguelCompany, line 377:
```suggestion
    // check it's an edp endpoint
```

src/cpp/statistics/rtps/StatisticsBase.cpp
```diff
@@ -341,3 +341,67 @@ void StatisticsParticipantImpl::on_entity_discovery(
...
+        default:
+            return; // ignore
+    }
+
+    EntityCount notification;
+    notification.guid(to_statistics_type(get_guid()));
+
+    {
+        std::lock_guard<std::recursive_mutex> lock(get_statistics_mutex());
+        notification.count(++edp_counter_);
```
MiguelCompany, line 394: The iterator loop would go here.

MiguelCompany (changes requested):

include/fastdds/rtps/common/EntityId_t.hpp
MiguelBarro: I've already verified with https://godbolt.org/ that is reduced to `bswap eax` in gcc, msvc and clang. Though your approach is less cryptic and I like it.

MiguelBarro: Unfortunately, [msvc](https://godbolt.org/z/47EfE6nb3) doesn't know how to optimize the new approach (gcc and clang do 👍)

MiguelCompany: Ok. Let's leave it the original way then

src/cpp/statistics/rtps/StatisticsBase.hpp
```diff
@@ -282,21 +293,46 @@ class StatisticsParticipantImpl
...
+     * We filtered the non discovery traffic here to minimize presence of statistics code in endpoints implementation.
      * @param sender_guid GUID_t to filter
-     * @param destination_locators_begin, start of locators range
-     * @param destination_locators_end, end of locators range
+     * @param destination_locators_begin start of locators range
+     * @param destination_locators_end end of locators range
      */
     template<class LocatorIteratorT>
-    void on_pdp_packet(
+    void on_dp_packet(
```
MiguelCompany, line 310:
```suggestion
    void on_discovery_packet(
```

src/cpp/statistics/rtps/StatisticsBase.hpp
```diff
@@ -282,21 +293,46 @@ class StatisticsParticipantImpl
...
+            switch (sender_guid.entityId.to_uint32())
+            {
+                case ENTITYID_SPDP_BUILTIN_RTPSParticipant_WRITER:
+                case ENTITYID_SPDP_BUILTIN_RTPSParticipant_READER:
+                    discovery_callback = &StatisticsParticipantImpl::on_pdp_packet;
+                    break;
+                case ENTITYID_SEDP_BUILTIN_PUBLICATIONS_WRITER:
+                case ENTITYID_SEDP_BUILTIN_PUBLICATIONS_READER:
+                case ENTITYID_SEDP_BUILTIN_SUBSCRIPTIONS_WRITER:
+                case ENTITYID_SEDP_BUILTIN_SUBSCRIPTIONS_READER:
```
MiguelCompany, line 329: When building with security, [these ids](https://github.com/eProsima/Fast-DDS/blob/af466cfe63a8319cc9d37514267de8952627a9a4/include/fastdds/rtps/common/EntityId_t.hpp#L55-L58) should also be accounted.

src/cpp/statistics/rtps/StatisticsBase.hpp
```diff
@@ -365,28 +402,27 @@ class StatisticsParticipantImpl
...
-     * We filtered the non-pdp traffic here to minimize presence of statistics code in endpoints implementation.
+     * Report discovery protocols message exchange.
+     * We filtered the non discovery traffic here to minimize presence of statistics code in endpoints implementation.
      * @param GUID_t to filter
      * @param start of locators range
      * @param end of locators range
      */
     template<class LocatorIteratorT>
-    inline void on_pdp_packet(
+    inline void on_dp_packet(
```
MiguelCompany, line 420:
```suggestion
    inline void on_discovery_packet(
```

MiguelCompany (changes requested):

src/cpp/statistics/rtps/StatisticsBase.hpp
MiguelBarro: didn't notice that

MiguelCompany (approved):
LGTM

PR #1955, opened 2021-05-12
[10792] Collection of data and notification to listeners PUBLICATION_THROUGHPUT

MiguelBarro:
It was decided to retake the concept of reporting rate instead of the number of bytes. The change simplifies things and if in the future the amount of bytes is necessary we can extend the **idl**.

The callback will return bytes/second ( payload size / time span since the last callback). We only need to keep the
timestamp of the last callback and initialize it on construction (creation is the first callback).


JLBuenoLopez-eProsima (approved):
LGTM. Only a suggestion to clarify the documentation

JLBuenoLopez-eProsima (approved):
:ok_hand:

PR #1967, opened 2021-05-17
Document and assert RTPS entity pointers lifetime [11579]

EduPonz:
This PR prevents accessing to an already deleted `RTPSParticipant`. This case may arise when an application does:
1. Create an RTPSParticipant an save the pointer
2. Call to `RTPSDomain::stopAll()`
3. Call to `RTPSDomain::removeRTPSParticipant()` with the saved pointer.

`stopAll()` already deleted all the created `RTPSParticipant`s, so accessing to the memory segment using the pointer obtained on creation is dangerous at best. We can not prevent applications to manipulate the pointer after calling `stopAll()` (thus breaking the contract), but we definitely should not do so within Fast DDS.

### DIFF
--- a/src/cpp/rtps/RTPSDomain.cpp
+++ b/src/cpp/rtps/RTPSDomain.cpp
@@ -212,22 +212,22 @@ RTPSParticipant* RTPSDomain::createParticipant(
 bool RTPSDomain::removeRTPSParticipant(
         RTPSParticipant* p)
 {
-    if (p != nullptr)
+    for (auto it = m_RTPSParticipants.begin(); it != m_RTPSParticipants.end(); ++it)
     {
-        p->mp_impl->disable();
-
-        std::unique_lock<std::mutex> lock(m_mutex);
-        for (auto it = m_RTPSParticipants.begin(); it != m_RTPSParticipants.end(); ++it)
+        if (it->first == p)
         {
-            if (it->second->getGuid().guidPrefix == p->getGuid().guidPrefix)
+            if (p->mp_impl != nullptr)
             {
-                RTPSDomain::t_p_RTPSParticipant participant = *it;
-                m_RTPSParticipants.erase(it);
-                m_RTPSParticipantIDs.erase(m_RTPSParticipantIDs.find(participant.second->getRTPSParticipantID()));
-                lock.unlock();
-                removeRTPSParticipant_nts(participant);
-                return true;
+                p->mp_impl->disable();
             }
+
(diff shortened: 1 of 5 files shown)


src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -212,22 +212,22 @@ RTPSParticipant* RTPSDomain::createParticipant(
...
-                RTPSDomain::t_p_RTPSParticipant participant = *it;
-                m_RTPSParticipants.erase(it);
-                m_RTPSParticipantIDs.erase(m_RTPSParticipantIDs.find(participant.second->getRTPSParticipantID()));
-                lock.unlock();
-                removeRTPSParticipant_nts(participant);
-                return true;
+                p->mp_impl->disable();
             }
+
+            std::unique_lock<std::mutex> lock(m_mutex);
```
IkerLuengo, line 224: This lock should be taken before traversing the list of participants, otherwise it may be changed during the processing, invalidating the iterators. However, this will have the effect of calling `ChannelResource::disable` with the mutex locked (during the `disable`). I don't think this is any problem, but be aware of it.

IkerLuengo (changes requested):

src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -237,6 +238,8 @@ bool RTPSDomain::removeRTPSParticipant(
 void RTPSDomain::removeRTPSParticipant_nts(
         RTPSDomain::t_p_RTPSParticipant& participant)
 {
+    // The destructor of RTPSParticipantImpl already deletes the associated RTPSParticipant and sets
+    // its pointter to the RTPSParticipant to nullptr, so  there is no need to do it here manually.
```
MiguelCompany, line 242:
```suggestion
    // its pointer to the RTPSParticipant to nullptr, so there is no need to do it here manually.
```

src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -214,6 +214,7 @@ bool RTPSDomain::removeRTPSParticipant(
 {
     if (p != nullptr)
     {
+        assert((p->mp_impl != nullptr) && "This participant has been previously invalidated");
```
MiguelCompany, line 217: Shouldn't we then add `mp_impl = nullptr` to the destructor of `RTPSParticipant`

MiguelCompany (changes requested):

MiguelCompany (approved):

PR #1969, opened 2021-05-18
[11278] UDPv6 refactor and tests

### LINKED ISSUE #1923: IPv6 Interface Whitelisting Does Not Work [11278]
When creating a UDPv6TransportDescriptor with an interface whitelist, Fast-DDS discovery fails.
When creating the equivalent UDPv4TransportDescriptor, Fast-DDS discovery succeeds.

Looking at #346, a fix was implemented to make this work for IPv4, but the equivalent fix was not made for IPv6.

## Expected Behavior
Using the whitelisted interface, the Fast-DDS discovery process should find all instances on the network.

## Current Behavior
It appears as the the discovery multi-cast group isn't joined on the white-listed interface

## Steps to Reproduce
1. Assign a UDPv6TransportDescriptor with a interfaceWhiteList to the DomainParticipantQos
2. Add a valid metatrafficMulticastLocator, and default_multicast_locator to the Qos
3. Create a participant with the QoS, and publisher/subscriber from that
4. Run multiple instances on the network, and the Fast-DDS apps will not discover each other
5. Remove the interfaceWhiteList, and discovery will work.

## System information
- Fast-RTPS version: v2.3.0
- OS: Ubuntu 18.04
- Network interfaces: 2001::1 (global addressable ipv6 address)

## Additional context
https://github.com/eProsima/Fast-DDS/pull/346

## Additional resources
I hacked in the following code, based on the pull-request made for IPv4, and the interface whitelisting worked for my situation.
```
diff --git a/src/cpp/rtps/transport/UDPv6Transport.cpp b/src/cpp/rtps/transport/UDPv6Transport.cpp
index 5118505dc..c09c70c5d 100644
--- a/src/cpp/rtps/transport/UDPv6Transport.cpp
+++ b/src/cpp/rtps/transport/UDPv6Transport.cpp
@@ -341,6 +341,45 @@ bool UDPv6Transport::OpenInputChannel(

     if (IPLocator::isMulticast(locator) && IsInputChannelOpen(locator))
     {
+#ifndef _WIN32
+        if (!is_interface_whitelist_empty())
+        {
+            std::string locatorAddressStr = IPLocator::toIPv6string(locator);
+            // Either wildcard address or the multicast address needs to be bound on non-windows systems
+            bool found = false;
+
+            // First check if the multicast address is already bound
+            auto& channelResources = mInputSockets.at(IPLocator::getPhysicalPort(locator));
+            for (UDPChannelResource* channelResource : channelResources)
+            {
+                if (channelResource->interface() == locatorAddressStr)
+                {
+                    found = true;
+                    break;
+                }
+            }
+
+            // Create a new resource if no one is found
+            if (!found)
+            {
+                try
+                {
+                    // Bind to multicast address
+                    UDPChannelResource* p_channel_resource;
+                    p_channel_resource = CreateInputChannelResource(locatorAddressStr, locator, true, maxMsgSize,
+                                    receiver);
+                    mInputSockets[IPLocator::getPhysicalPort(locator)].push_back(p_channel_resource);
+                }
+                catch (asio::system_error const& e)
+                {
+                    (void)e;
+                    logWarning(RTPS_MSG_OUT, "UDPTransport Error binding " << locatorAddressStr << " at port: (" << IPLocator::getPhysicalPort(
+                                locator) << ")"
+                                                                           << " with msg: " << e.what());
+                }
+            }
+        }
+#endif // ifndef _WIN32
         // The multicast group will be joined silently, because we do not
         // want to return another resource.
         auto pChannelResources = mInputSockets.at(IPLocator::getPhysicalPort(locator));
```

JLBuenoLopez-eProsima:
This PR updates UDPv6 and includes several tests to check this feature. This PR fixes the whitelist interface feature for UDPv6 (fixes #1923).

There was a bug when comparing IPv6. IPv6 address has an scope that is included after the address separated by `%`. There is a new method that allows the comparison of IPv6 addresses without taking into account the scope. This way the user can simply whitelist using the IPv6 address without having to include the name of the interface so Asio can infer the scope of the given address. It also works if the scope is given.

IPv6 feature is not completely tested so this PR is going to be a draft while some tests are included to ensure that the feature is stable and working in the most usual cases.

### DIFF
--- a/src/cpp/rtps/transport/UDPv4Transport.cpp
+++ b/src/cpp/rtps/transport/UDPv4Transport.cpp
@@ -116,7 +116,7 @@ UDPv4Transport::UDPv4Transport(
 
         if (interface_whitelist_.empty())
         {
-            logError(TRANSPORT, "All whitelist interfaces where filtered out");
+            logError(TRANSPORT, "All whitelist interfaces were filtered out");
             interface_whitelist_.emplace_back(ip::address_v4::from_string("192.0.2.0"));
         }
     }
--- a/src/cpp/rtps/transport/UDPv6Transport.cpp
+++ b/src/cpp/rtps/transport/UDPv6Transport.cpp
@@ -342,22 +342,85 @@ bool UDPv6Transport::OpenInputChannel(
 
     if (IPLocator::isMulticast(locator) && IsInputChannelOpen(locator))
     {
-        // The multicast group will be joined silently, because we do not
-        // want to return another resource.
-        auto pChannelResources = mInputSockets.at(IPLocator::getPhysicalPort(locator));
-        for (auto& channelResource : pChannelResources)
+        std::string locatorAddressStr = IPLocator::toIPv6string(locator);
+        ip::address_v6 locatorAddress = ip::address_v6::from_string(locatorAddressStr);
+
+#ifndef _WIN32
+        if (!is_interface_whitelist_empty())
         {
-            if (channelResource->interface() == s_IPv4AddressAny)
+            //Either wildcard address or the multicast address needs to be bound on non-windows systems
+            bool found = false;
+
+            // First check if the multicast address is already bound
+            auto& channelResources = mInputSockets.at(IPLocator::getPhysicalPort(locator));
+            for (UDPChannelResource* channelResource : channelResources)
             {
-                std::vector<IPFinder::info_IP> locNames;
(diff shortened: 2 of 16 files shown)


MiguelBarro (approved):
LGTM

MiguelCompany (approved):

PR #1991, opened 2021-05-28
WaitSet, Condition and GuardCondition implementation [11608, 11609]

MiguelCompany:


JLBuenoLopez-eProsima (changes requested):
Changes look good. However there are 3 tests that are failing so either the test is not working or the implementation lacks something.

* ConditionTests.waitset_wait
* ConditionTests.guard_condition_methods
* WaitSetImplTests.wait (child aborted)

The first two tests fail in Ubuntu and the third in all platforms (Ubuntu, Windows and Mac)

JLBuenoLopez-eProsima (approved):
LGTM and the tests are now passing

PR #1992, opened 2021-05-31
Add load persistence_guid from UserData [11689]

richiware:
Feature for distributing persistence guid through participant user_data when EDP static is used

### DIFF
--- a/src/cpp/rtps/builtin/discovery/endpoint/EDPStatic.cpp
+++ b/src/cpp/rtps/builtin/discovery/endpoint/EDPStatic.cpp
@@ -214,9 +214,35 @@ bool EDPStatic::removeLocalWriter(
 void EDPStatic::assignRemoteEndpoints(
         const ParticipantProxyData& pdata)
 {
+    GUID_t persistence_guid;
+
+    // Fill persistence GUID if present in UserData.
+    bool is_persistent_guid = (18 <= pdata.m_userData.size()) &&
+            ('V' == pdata.m_userData.at(0)) && ('G' == pdata.m_userData.at(1)) && ('W' == pdata.m_userData.at(2));
+    if (is_persistent_guid)
+    {
+        persistence_guid.guidPrefix.value[0] = pdata.m_userData.at(3);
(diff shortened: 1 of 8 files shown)


MiguelCompany:

src/cpp/rtps/builtin/discovery/endpoint/EDPStatic.cpp
```diff
@@ -214,9 +214,35 @@ bool EDPStatic::removeLocalWriter(
...
+        persistence_guid.guidPrefix.value[11] = pdata.m_userData.at(14);
+        persistence_guid.entityId.value[0] = pdata.m_userData.at(15);
+        persistence_guid.entityId.value[1] = pdata.m_userData.at(16);
+        persistence_guid.entityId.value[2] = pdata.m_userData.at(17);
+    }
+
     for (ParameterPropertyList_t::const_iterator pit = pdata.m_properties.begin();
             pit != pdata.m_properties.end(); ++pit)
     {
+        persistence_guid.entityId.value[3] = 0;
```
MiguelCompany, line 244: value[3] is the `entityKind`, so I think the user id should be assigned to either index 0 or 2

MiguelCompany (changes requested):
1. Shouldn't we have a test for this feature?
2. Fix linters

src/cpp/rtps/builtin/discovery/endpoint/EDPStatic.cpp
richiware: This is actually not an option because will break interoperability

MiguelCompany: As talked offline, the persistence_guid is always set by the user, and doesn't need to follow the standard GUID rules.

richiware:
@MiguelCompany Applied suggestions

MiguelCompany (dismissed):
LGTM

MiguelCompany (approved):
LGTM. Failures do seem unrelated.

PR #1998, opened 2021-06-04
Only export public symbols on non-windows platforms [11754]

### LINKED ISSUE #1994: SQLite conflict [11751]
My application uses Fast-DDS and Qt. It is accessing a SQLite database via Qt, which uses the system-installed libsqlite3.so. There are symbol conflicts where some symbols are resolved to libsqlite3.so and some symbols are resolved to libfastrtps.so, and in a few situations causes my application to crash, presumably due to an ABI mismatch.

## Expected Behavior
Configure Fast-DDS to build using the system-installed or user-provided sqlite3 shared library. Another option would be the ability to configure the embedded sqlite3 to have all hidden symbols.

## Current Behavior
Can only disable or enable sqlite3 support for persistence. Enabled always uses an embedded sqlite3.

## Steps to Reproduce
I don't have a simplified use-case (yet) but here is a sample stack trace:
```
received signal SIGSEGV, Segmentation fault.
__memmove_avx_unaligned_erms () from /lib64/libc.so.6
sqlite3_str_vappendf () from /lib64/libfastrtps.so.2.3
sqlite3_str_appendf () from /lib64/libfastrtps.so.2.3
sqlite3VdbeMemStringify.lto_priv.0 () from /lib64/libfastrtps.so.2.3
valueToText.lto_priv.0 () from /lib64/libfastrtps.so.2.3
columnName () from /lib64/libsqlite3.so.0
QSQLiteResultPrivate::initColumns from /usr/lib64/qt5/plugins/sqldrivers/libqsqlite.so

## System information
The problem has been observed on the following:
Fedora 34
Ubuntu 20.04

- Fast-RTPS version: Fast-DDS 2.3.2
- OS: Fedora 34, Ubuntu 20.04
- Network interfaces: N/A
- ROS2: N/A, using Fast-DDS directly

## Additional context
N/A

## Additional resources
N/A

MiguelCompany:
This should fix #1994 by using [LD version scripts](https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html) to control the list of exported symbols.


JLBuenoLopez-eProsima (approved):
LGTM, though @MiguelBarro knows more about CMake and might see some issue.

MiguelBarro (dismissed):
LGTM

MiguelCompany:
@MiguelBarro CMake documentation on [CheckCXXSourceCompiles](https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html) specifically mentions `CMAKE_REQUIRED_FLAGS`, so that is what I used.

Was I mistaken?

MiguelBarro (approved):
Using CMAKE_REQUESTED_FLAGS is acceptable because its documented within the check module.

JLBuenoLopez-eProsima (approved):
:+1: :sweat_smile:

PR #2024, opened 2021-06-24
Change constants initialization [11867]

richiware:
With this change we can upgrade the sphinx version for documentation.


MiguelCompany (approved):
LGTM with green CI

PR #2027, opened 2021-06-28
Limit SequenceNumberSet number of bits on deserialization [11875]

MiguelCompany:
This PR limits the maximum number of bits on the deserialization of SequenceNumberSet (i.e. ACKNACK and GAP submessages) to avoid a signed integer overflow on the high part of the sequence number.

As a bonus, it also drops messages with a protocol major version number lower than 2.

### DIFF
--- a/include/fastdds/rtps/messages/CDRMessage.hpp
+++ b/include/fastdds/rtps/messages/CDRMessage.hpp
@@ -265,7 +265,6 @@ inline SequenceNumberSet_t CDRMessage::readSequenceNumberSet(
     if (valid)
     {
         SequenceNumberSet_t sns(seqNum, numBits);
-        sns.base(seqNum);
         sns.bitmap_set(numBits, bitmap);
         return sns;
     }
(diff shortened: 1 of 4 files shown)


include/fastdds/rtps/messages/CDRMessage.hpp
```diff
@@ -265,11 +264,13 @@ inline SequenceNumberSet_t CDRMessage::readSequenceNumberSet(
 
     if (valid)
     {
+        SequenceNumberSet_t sns(seqNum, numBits);
         sns.base(seqNum);
```
richiware, line 268: Is it necessary this line now? I think the constructor of BitmapRage makes this work now.

richiware (changes requested):

include/fastdds/rtps/messages/CDRMessage.hpp
MiguelCompany: You are right! Fixed!

richiware (approved):

PR #2028, opened 2021-06-28
Remove assert on network statistics processing code

MiguelCompany:
As receiving the same datagram twice may be forced by a malicious party, we shouldn't assert on code that is part of the processing of a received datagram.

This PR changes one of such assertions into a harmless conditional check


MiguelCompany:
@richiware Please test mac

MiguelCompany:
@richiprosima Please test mac

IkerLuengo (approved):
LGTM

PR #2030, opened 2021-06-30
Notify status changes [11835]

MiguelCompany:
This PR:
* Performs notification of status changes using the `StatusCondition`
* Fixes listener calls for deadline statuses
* Implements `DataReader::get_subscription_matched_status`
* Implements `DataWriter::get_publication_matched_status`
* Changes the implementation of `Entity::get_status_changes` to use the raw status on `StatusConditionImpl`


IkerLuengo (approved):
LGTM

PR #2033, opened 2021-07-02
Fix segmentation fault on discovery server [11922]

richiware:
`DSClientEvent` accesses the information of `PDPSimple` without locking its mutex. When this event is running and a new Participant is discovered, a segmentation fault occurs. This PR adds the locking of `PDPSimple`'s mutex.


MiguelCompany (approved):
LGTM

MiguelCompany:
This PR is only affecting discovery server, and those test have run successfully. Going in.

PR #2036, opened 2021-07-02
Fix bug using not protected code in a test [11924]

richiware:
Fixes a bug in a test with multiple threads accessing a `std::vector` without protection.


MiguelCompany (approved):
Whoa! Nice catch!
LGTM.

richiware:
Tests were passed in #2047.

PR #2037, opened 2021-07-02
Fix deadlock with LivelinessManager [11925]

richiware:
Fixes a deadlock when asserting liveliness of a writer (due to a sample was received from it).

### DIFF
--- a/src/cpp/rtps/reader/StatefulReader.cpp
+++ b/src/cpp/rtps/reader/StatefulReader.cpp
@@ -619,10 +619,10 @@ bool StatefulReader::processHeartbeatMsg(
                     if (liveliness_kind_ == MANUAL_BY_TOPIC_LIVELINESS_QOS ||
                             writer->liveliness_kind() == MANUAL_BY_TOPIC_LIVELINESS_QOS)
                     {
-                        lock.unlock(); // Avoid deadlock with LivelinessManager.
                         auto wlp = this->mp_RTPSParticipant->wlp();
                         if ( wlp != nullptr)
                         {
+                            lock.unlock(); // Avoid deadlock with LivelinessManager.
                             wlp->sub_liveliness_manager_->assert_liveliness(
                                 writerGUID,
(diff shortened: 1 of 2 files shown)


src/cpp/rtps/reader/StatefulReader.cpp
```diff
@@ -612,6 +619,7 @@ bool StatefulReader::processHeartbeatMsg(
                     if (liveliness_kind_ == MANUAL_BY_TOPIC_LIVELINESS_QOS ||
                             writer->liveliness_kind() == MANUAL_BY_TOPIC_LIVELINESS_QOS)
                     {
+                        lock.unlock(); // Avoid deadlock with LivelinessManager.
```
MiguelCompany, line 622: If the deadlock is with the liveliness manager `assert_liveliness` method, I would move this line inside the `if (wlp != nullptr)`

If the deadlock is with the `wlp()` getter, it should stay here.

MiguelCompany (changes requested):
Maybe unrelated to this PR, but should we add liveliness assertion on `StatelessReader::processHeartbeatMsg` ?

richiware:
Reliable Datawriters will never send a heartbeat to Besteffort Datareaders. Only in case of using multicast but I think a Besteffort Datareader will never expect a heartbeat, then it will never do something if received one.

MiguelCompany:
> Reliable Datawriters will never send a heartbeat to Besteffort Datareaders

I am thinking of the case when the liveliness is set to `MANUAL_BY_TOPIC`, where an explicit liveliness heartbeat may be sent.

MiguelCompany:
> > Reliable Datawriters will never send a heartbeat to Besteffort Datareaders
> 
> I am thinking of the case when the liveliness is set to `MANUAL_BY_TOPIC`, where an explicit liveliness heartbeat may be sent.

After discussing this offline, we will do this in a separate PR.

MiguelCompany (approved):
LGTM

PR #2038, opened 2021-07-02
Add warning about Datasharing limitation with Datareader's history size [11926]

richiware:
There is a limitation about Datareader's HistoryQos and Datasharing. In case it is detected, a warning is shown to the user.

### DIFF
--- a/src/cpp/rtps/DataSharing/DataSharingListener.cpp
+++ b/src/cpp/rtps/DataSharing/DataSharingListener.cpp
@@ -205,7 +205,8 @@ void DataSharingListener::process_new_data ()
 
 bool DataSharingListener::add_datasharing_writer(
         const GUID_t& writer_guid,
-        bool is_volatile)
+        bool is_volatile,
+        int32_t reader_history_max_samples)
 {
     std::lock_guard<std::mutex> lock(mutex_);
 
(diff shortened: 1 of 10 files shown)


src/cpp/rtps/DataSharing/DataSharingListener.cpp
```diff
@@ -217,11 +218,21 @@ bool DataSharingListener::add_datasharing_writer(
 
     std::shared_ptr<ReaderPool> pool =
             std::static_pointer_cast<ReaderPool>(DataSharingPayloadPool::get_reader_pool(is_volatile));
-    pool->init_shared_memory(writer_guid, datasharing_pools_directory_);
-    writer_pools_.emplace_back(pool, pool->last_liveliness_sequence());
-    writer_pools_changed_.store(true);
+    if (pool->init_shared_memory(writer_guid, datasharing_pools_directory_))
+    {
+        if (0 == reader_history_max_samples ||
+                reader_history_max_samples > static_cast<int32_t>(pool->history_size() - 1))
```
MiguelCompany, line 224: Negative values also mean infinite.
`A > (B - 1)` ==> `A >= B`

```suggestion
        if (0 <= reader_history_max_samples ||
                reader_history_max_samples >= static_cast<int32_t>(pool->history_size()))
```

src/cpp/rtps/DataSharing/DataSharingListener.cpp
```diff
@@ -217,11 +218,21 @@ bool DataSharingListener::add_datasharing_writer(
...
+        {
+            logWarning(RTPS_READER,
+                    "Reader " << reader_->getGuid() << " was configured to have a large history (" << reader_history_max_samples << " max samples), but the history size used with writer " << writer_guid << " will be " << pool->history_size() <<
```
MiguelCompany, line 227: I think the linter will complain about this long line

MiguelCompany (changes requested):

src/cpp/rtps/DataSharing/DataSharingListener.cpp
richiware:
```suggestion
        if (0 >= reader_history_max_samples ||
                reader_history_max_samples >= static_cast<int32_t>(pool->history_size()))
```

richiware: Should be in this way to catch infinitive values?

MiguelCompany: The first condition is for the case where `reader_history_max_samples` is infinite. I'm assuming that a negative value also means infinite.

On the second condition, I think that `pool->history_size()` will always return a strictly positive value, as it is the way the pool is designed to be.

MiguelCompany: Now I see what you mean! It is `0 >= reader_history_max_samples`, not `0 <= reader_history_max_samples` !

MiguelCompany (approved):
LGTM

richiware:
Tests were passed in #2047.

PR #2040, opened 2021-07-02
Warning if cannot open multicast port [11928]

richiware:
If main metadata multicast port cannot be opened, warning to the user because discovery can fail with other applications.


MiguelCompany (commented):
When would this happen? We are setting socket options (i.e. `REUSEADDR`) to allow the same port to be open by different applications.

Is it possible to have a regression test?

richiware:
This happens if another application (not DDS) have the port opened. In my case happened running tests (random port) and sometimes the test tries to open a port owned by Firefox. Then the test fails. With this warning I have see quickly there is no problem about the failed test. And also I think it will be useful for users.

I don't know how to do a regression test of this. Is only a log warning.

MiguelCompany (approved):
LGTM

richiware:
Tests were passed in #2047.

PR #2041, opened 2021-07-02
Fix interval loop on events [11929]

richiware:
When the interval of an event is too low, the `ResourceEvent` thread can enter in to a loop.

### DIFF
--- a/src/cpp/rtps/resources/ResourceEvent.cpp
+++ b/src/cpp/rtps/resources/ResourceEvent.cpp
@@ -182,7 +182,7 @@ void ResourceEvent::event_service()
         auto current_time = std::chrono::steady_clock::now();
         if (current_time > next_trigger)
         {
-            next_trigger = current_time += std::chrono::microseconds(1);
+            next_trigger = current_time += std::chrono::microseconds(10);
         }
 
         cv_.wait_until(lock, next_trigger);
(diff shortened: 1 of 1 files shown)


src/cpp/rtps/resources/ResourceEvent.cpp
```diff
@@ -179,6 +179,12 @@ void ResourceEvent::event_service()
                 current_time_ + std::chrono::seconds(1) :
                 active_timers_[0]->next_trigger_time();
 
+        auto current_time = std::chrono::steady_clock::now();
+        if (current_time > next_trigger)
+        {
+            next_trigger = current_time += std::chrono::microseconds(10);
```
MiguelCompany, line 185: As `current_time` is not used after this line, it makes no sense to use `+=`

```suggestion
            next_trigger = current_time + std::chrono::microseconds(10);
```

MiguelCompany (changes requested):

src/cpp/rtps/resources/ResourceEvent.cpp
richiware: WOW! And was it compiling? Good catch.

MiguelCompany (approved):
LGTM

richiware:
Tests were passed in #2047.

PR #2070, opened 2021-07-15
Fix error in read_sample()/take_sample() operations with Datasharing [12130]

richiware:
The bug was reproduced when reader's history is 1 depth and while read_sample()/take_sample() operation is deserializing the data, the data  is overwritten by the datawriter. The `ReadTakeCommand` returned OK and the code tries to access a 0 length vector.

Also the code was changed to return NO_DATA if the operation will return no data.

### DIFF
--- a/src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
+++ b/src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
@@ -273,7 +273,6 @@ private:
             CacheChange_t* change,
             bool& deserialization_error)
     {
-        // Mark that some data is available
         bool ret_val = false;
         deserialization_error = false;
 
--- a/src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
+++ b/src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
@@ -274,7 +274,6 @@ private:
             bool& deserialization_error)
     {
         // Mark that some data is available
-        return_value_ = ReturnCode_t::RETCODE_OK;
         bool ret_val = false;
         deserialization_error = false;
 
@@ -297,6 +296,8 @@ private:
                     deserialization_error = true;
                     return false;
                 }
+
+                return_value_ = ReturnCode_t::RETCODE_OK;
             }
 
             ++current_slot_;
--- a/src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
+++ b/src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
@@ -297,6 +297,7 @@ private:
                     return false;
                 }
 
+                // Mark that some data is available
                 return_value_ = ReturnCode_t::RETCODE_OK;
             }
 
--- a/test/unittest/dds/subscriber/DataReaderTests.cpp
+++ b/test/unittest/dds/subscriber/DataReaderTests.cpp
@@ -1069,7 +1069,7 @@ TEST_F(DataReaderTests, resource_limits)
         FooSeq data_seq;
         SampleInfoSeq info_seq;
 
-        // With length == 0 the returned code should be okNO_DATA
+        // With length == 0 the returned code should be NO_DATA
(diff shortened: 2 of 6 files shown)


test/unittest/dds/subscriber/DataReaderTests.cpp
```diff
@@ -1065,12 +1066,11 @@ TEST_F(DataReaderTests, resource_limits)
         FooSeq data_seq;
         SampleInfoSeq info_seq;
 
-        // The standard is not clear on what shold be done if max_samples is 0. NO_DATA? OK with length = 0?
-        // We have assumed the correct interpretation is the second one, so the following loop starts at 0.
-        // This test should change whenever this interpretation becomes invalid.
+        // With length == 0 the returned code should be okNO_DATA
```
MiguelCompany, line 1069:
```suggestion
        // With length == 0 the returned code should be NO_DATA
```

src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
```diff
@@ -275,7 +275,6 @@ struct ReadTakeCommand
             bool& deserialization_error)
     {
         // Mark that some data is available
```
MiguelCompany, line 277: Move this comment below

src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
```diff
@@ -298,6 +297,8 @@ struct ReadTakeCommand
                     deserialization_error = true;
                     return false;
                 }
+
+                return_value_ = ReturnCode_t::RETCODE_OK;
```
MiguelCompany, line 301:
```suggestion
                // Mark that some data is available
                return_value_ = ReturnCode_t::RETCODE_OK;
```

MiguelCompany (changes requested):

MiguelCompany:
> Also the code was changed to return NO_DATA if the operation will return no data.

Instead of doing this, we could add the following lines on `DataReaderImpl::read_or_take_next_sample`, right after `ReturnCode_t code = cmd.return_value();`

```
    if ((ReturnCode_t::RETCODE_OK == code) && (0 == sample_infos.length()))
    {
        code = ReturnCode_t::RETCODE_NO_DATA;
    }
```

MiguelCompany:
@richiware I would also add a regression test for `read_next_sample`

src/cpp/fastdds/subscriber/DataReaderImpl/ReadTakeCommand.hpp
richiware: Done

richiware:
@MiguelCompany Regression test was implemented.

About your conditional statement suggestion in the `read_or_take_next_sample()`, I'm not sure. It will only fix the error in the `read_next_sample/take_next_sample`functions.
In case of calling `read/take` function with one sample in the history and the deserialization of that change fails, the function will return OK.

test/unittest/dds/subscriber/DataReaderTests.cpp
```diff
@@ -1065,12 +1066,11 @@ TEST_F(DataReaderTests, resource_limits)
         FooSeq data_seq;
         SampleInfoSeq info_seq;
 
-        // The standard is not clear on what shold be done if max_samples is 0. NO_DATA? OK with length = 0?
-        // We have assumed the correct interpretation is the second one, so the following loop starts at 0.
-        // This test should change whenever this interpretation becomes invalid.
+        // With length == 0 the returned code should be NO_DATA
```
MiguelCompany, line 1069:
```suggestion
        //  The standard is not clear on what should be done if max_samples is 0. NO_DATA? OK with length = 0?
        // We have assumed the correct interpretation is the first one.
        // This test should change whenever this interpretation becomes invalid.
```

MiguelCompany (changes requested):

MiguelCompany (approved):
LGTM with green CI

MiguelCompany (approved):

MiguelCompany:
@richiware This needs a rebase

richiware:
Rebase was done.

MiguelCompany (approved):
LGTM with green CI

Issue #2071, opened 2021-07-15
fastdds shm clean crashes


aleksas:
running `fastdds shm clean` cleans shared memory zombie resources

```python
$ fastdds shm clean
Traceback (most recent call last):
  File "/usr/local/bin/../tools/fastdds/fastdds.py", line 112, in <module>
    FastDDSParser()
  File "/usr/local/bin/../tools/fastdds/fastdds.py", line 85, in __init__
    getattr(self, args.command)()
  File "/usr/local/bin/../tools/fastdds/fastdds.py", line 103, in shm
    ShmParser(sys.argv[2:])
  File "/usr/local/tools/fastdds/shm/parser.py", line 71, in __init__
    Clean().run()
  File "/usr/local/tools/fastdds/shm/clean.py", line 44, in run
    zombie_segments = self.__clean_zombie_segments()
  File "/usr/local/tools/fastdds/shm/clean.py", line 114, in __clean_zombie_segments
    self.segments_in_use += 1
AttributeError: 'Clean' object has no attribute 'segments_in_use'

```

prepending **[segments_in_use](https://github.com/eProsima/Fast-DDS/blob/master/tools/fastdds/shm/clean.py#L114)** with `__` and running again

```python
$ fastdds shm clean
Traceback (most recent call last):
  File "/usr/local/bin/../tools/fastdds/fastdds.py", line 112, in <module>
    FastDDSParser()
  File "/usr/local/bin/../tools/fastdds/fastdds.py", line 85, in __init__
    getattr(self, args.command)()
  File "/usr/local/bin/../tools/fastdds/fastdds.py", line 103, in shm
    ShmParser(sys.argv[2:])
  File "/usr/local/tools/fastdds/shm/parser.py", line 71, in __init__
    Clean().run()
  File "/usr/local/tools/fastdds/shm/clean.py", line 45, in run
    zombie_ports = self.__clean_zombie_ports()
  File "/usr/local/tools/fastdds/shm/clean.py", line 149, in __clean_zombie_ports
    self.ports_in_use += 1
AttributeError: 'Clean' object has no attribute 'ports_in_use'
```

prepending **[ports_in_use](https://github.com/eProsima/Fast-DDS/blob/master/tools/fastdds/shm/clean.py#L149)** with `__` and running again works as intended.

ENV:
- centos 7
- python 2.7 , 3.4, 3.6

aleksasp:
[...]

PR #2072, opened 2021-07-15
Fix attribute error #2071

aleksas:


MiguelCompany (approved):
LGTM. Failures are known

PR #2074, opened 2021-07-15
Read/Take return NO_DATA if a future change is found [12134]

richiware:
`ReadTakeCommand` breaks the search of new changes if it found a future change of a writer although there is unread changes from other writers.

### DIFF
--- a/test/unittest/dds/subscriber/CMakeLists.txt
+++ b/test/unittest/dds/subscriber/CMakeLists.txt
@@ -22,7 +22,7 @@ if(NOT ((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER))
 
         set(SUBSCRIBERTESTS_SOURCE SubscriberTests.cpp)
         set(DATAREADERTESTS_SOURCE DataReaderTests.cpp)
-        
+
         if(WIN32)
             add_definitions(-D_WIN32_WINNT=0x0601)
         endif()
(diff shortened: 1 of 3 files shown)


MiguelCompany (changes requested):
How about a regression test?

richiware:
@MiguelCompany regression test added

test/unittest/dds/subscriber/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ if(NOT ((MSVC OR MSVC_IDE) AND EPROSIMA_INSTALLER))
 
         add_executable(DataReaderTests ${DATAREADERTESTS_SOURCE})
         target_compile_definitions(DataReaderTests PRIVATE FASTRTPS_NO_LIB
+            BOOST_ASIO_STANDALONE
+            ASIO_STANDALONE
```
MiguelCompany, line 47: No need of these, as we are linking fastrtps (see my next comment)

test/unittest/dds/subscriber/DataReaderTests.cpp
```diff
@@ -57,6 +57,9 @@
 
 #include "../../logging/mock/MockConsumer.h"
 
+#include <rtps/transport/test_UDPv4Transport.h>
```
MiguelCompany, line 60: This is adding an unnecessary dependency on asio (and thus the build fails), but only the transport descriptor is necessary:

```suggestion
#include <fastdds/rtps/transport/test_UDPv4TransportDescriptor.h>
```

MiguelCompany (changes requested):

test/unittest/dds/subscriber/DataReaderTests.cpp
richiware: I need access to `rtps::test_UDPv4Transport::test_UDPv4Transport_ShutdownAllNetwor` in line 1974.

test/unittest/dds/subscriber/CMakeLists.txt
richiware: Without that compiler is trying to find a boost header file.

test/unittest/dds/subscriber/DataReaderTests.cpp
MiguelCompany: You could achieve the same with `percentageOfMessagesToDrop` or `messages_filter_` on `test_UDPv4TransportDescriptor`

richiware: Done

test/unittest/dds/subscriber/CMakeLists.txt
richiware: Done

MiguelCompany (approved):
LGTM

PR #2094, opened 2021-07-27
Fix publication throughput statistic on volatile writers [12269]

MiguelCompany:


jparisu (approved):
LGTM
(Tested by FastDDS Monitor)

PR #2096, opened 2021-07-27
Fix late joiners example

### LINKED ISSUE #2062: LateJoiner example  The Volatile Subsciber not reading anything
I run the example FastDDS/examples/DDS/LateJoiners, build with CMake. and run exe. 

application behavior should be according to the readme
* The Volatile Subscriber only holds the last 20 samples sent after its creation.
* The Transient-Local Subscriber holds all samples, including the 20 that were sent before its creation.

issue - but I am not receiving anything from Volatile Subscriber 

also tried changing wqos.durability().kind = VOLATILE_DURABILITY_QOS;
still same result,  VOLATILE subscriber not receiving anyhting 

Transient-Local is okay, it holds 40 samples

MiguelCompany:
Correctly fix #2062

### DIFF
--- a/examples/C++/DDS/LateJoiners/latejoiners.cpp
+++ b/examples/C++/DDS/LateJoiners/latejoiners.cpp
@@ -156,6 +156,14 @@ void latejoiners()
         return;
     }
 
+    // Wait for volatile reader to match with writer, otherwise it may consider the new samples as historical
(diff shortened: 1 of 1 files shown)


EduPonz (approved):

EduPonz:
I'm merging it since the change only affects an example (for which there are no tests)

PR #2113, opened 2021-08-03
[12196] DomainParticipantQos propagation to the RTPS layer implementation

JLBuenoLopez-eProsima:
This PR implements the link between the DDS and the RTPS layer to propagate the updates to the `DomainParticipantQos`.
It also includes the access protection of the remote servers list with the `PDP` mutex.

### DIFF
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
@@ -351,6 +351,9 @@ ReturnCode_t DomainParticipantImpl::set_qos(
         fastrtps::rtps::RTPSParticipantAttributes patt;
         set_attributes_from_qos(patt, qos_);
         rtps_participant_->update_attributes(patt);
+
+        // Reset flag
+        qos_.user_data().hasChanged = false;
     }
 
     return ReturnCode_t::RETCODE_OK;
@@ -1754,7 +1757,10 @@ void DomainParticipantImpl::set_qos(
     if (!(to.user_data() == from.user_data()))
     {
         to.user_data() = from.user_data();
-        to.user_data().hasChanged = true;
+        if (!first_time)
+        {
+            to.user_data().hasChanged = true;
+        }
     }
     if (first_time && !(to.allocation() == from.allocation()))
     {
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
@@ -344,7 +344,7 @@ ReturnCode_t DomainParticipantImpl::set_qos(
         return ReturnCode_t::RETCODE_IMMUTABLE_POLICY;
     }
 
-    if (enabled && set_qos(qos_, qos_to_set, !enabled))
+    if (set_qos(qos_, qos_to_set, !enabled) && enabled)
     {
         // Notify the participant that there is a QoS update
         fastrtps::rtps::RTPSParticipantAttributes patt;
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
@@ -343,17 +343,13 @@ ReturnCode_t DomainParticipantImpl::set_qos(
     {
         return ReturnCode_t::RETCODE_IMMUTABLE_POLICY;
     }
-    set_qos(qos_, qos_to_set, !enabled);
 
-    if (enabled && qos_.user_data().hasChanged)
+    if (enabled && set_qos(qos_, qos_to_set, !enabled))
     {
         // Notify the participant that there is a QoS update
         fastrtps::rtps::RTPSParticipantAttributes patt;
         set_attributes_from_qos(patt, qos_);
         rtps_participant_->update_attributes(patt);
-
-        // Reset flag
-        qos_.user_data().hasChanged = false;
     }
 
     return ReturnCode_t::RETCODE_OK;
@@ -1745,11 +1741,13 @@ bool DomainParticipantImpl::has_active_entities()
     return false;
 }
 
-void DomainParticipantImpl::set_qos(
+bool DomainParticipantImpl::set_qos(
         DomainParticipantQos& to,
         const DomainParticipantQos& from,
         bool first_time)
 {
+    bool qos_has_changed = false;
+
     if (!(to.entity_factory() == from.entity_factory()))
     {
         to.entity_factory() = from.entity_factory();
@@ -1757,9 +1755,10 @@ void DomainParticipantImpl::set_qos(
     if (!(to.user_data() == from.user_data()))
     {
         to.user_data() = from.user_data();
+        to.user_data().hasChanged = true;
         if (!first_time)
         {
-            to.user_data().hasChanged = true;
+            qos_has_changed = true;
         }
     }
     if (first_time && !(to.allocation() == from.allocation()))
@@ -1782,6 +1781,8 @@ void DomainParticipantImpl::set_qos(
     {
         to.name() = from.name();
     }
+
+    return qos_has_changed;
 }
 
 fastrtps::types::ReturnCode_t DomainParticipantImpl::check_qos(
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
@@ -345,7 +345,7 @@ ReturnCode_t DomainParticipantImpl::set_qos(
     }
     set_qos(qos_, qos_to_set, !enabled);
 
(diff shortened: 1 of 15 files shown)


src/cpp/rtps/builtin/discovery/participant/PDPServer.cpp
```diff
@@ -1224,6 +1236,7 @@ fastdds::rtps::ddb::DiscoveryDataBase& PDPServer::discovery_db()
 
 const RemoteServerList_t& PDPServer::servers()
 {
+    std::lock_guard<std::recursive_mutex> lock(*getMutex());
```
EduPonz, line 1239: This does not protect anything as the lock is released on returning, and the return is a const ref, which merely means that the caller cannot modify the object, but it does not prevent other threads from doing so.

In any case, this method is not used in Fast DDS or Discovery Server not even once. Although we cannot just remove public API, we can add some doxygen documentation in the header clearly stating that the returned `RemoteServerList_t&` may be modified by another thread. What do you think?

EduPonz (changes requested):

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -343,6 +344,15 @@ ReturnCode_t DomainParticipantImpl::set_qos(
...
     }
     set_qos(qos_, qos_to_set, !enabled);
+
+    if (enabled && qos_.user_data().hasChanged)
+    {
+        // Notify the participant that there is a QoS update
+        fastrtps::rtps::RTPSParticipantAttributes patt;
+        set_attributes_from_qos(patt, qos_);
+        rtps_participant_->update_attributes(patt);
+    }
```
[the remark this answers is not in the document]
JLBuenoLopez-eProsima: In the [`DataWriterImpl`](https://github.com/eProsima/Fast-DDS/blob/master/src/cpp/fastdds/publisher/DataWriterImpl.cpp#L890) and the [`DataReaderImpl`](https://github.com/eProsima/Fast-DDS/blob/master/src/cpp/fastdds/subscriber/DataReaderImpl.cpp#L733) it is done in this other way. I have tried to follow the same logic to help read and compare code. But I am not against moving this logic inside the other method. What do you think?

src/cpp/rtps/builtin/BuiltinProtocols.cpp
```diff
@@ -129,6 +126,13 @@ bool BuiltinProtocols::initBuiltinProtocols(
         return false;
     }
 
+    {
+        std::unique_lock<std::recursive_mutex> lock(*mp_PDP->getMutex());
+        m_DiscoveryServers = m_att.discovery_config.m_DiscoveryServers;
+    }
+
+    transform_server_remote_locators(p_part->network_factory());
```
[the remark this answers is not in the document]
JLBuenoLopez-eProsima: :man_facepalming: I thought of running the Discovery Server tests locally and then I forgot. I think that the best solution right now is to not take the mutex. Although I would suggest leaving a `TODO` comment stating this situation in case in the future needs to be refactored.

src/cpp/rtps/builtin/discovery/participant/PDPServer.cpp
JLBuenoLopez-eProsima: You are right. The method is not being called within Fast DDS. I agree that we can add doxygen documentation stating that the access to the returned list may be unsafe if the remote servers list is modified at the same time by the user.

src/cpp/rtps/builtin/BuiltinProtocols.cpp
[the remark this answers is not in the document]
EduPonz: I agree

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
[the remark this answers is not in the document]
EduPonz: Fair enough, I thought it was as I suggested in `DataWriterImpl`, but it's not 🤦🏼 . Leave it as it is then!

EduPonz (approved):
LGTM

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -1745,21 +1741,24 @@ bool DomainParticipantImpl::has_active_entities()
     return false;
 }
 
-void DomainParticipantImpl::set_qos(
+bool DomainParticipantImpl::set_qos(
         DomainParticipantQos& to,
         const DomainParticipantQos& from,
         bool first_time)
 {
+    bool qos_has_changed = false;
```
EduPonz, line 1749: This flag is more "something that should be communicated to the RTPS layer has changed" rather than something has changed, isn't it?

src/cpp/fastdds/domain/DomainParticipantImpl.hpp
```diff
@@ -559,7 +559,7 @@ class DomainParticipantImpl
     std::string get_inner_type_name(
             const fastrtps::rtps::SampleIdentity& id) const;
 
-    static void set_qos(
+    static bool set_qos(
```
EduPonz, line 562: It'd be nice to have some doxygen here. Furthermore, there is a mock of this function which signature we should update

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -1103,27 +1103,39 @@ bool RTPSParticipantImpl::registerReader(
...
-                patt.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SUPER_CLIENT ||
-                patt.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SERVER ||
-                patt.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::BACKUP)
+        if (m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::CLIENT ||
+                m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SUPER_CLIENT ||
+                m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SERVER ||
+                m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::BACKUP)
         {
             mp_builtinProtocols->m_DiscoveryServers = patt.builtin.discovery_config.m_DiscoveryServers;
             mp_builtinProtocols->m_att.discovery_config.m_DiscoveryServers =
```
EduPonz, line 1135: This we don't need to do since `mp_builtinProtocols->m_att == this->m_att.builtin`

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -1103,27 +1103,39 @@ bool RTPSParticipantImpl::registerReader(
...
-        if (patt.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::CLIENT ||
-                patt.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SUPER_CLIENT ||
-                patt.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SERVER ||
-                patt.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::BACKUP)
+        if (m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::CLIENT ||
+                m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SUPER_CLIENT ||
+                m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SERVER ||
+                m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::BACKUP)
         {
             mp_builtinProtocols->m_DiscoveryServers = patt.builtin.discovery_config.m_DiscoveryServers;
```
EduPonz, line 1134: This way whenever we introduce validation above, we don't need to do it twice
```suggestion
            mp_builtinProtocols->m_DiscoveryServers = m_att.builtin.discovery_config.m_DiscoveryServers;
```

EduPonz (changes requested):

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
JLBuenoLopez-eProsima: I have renamed the flag to `qos_should_be_updated`. It is true that it is not only enough that the QoS has changed because only a few attributes are allowed to be changed.

src/cpp/fastdds/domain/DomainParticipantImpl.hpp
JLBuenoLopez-eProsima: Nice catch!

EduPonz (approved):
LGTM

EduPonz (approved):
LGTM

PR #2128, opened 2021-08-06
[12371] Waitset implementation

MiguelCompany:
Add WaitSet, GuardCondition, and StatusCondition implementation and tests.

Documentation PR: eProsima/Fast-DDS-docs#279


MiguelCompany:

IkerLuengo:

IkerLuengo (changes requested):
`Entity::get_statuscondition()` is returning a const reference. We should be returning a non-const reference so that we can apply modifications, like `StatusCondition::set_enabled_statuses()`, which is a mutator method. Although `Condition` itself has no mutating methods, we cannot assume that its derivatives have none.

This was found during the review of the examples in the documentation

MiguelCompany:
> `Entity::get_statuscondition()` is returning a const reference. We should be returning a non-const reference so that we can apply modifications, like `StatusCondition::set_enabled_statuses()`, which is a mutator method. Although `Condition` itself has no mutating methods, we cannot assume that its derivatives have none.
> 
> This was found during the review of the examples in the documentation
> 
> And while we are at it, we could also remove the logWarning stating that `get_statuscondition` is not implemented

This is part of the changes .... Were you looking at the code in `master`, perhaps?

IkerLuengo (approved):
> This is part of the changes .... Were you looking at the code in `master`, perhaps?

Sincerely, I do not know what I was looking 😥

PR #2131, opened 2021-08-06
WireProtocolConfigQos update through DomainParticipant::set_qos [12197]

EduPonz:
This PR enables updating `WireProtocolConfigQos` iff the only modified field is the list of remote discovery servers.

**IMPORTANT**: Merge after #2113

### DIFF
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
@@ -1848,6 +1848,7 @@ bool DomainParticipantImpl::can_qos_be_updated(
             !(to.wire_protocol().builtin.discovery_config.leaseDuration_announcementperiod == from.wire_protocol().builtin.discovery_config.leaseDuration_announcementperiod) ||
             !(to.wire_protocol().builtin.discovery_config.initial_announcements == from.wire_protocol().builtin.discovery_config.initial_announcements) ||
             !(to.wire_protocol().builtin.discovery_config.m_simpleEDP == from.wire_protocol().builtin.discovery_config.m_simpleEDP) ||
+            !(strcmp(to.wire_protocol().builtin.discovery_config.static_edp_xml_config(), from.wire_protocol().builtin.discovery_config.static_edp_xml_config()) == 0) ||
             !(to.wire_protocol().builtin.discovery_config.ignoreParticipantFlags == from.wire_protocol().builtin.discovery_config.ignoreParticipantFlags))))
         {
             updatable = false;
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
@@ -1819,41 +1819,64 @@ bool DomainParticipantImpl::can_qos_be_updated(
     if (!(to.wire_protocol() == from.wire_protocol()))
     {
         // Check that the only modification was in wire_protocol().discovery_config.m_DiscoveryServers
-        if ((to.wire_protocol().builtin.discovery_config.m_DiscoveryServers == from.wire_protocol().builtin.discovery_config.m_DiscoveryServers) ||
-            (!(to.wire_protocol().builtin.discovery_config.m_DiscoveryServers == from.wire_protocol().builtin.discovery_config.m_DiscoveryServers) &&
-            (!(to.wire_protocol().prefix == from.wire_protocol().prefix) ||
-            !(to.wire_protocol().participant_id == from.wire_protocol().participant_id) ||
-            !(to.wire_protocol().port == from.wire_protocol().port) ||
-            !(to.wire_protocol().throughput_controller == from.wire_protocol().throughput_controller) ||
-            !(to.wire_protocol().default_unicast_locator_list == from.wire_protocol().default_unicast_locator_list) ||
-            !(to.wire_protocol().default_multicast_locator_list == from.wire_protocol().default_multicast_locator_list) ||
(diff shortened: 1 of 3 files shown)


src/cpp/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -1813,8 +1818,43 @@ bool DomainParticipantImpl::can_qos_be_updated(
...
+            !(to.wire_protocol().builtin.discovery_config.discoveryServer_client_syncperiod == from.wire_protocol().builtin.discovery_config.discoveryServer_client_syncperiod) ||
+            !(to.wire_protocol().builtin.discovery_config.m_PDPfactory == from.wire_protocol().builtin.discovery_config.m_PDPfactory) ||
+            !(to.wire_protocol().builtin.discovery_config.leaseDuration == from.wire_protocol().builtin.discovery_config.leaseDuration) ||
+            !(to.wire_protocol().builtin.discovery_config.leaseDuration_announcementperiod == from.wire_protocol().builtin.discovery_config.leaseDuration_announcementperiod) ||
+            !(to.wire_protocol().builtin.discovery_config.initial_announcements == from.wire_protocol().builtin.discovery_config.initial_announcements) ||
+            !(to.wire_protocol().builtin.discovery_config.m_simpleEDP == from.wire_protocol().builtin.discovery_config.m_simpleEDP) ||
+            !(to.wire_protocol().builtin.discovery_config.ignoreParticipantFlags == from.wire_protocol().builtin.discovery_config.ignoreParticipantFlags))))
```
JLBuenoLopez-eProsima, line 1851: Private variable `DiscoverySettings::static_edp_xml_config_` is not checked. A getter of this member is available: `static_edp_xml_config`.

I am not sure if it is better to split the code in several wrapper methods that helps the code readability. There are several nested conditions that can be difficult to read.

test/unittest/dds/participant/ParticipantTests.cpp
```diff
@@ -376,6 +376,237 @@ TEST(ParticipantTests, ChangePSMDomainParticipantQos)
...
+    qos.wire_protocol().builtin.discovery_config.m_DiscoveryServers.pop_front();
+    ASSERT_TRUE(participant->set_qos(qos) == ReturnCode_t::RETCODE_OK);
+    participant->get_qos(set_qos);
+    ASSERT_EQ(set_qos, qos);
+
+    // Check that removing all servers is OK
+    fastdds::rtps::RemoteServerList_t servers;
+    qos.wire_protocol().builtin.discovery_config.m_DiscoveryServers = servers;
+    ASSERT_TRUE(participant->set_qos(qos) == ReturnCode_t::RETCODE_OK);
+    participant->get_qos(set_qos);
```
JLBuenoLopez-eProsima, line 412: Having added only one server, is it not the same case? I suggest adding a second server at the beginning and then both cases will check different behaviors.

test/unittest/dds/participant/ParticipantTests.cpp
```diff
@@ -376,6 +376,237 @@ TEST(ParticipantTests, ChangePSMDomainParticipantQos)
...
+    ASSERT_FALSE(set_qos == qos);
+
+    // Check changing wire_protocol().participant_id is NOT OK
+    participant->get_qos(qos);
+    qos.wire_protocol().participant_id = 7;
+    ASSERT_TRUE(participant->set_qos(qos) == ReturnCode_t::RETCODE_IMMUTABLE_POLICY);
+    participant->get_qos(set_qos);
+    ASSERT_FALSE(set_qos == qos);
+
```
JLBuenoLopez-eProsima, line 428: Maybe it would be nice to have a comment explaining that `wire_protocol().port` changes are not checked (the members are private and it is not possible to modify them from the test). Also with `wire_protocol().builtin.discovery_config.m_PDPFactory`.

test/unittest/dds/participant/ParticipantTests.cpp
```diff
@@ -376,6 +376,237 @@ TEST(ParticipantTests, ChangePSMDomainParticipantQos)
...
+    // Check changing wire_protocol().builtin.discovery_config.ignoreParticipantFlags is NOT OK
+    participant->get_qos(qos);
+    qos.wire_protocol().builtin.discovery_config.ignoreParticipantFlags = fastrtps::rtps::ParticipantFilteringFlags::FILTER_DIFFERENT_HOST;
+    ASSERT_TRUE(participant->set_qos(qos) == ReturnCode_t::RETCODE_IMMUTABLE_POLICY);
+    participant->get_qos(set_qos);
+    ASSERT_FALSE(set_qos == qos);
+
```
JLBuenoLopez-eProsima, line 606: Adding `DiscoverySettings::static_edp_xml_config_` check in the implementation will allow to check this change as well in this test.

JLBuenoLopez-eProsima (changes requested):

JLBuenoLopez-eProsima:
Please, fix also linters.

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
EduPonz: I don't think I like the idea of splitting the check into different functions that are only called once from here, since they will not be reusable and their only purpose would be to move some of this code behind a courtain.

JLBuenoLopez-eProsima (commented):
LGTM though linters are still complaining.

JLBuenoLopez-eProsima (approved):
:ok_hand:

PR #2138, opened 2021-08-10
Propagate servers list updates to PDPServer and PDPClient [12278]

EduPonz:
This PR brings the following functionality:

1. `DomainParticipantQos::wire_protocol().builtin.discovery_config.m_DiscoveryServers` can only be modified to add servers, but not to remove them.
2. Propagate new servers to `PDPServer` and `PDPClient` so that discovery can take place
3. Add simple discovery test using discovery server to check that servers added both to a SERVER and a CLIENT have the expected effect.

### DIFF
--- a/src/cpp/rtps/builtin/discovery/participant/PDPClient.h
+++ b/src/cpp/rtps/builtin/discovery/participant/PDPClient.h
@@ -139,13 +139,15 @@ protected:
      * Manually match PDP writer endpoint of a given server. The function is not thread safe (nts) in
      * the sense that it does not take the PDP mutex. It does however take temp_data_lock_
      */
-    void match_pdp_writer_nts_(const eprosima::fastdds::rtps::RemoteServerAttributes& server_att);
+    void match_pdp_writer_nts_(
+            const eprosima::fastdds::rtps::RemoteServerAttributes& server_att);
 
     /**
      * Manually match PDP reader endpoint of a given server. The function is not thread safe (nts) in
      * the sense that it does not take the PDP mutex. It does however take temp_data_lock_
      */
-    void match_pdp_reader_nts_(const eprosima::fastdds::rtps::RemoteServerAttributes& server_att);
+    void match_pdp_reader_nts_(
+            const eprosima::fastdds::rtps::RemoteServerAttributes& server_att);
 
 private:
 
--- a/src/cpp/rtps/builtin/discovery/participant/PDPClient.h
+++ b/src/cpp/rtps/builtin/discovery/participant/PDPClient.h
@@ -128,6 +128,25 @@ public:
      */
     bool match_servers_EDP_endpoints();
 
+    /*
+     * Update the list of remote servers
+     */
+    void update_remote_servers_list();
+
+protected:
+
+    /**
+     * Manually match PDP writer endpoint of a given server. The function is not thread safe (nts) in
+     * the sense that it does not take the PDP mutex. It does however take temp_data_lock_
+     */
+    void match_pdp_writer_nts_(const eprosima::fastdds::rtps::RemoteServerAttributes& server_att);
+
+    /**
+     * Manually match PDP reader endpoint of a given server. The function is not thread safe (nts) in
+     * the sense that it does not take the PDP mutex. It does however take temp_data_lock_
+     */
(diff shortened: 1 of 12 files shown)


test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
```diff
@@ -52,4 +61,110 @@ TEST(DDSDiscovery, IgnoreParticipantFlags)
...
+ *    3. Finally connect the two servers by adding one of them to the others list
+ */
+TEST(DDSDiscovery, AddDiscoveryServerToList)
+{
+    using namespace eprosima::fastdds::dds;
+    using namespace eprosima::fastrtps::rtps;
+
+    /* Get random port from the environment */
+    char* value = nullptr;
+    value = std::getenv("W_UNICAST_PORT_RANDOM_NUMBER");
```
JLBuenoLopez-eProsima, line 81: Fast DDS provides a method to read environment variables get_env(). It can be found in private header file utils/SystemInfo.hpp. This method is tested for all supported platforms. I suggest using it so we have only to maintain one method in case we start supporting new platforms in the future.

test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
```diff
@@ -52,4 +61,110 @@ TEST(DDSDiscovery, IgnoreParticipantFlags)
...
+        value = &std::string("11811")[0];
+    }
+
+    /* Create first server */
+    PubSubParticipant<HelloWorldType> server_1(0u, 0u, 0u, 0u);
+    // Set participant as server
+    WireProtocolConfigQos server_1_qos;
+    server_1_qos.builtin.discovery_config.discoveryProtocol = DiscoveryProtocol_t::SERVER;
+    // Generate random GUID prefix
+    srand((unsigned)time(NULL));
```
JLBuenoLopez-eProsima, line 93: This is a small nitpick (and a come-back from a previous review :joy:)

```suggestion
    srand(static_cast<unsigned>(time(nullptr)));
```

More C++ like :stuck_out_tongue_closed_eyes:

JLBuenoLopez-eProsima (changes requested):

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -1131,13 +1131,41 @@ void RTPSParticipantImpl::update_attributes(
                 m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::SERVER ||
                 m_att.builtin.discovery_config.discoveryProtocol == DiscoveryProtocol::BACKUP)
         {
+            // Add incoming servers iff we don't know about them already
```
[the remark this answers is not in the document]
EduPonz: "iff" is short for "if and only if"

src/cpp/rtps/builtin/discovery/participant/PDPClient.h
```diff
@@ -128,6 +128,27 @@ class PDPClient : public PDP
...
 
+    /*
+     * Update the list of remote servers
+     */
+    void update_remote_servers_list();
+
+protected:
+
+    /**
+     * Manually match PDP writer endpoint of a given server. The function is not thread safe (nts) in
```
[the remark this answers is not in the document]
EduPonz: What is means is that the local PDP reader is matched with the server's PDP writer, I'll clarify

test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
JLBuenoLopez-eProsima: Didnn't you like the `nullptr` instead of the `NULL`?

EduPonz: > Didnn't you like the `nullptr` instead of the `NULL`?

🤦🏼‍♂️ I missed that!

JLBuenoLopez-eProsima (approved):
:+1:

JLBuenoLopez-eProsima (approved):
LGTM.

Discovery server test failure is not one of the common ones but I have reproduce it in `master`. It seems that it is flaky and fails from time to time (but really sparingly).

PR #2141, opened 2021-08-12
Fix wrong iterator after vector reallocation ln LivelinessManager [12362]

### LINKED ISSUE #2129: LivelinessManager "writers_" vector reallocation issue [12361]
Hi
Maybe there is an issue in LivelinessManager.cpp(2.0.x version) about "writers_" vector.  

As we know that the std::vector will reallocation when the new item is added and the contiguous vector space is not enough at the same time, the new heap address will be filled with all the items, the old heap will be destroyed.
The issue is that the std::iterator pointer(for example: "timer_owner_" in LivelinessManager) will be invalidated if the std::iterator pointer is pointing to one of them before reallocating, caused by the address is destroyed by reallocation.

Although it's an extreme case, we indeed encounted the issue.
My suggestion for fixing the issue is to add checking methods to change the std::iterator pointer when add new item added(for example: LivelinessManager::add_writer).
Could you pls confirm whether my suggestion is right,  and if not right,  pls give fix method.

Thanks a lot.

richiware:
Fixes #2129

### DIFF
--- a/src/cpp/rtps/writer/LivelinessManager.cpp
+++ b/src/cpp/rtps/writer/LivelinessManager.cpp
@@ -65,7 +65,7 @@ bool LivelinessManager::add_writer(
     if (!calculate_next())
     {
         timer_.cancel_timer();
-        return false;
+        return true;
     }
 
     // Some times the interval could be negative if a writer expired during the call to this function
--- a/src/cpp/rtps/writer/LivelinessManager.cpp
+++ b/src/cpp/rtps/writer/LivelinessManager.cpp
@@ -69,7 +69,7 @@ bool LivelinessManager::add_writer(
     }
 
     // Some times the interval could be negative if a writer expired during the call to this function
-    // Once in this situation there is not much we can do but let asio timers expire inmediately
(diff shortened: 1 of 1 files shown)


src/cpp/rtps/writer/LivelinessManager.cpp
```diff
@@ -61,6 +61,19 @@ bool LivelinessManager::add_writer(
...
     writers_.emplace_back(guid, kind, lease_duration);
+
+    if (!calculate_next())
+    {
+        timer_.cancel_timer();
+        return true;
+    }
+
+    // Some times the interval could be negative if a writer expired during the call to this function
+    // Once in this situation there is not much we can do but let asio timers expire inmediately
```
EduPonz, line 72: It's misspelled in remove_writer as well
```suggestion
    // Once in this situation there is not much we can do but let asio timers expire immediately
```

EduPonz (dismissed):
Just a nitpick

EduPonz:
@Mergifyio backport 2.0.x 2.1.x 2.2.x 2.3.x

PR #2155, opened 2021-08-16
[12199] Add SystemInfo::file_exists

JLBuenoLopez-eProsima:
Add a method that checks that given a `filename` the file exists (ensuring also that it is a proper file and not a directory).

### DIFF
--- a/src/cpp/utils/SystemInfo.cpp
+++ b/src/cpp/utils/SystemInfo.cpp
@@ -79,6 +79,7 @@ bool SystemInfo::file_exists(
         const std::string& filename)
 {
     struct stat s;
+    // Check existence and that it is a regular file (and not a folder)
     return (stat(filename.c_str(), &s) == 0 && s.st_mode & S_IFREG);
 }
 
--- a/src/cpp/utils/SystemInfo.cpp
+++ b/src/cpp/utils/SystemInfo.cpp
@@ -14,6 +14,7 @@
 
 #include "SystemInfo.hpp"
 
+#include <sys/stat.h>
 #ifdef _WIN32
 #include <windows.h>
 #else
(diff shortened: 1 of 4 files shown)


src/cpp/utils/SystemInfo.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include "SystemInfo.hpp"
 
+#include <sys/stat.h>
```
IkerLuengo, line 17: How portable is this? (e.g., Windows). I'm relaunching tests of windows because they failed for other reasons.

test/unittest/utils/SystemInfoTests.cpp
```diff
@@ -67,6 +76,44 @@ TEST(SystemInfoTests, GetUsernameTest)
...
+    EXPECT_TRUE(eprosima::SystemInfo::file_exists(filename));
+
+    // 2. Check that a valid path and filename returns correctly
+    char buffer[SIZE];
+    char* current_dir = getcwd(buffer, SIZE);
+    if (current_dir)
+    {
+        filename = current_dir;
+        filename += "/environment_test_file.json";
+    }
```
IkerLuengo, line 95: If (for whatever reason) we don't get `current_dir`, we will be checking with the same filename again. Making the test fail in that case would be more informative. It was done this way for the `non_existent.txt` file below.

Or, if we know that `getcwd` will fail for some platforms, we can include the `EXPECT` inside the if block and skip it completely on those platforms. Adding a comment stating why we are doing this would be a plus.

IkerLuengo (changes requested):

IkerLuengo:

src/cpp/utils/SystemInfo.cpp
IkerLuengo: It seems to work, so OK :+1

test/unittest/utils/SystemInfoTests.cpp
IkerLuengo: I addressed this on commit c98fec6d9

PR #2157, opened 2021-08-17
[12199] Load environment file

JLBuenoLopez-eProsima:
This PR should be merged after #2155. It implements `SystemInfo::load_environment_file` and tests its functionality.

### DIFF
--- a/src/cpp/utils/SystemInfo.cpp
+++ b/src/cpp/utils/SystemInfo.cpp
@@ -22,6 +22,9 @@
 #include <unistd.h>
 #endif // _WIN32
 
+#include <fstream>
+
+#include <json.hpp>
 #include <fastrtps/types/TypesBase.h>
 
 namespace eprosima {
@@ -81,6 +84,37 @@ bool SystemInfo::file_exists(
     struct stat s;
     // Check existence and that it is a regular file (and not a folder)
     return (stat(filename.c_str(), &s) == 0 && s.st_mode & S_IFREG);
+}
+
+ReturnCode_t SystemInfo::load_environment_file(
+        const std::string& filename,
+        const std::string& env_name,
+        std::string& env_value)
+{
+    // Check that the file exists
+    if (!SystemInfo::file_exists(filename))
+    {
+        return ReturnCode_t::RETCODE_BAD_PARAMETER;
+    }
+
+    // Read json file
+    std::ifstream file(filename);
+    nlohmann::json file_content;
+    file >> file_content;
+
+    try
+    {
+        env_value = file_content.at(env_name);
+        if (env_value.empty() || env_value.compare("") == 0)
+        {
+            return ReturnCode_t::RETCODE_NO_DATA;
+        }
+    }
+    catch(const std::exception&)
+    {
+        return ReturnCode_t::RETCODE_NO_DATA;
+    }
+    return ReturnCode_t::RETCODE_OK;
 }
 
 } // eprosima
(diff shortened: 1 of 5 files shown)


src/cpp/utils/SystemInfo.cpp
```diff
@@ -83,4 +86,35 @@ bool SystemInfo::file_exists(
...
+    // Check that the file exists
+    if (!SystemInfo::file_exists(filename))
+    {
+        return ReturnCode_t::RETCODE_BAD_PARAMETER;
+    }
+
+    // Read json file
+    std::ifstream file(filename);
+    nlohmann::json file_content;
+    file >> file_content;
```
IkerLuengo, line 103: What if the file is empty or the json is malformed? This will throw. We should return some kind of error code, probably RETCODE_ERROR.

src/cpp/utils/SystemInfo.cpp
```diff
@@ -83,4 +86,35 @@ bool SystemInfo::file_exists(
...
+
+    try
+    {
+        env_value = file_content.at(env_name);
+        if (env_value.empty() || env_value.compare("") == 0)
+        {
+            return ReturnCode_t::RETCODE_NO_DATA;
+        }
+    }
+    catch(const std::exception&)
```
IkerLuengo: I would capture only `json::exception`, and let all other exception pass through.

test/unittest/utils/SystemInfoTests.cpp
```diff
@@ -114,6 +115,38 @@ TEST(SystemInfoTests, FileExistsTest)
...
+    std::string empty_environment_variable = "EMPTY_ENV_VAR";
+    EXPECT_EQ(ReturnCode_t::RETCODE_NO_DATA, eprosima::SystemInfo::load_environment_file(filename,
+            empty_environment_variable, environment_value));
+    EXPECT_TRUE(environment_value.empty());
+
+    // 4. Check that a non-existent file returns RETCODE_BAD_PARAMETER
+    filename = "non_existent.json";
+    EXPECT_EQ(ReturnCode_t::RETCODE_BAD_PARAMETER, eprosima::SystemInfo::load_environment_file(filename,
+            eprosima::fastdds::rtps::DEFAULT_ROS2_MASTER_URI, environment_value));
+    EXPECT_TRUE(environment_value.empty());
```
IkerLuengo, line 147: What about an file that exists but is empty or has a malformed json?

IkerLuengo (changes requested):

src/cpp/utils/SystemInfo.cpp
IkerLuengo: addressed in f67682c

src/cpp/utils/SystemInfo.cpp
IkerLuengo: addressed in f67682c

test/unittest/utils/SystemInfoTests.cpp
IkerLuengo: New tests in f67682c

src/cpp/utils/SystemInfo.cpp
JLBuenoLopez-eProsima: I have completed the doxygen documentation to include this new case in afb20af49b76a0b9f1eb3b9241f468218c87b0b0

IkerLuengo (approved):
LGTM

PR #2159, opened 2021-08-17
[12199] Refactor load environment server info to take into account environment file

JLBuenoLopez-eProsima:
This PR is built over #2157. It refactors `load_environment_server_info` to check first if the environment file exists and if so, read the environment info from there. It also adds some test cases that checks this new feature.

### DIFF
--- a/test/blackbox/common/BlackboxTestsDiscovery.cpp
+++ b/test/blackbox/common/BlackboxTestsDiscovery.cpp
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <stdlib.h>
+#endif // _WIN32
+
 #include "BlackboxTests.hpp"
 
 #include "PubSubWriterReader.hpp"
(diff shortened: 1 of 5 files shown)


test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -1319,4 +1325,38 @@ TEST(Discovery, ServerClientEnvironmentSetUp)
 
     ASSERT_TRUE(load_environment_server_info(text, output));
     ASSERT_EQ(output, standard);
+
+    // 11. Check loading from environment file
+    std::string filename = "../unittest/utils/environment_test_file.json";
```
IkerLuengo, line 1328: Better for this test to have its own configuration file, so that both tests (this blackbox and the unit test) become independent and can evolve separately.

IkerLuengo (changes requested):

test/blackbox/common/BlackboxTestsDiscovery.cpp
JLBuenoLopez-eProsima: Done in c3bea832946981b011fe8800677087471a68e1e1.

Good suggestion. As the environment file could be changed at runtime we would need to add a blackbox test to check this feature and could directly use this same file instead of modifying the one included in the unit test folder.

IkerLuengo (approved):
LGTM

PR #2170, opened 2021-08-27
Adding enable value to RTPS Participant [13981]

jparisu:
In case of TCP, when closing the application each locator is eliminated twice (second time without success).
This is because in `RTPSDomain::removeRTPSParticipant` is used `disabled()` and is used again in destruction.
These duplication of erasing a locator leads in some cases to a Warning (when the locator receives a data, but this locator has already been closed):
`[RTCP Warning] Received Message, but no TransportReceiverInterface attached: 11802 -> Function perform_listen_operation`

**No tests implemented.**

### DIFF
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
@@ -129,6 +129,7 @@ RTPSParticipantImpl::RTPSParticipantImpl(
     , mp_ResourceSemaphore(new Semaphore(0))
     , IdCounter(0)
     , type_check_fn_(nullptr)
+    , enabled_(false)
     , client_override_(false)
     , internal_metatraffic_locators_(false)
     , internal_default_locators_(false)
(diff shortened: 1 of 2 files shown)


jparisu:
> I think it would be better to use an `atomic_bool` and test-and-set functions.

I do not see the use of having an atomic variable that must be checked from a non atomic method

MiguelCompany:
@jparisu This may be related with #2171, could you check?

JLBuenoLopez-eProsima:
@jparisu Could you check if the merging of #2171 solved the issue? Is this PR still interesting to be merged? If so, please assign milestone v2.10.1. Otherwise, can you please close it?

jparisu:
I do not think this PR makes much sense as it is. It would avoid a possible dangerous data race, but it still leave some inconsistencies (as you call enable and function returns without enabling participant). Probably the correct way to do so is using a mutex to block enable and disable calls. However I do not know if you want to add a mutex lock here.
If you agree with it I can make the changes and could be ready soon.

JLBuenoLopez-eProsima:
@jsantiago-eProsima could you have a look?

PR #2173, opened 2021-08-30
Fixing TLS Exceptions and Example

jparisu:
The actual certificates does not work in `HelloWorldExampleTCP` using `TLS`.
`ca.pem` has been regenerated and used in Publisher side to check the certificate of the subscriber.


MiguelCompany (approved):
LGTM

PR #2175, opened 2021-08-31
Fix ABBA deadlock in flow controller core [12394]

richiware:
`RTPSMessageGroup` used by the flow controller could lead in a ABBA deadlock when the sender is changed. This PR fixes this. Instead of use the writer's mutex, the LocatorSelectorSender has its own mutex.

### DIFF
--- a/test/mock/rtps/ReaderLocator/fastdds/rtps/writer/ReaderLocator.h
+++ b/test/mock/rtps/ReaderLocator/fastdds/rtps/writer/ReaderLocator.h
@@ -233,7 +233,7 @@ public:
         return 0;
     }
 
-    void lock()
+    void lock() override
     {
     }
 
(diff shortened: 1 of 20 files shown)


test/mock/rtps/ReaderLocator/fastdds/rtps/writer/ReaderLocator.h
```diff
@@ -233,6 +233,14 @@ class ReaderLocator : public RTPSMessageSenderInterface
         return 0;
     }
 
+    void lock()
```
MiguelCompany, line 236:
```suggestion
    void lock() override
```

MiguelCompany (changes requested):
Just a small nit

MiguelCompany (approved):
LGTM with green CI!

PR #2176, opened 2021-08-31
Fix deadlock using asynchronous publication and statistics [12396]

richiware:
From a call of a user's writer `write()` function, internally is a call of a statistics's writer `write()` function. Using asynchronous publication mode, this second call will call again the FlowController's mutex leading into a deadlock. This PR creates another asynchronous thread specially for statistics writers.


src/cpp/statistics/fastdds/publisher/qos/DataWriterQos.cpp
```diff
@@ -31,6 +31,7 @@ DataWriterQos::DataWriterQos()
     reliability().kind = eprosima::fastdds::dds::RELIABLE_RELIABILITY_QOS;
     durability().kind = eprosima::fastdds::dds::TRANSIENT_LOCAL_DURABILITY_QOS;
     publish_mode().kind = eprosima::fastdds::dds::ASYNCHRONOUS_PUBLISH_MODE;
+    publish_mode().flow_controller_name = eprosima::fastdds::rtps::FASTDDS_STATISTICS_FLOW_CONTROLLER_DEFAULT;
```
MiguelCompany, line 34: This will need a documentation update

MiguelCompany (changes requested):

PR #2123, opened 2021-08-06
Move optionparser to thirdparty [12336]

jparisu:
Tested with CMake Options:
```
-DCMAKE_BUILD_TYPE=Release
-DCOMPILE_EXAMPLES=ON
-DGTEST_INDIVIDUAL=ON
-DFASTRTPS_API_TESTS=ON
-DFASTDDS_PIM_API_TESTS=ON
-DPERFORMANCE_TESTS=ON
-DEPROSIMA_BUILD_TESTS=ON
-DVIDEO_TESTS=ON
```

Signed-off-by: jparisu <javierparis@eprosima.com>

PR #2178, opened 2021-09-01
New DDS example: Discovery Server Example [12368]

juanlofer-eprosima:
Signed-off-by: juanlofer-eprosima juanlopez@eprosima.com


juanlofer-eprosima:
Merge after pull request #2123 has been merged.

MiguelCompany:
@richiprosima Please test mac

examples/C++/DDS/HelloWorldExampleDiscoveryServer/arg_configuration.h
```diff
@@ -0,0 +1,178 @@
...
+
+    static const std::regex ipv4;
+};
+
+enum  optionIndex
+{
+    UNKNOWN_OPT,
+    HELP,
+    TOPIC,
+    WAIT,
```
jparisu, line 130: NIT: 
Some arguments may make more sense in a Basic Example, but they have nothing to do with the Discovery Server.
Think if it would be better to remove them in order to make the example simpler (`WAIT`, less important but also could be removed: `INTERVAL`)

examples/C++/DDS/HelloWorldExampleDiscoveryServer/arg_configuration.h
```diff
@@ -0,0 +1,178 @@
...
+        {
+            print_error("Option '", option, "' requires an IPaddress[:portnumber] argument\n");
+        }
+        return option::ARG_ILLEGAL;
+    }
+
+    static const std::regex ipv4;
+};
+
+enum  optionIndex
```
jparisu, line 125: I would add new argument to set the **ID** of the server, in the server itself as in the pub & sub.
The way it is right now is simple and fine, but a user that use this example will not be able to work with its own discovery server without that configuration, and it is better to make it visible here.
Use the id as it is used in fastdds cli

examples/C++/DDS/HelloWorldExampleDiscoveryServer/arg_configuration.h
```diff
@@ -0,0 +1,178 @@
...
+
+    { UNKNOWN_OPT, 0, "", "",                Arg::None,      "\nSubscriber options:"},
+    { TOPIC, 0, "t", "topic",                  Arg::String,
+      "  -t <topic_name> \t--topic=<topic_name>  \tTopic name (Default: HelloWorldTopic)." },
+    { SAMPLES, 0, "s", "samples",              Arg::Numeric,
+      "  -s <num> \t--samples=<num>  \tNumber of samples to wait for (Default: 0 => infinite samples)." },
+    { LOCATOR, 0, "", "ip",                 Arg::Locator,
+      "  \t--ip=<IPaddress[:port number]>  \tServer address (Default address: 127.0.0.1, default port: 60006)." },
+
+    { UNKNOWN_OPT, 0, "", "",                Arg::None,      "\nDiscoveryServer options:"},
```
jparisu, line 162: It is not clear if this parameter is to set the listening address or the connection address.
I would use `ip` as connection (as in pub & sub) and use a new parameter to set its own locators

examples/C++/DDS/HelloWorldExampleDiscoveryServer/HelloWorld_main.cpp
```diff
@@ -0,0 +1,244 @@
...
+                                port = static_cast<uint16_t>(port_int);
+                            }
+                        }
+                    }
+
+                    if (!ip_address.empty() && port > 1000)
+                    {
+                        eprosima::fastrtps::rtps::IPLocator::setPhysicalPort(server_address, port);
+                        eprosima::fastrtps::rtps::IPLocator::setIPv4(server_address, ip_address);
+                    }
```
jparisu, line 189: Add logs in error cases

examples/C++/DDS/HelloWorldExampleDiscoveryServer/HelloWorld_main.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+    else
+    {
+        std::cerr << "ERROR: <publisher|subscriber|server> argument is required." << std::endl;
+        option::printUsage(fwrite, stdout, usage, columns);
+        return 1;
+    }
+
+    // Set default IP address if not specified
+    if (!IsAddressDefined(server_address))
```
jparisu, line 208: NIT: Why not setting an int and a string in the arguments, and here set the whole Locator (once the possible errors has been checked)

examples/C++/DDS/HelloWorldExampleDiscoveryServer/HelloWorldSubscriber.h
```diff
@@ -0,0 +1,129 @@
...
+
+    private:
+
+        HelloWorld hello_;
+
+        //! Number of DataWriters matched to the associated DataReader
+        int matched_;
+
+        //! Number of samples received
+        uint32_t samples_;
```
jparisu, line 112:
```suggestion
        uint32_t samples_received_;
```

examples/C++/DDS/HelloWorldExampleDiscoveryServer/HelloWorldSubscriber.cpp
```diff
@@ -0,0 +1,218 @@
...
+    DomainParticipantQos pqos;
+    pqos.name("Participant_sub");
+
+    // Set participant as DS CLIENT
+    pqos.wire_protocol().builtin.discovery_config.discoveryProtocol =
+            eprosima::fastrtps::rtps::DiscoveryProtocol_t::CLIENT;
+
+    // Set SERVER's GUID prefix
+    RemoteServerAttributes remote_server_att;
+    remote_server_att.ReadguidPrefix("44.53.00.5f.45.50.52.4f.53.49.4d.41");
```
jparisu, line 75: Get this GUID from where fast set it for the default discovery server

jparisu (changes requested):

JLBuenoLopez-eProsima:
@juanlofer-eprosima What is the status of this PR? The latest comment mentions that TCP is not working but maybe we can remove that option and merge it. Can you do it? Please, assign in that case the milestone v2.10.1

juanlofer-eprosima:
Closing in favor of https://github.com/eProsima/Fast-DDS/pull/2924

PR #2122, opened 2021-08-05
New DDS example: Basic Configuration HelloWorld [12329]

juanlofer-eprosima:
Signed-off-by: juanlofer-eprosima <juanlopez@eprosima.com>

PR #2181, opened 2021-09-02
Clean HelloWorldExample [12418]

jparisu:
Clean HelloWorldExample so it is an actual HelloWorld
In PR https://github.com/eProsima/Fast-DDS/pull/2122 a new more complex Example will be added to FastDDS


jparisu:
This example requires more than just a clean up

PR #2182, opened 2021-09-03
Keep changes inside instances sorted by source timestamp [12421]

MiguelCompany:
On a reliable reader, when using a topic with keys, and changes arrive unordered, they are returned in the received order instead of the sequence number order.

This PR fixes this misbehavior, and also adds a generic utility function to keep changes on a vector sorted.

### DIFF
--- a/src/cpp/utils/collections/sorted_vector_insert.hpp
+++ b/src/cpp/utils/collections/sorted_vector_insert.hpp
@@ -1,0 +1,59 @@
+// Copyright 2021 Proyectos y Sistemas de Mantenimiento SL (eProsima).
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @file sorted_vector_insert.hpp
+ */
+
+#ifndef SRC_CPP_UTILS_COLLECTIONS_SORTED_VECTOR_INSERT_HPP_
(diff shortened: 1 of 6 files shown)


MiguelCompany:
@mergifyio backport 2.3.x 2.2.x 2.1.x 2.0.x

MiguelCompany:

IkerLuengo (dismissed):
Just some syntactic sugar suggestion, feel free to dismiss it if we want to go for a quick merge.

src/cpp/utils/collections/sorted_vector_insert.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * @file sorted_vector_insert.hpp
+ */
+
+#ifndef SRC_CPP_UTILS_COLLECTIONS_SORTED_VECTOR_INSERT_HPP_
+#define SRC_CPP_UTILS_COLLECTIONS_SORTED_VECTOR_INSERT_HPP_
+
+#include <algorithm>
+#include <functional>
+
+namespace eprosima {
```
richiware, line 25: Why does it belong only to namespace `eprosima` instead of `eprosima::fastdds`?

MiguelCompany: It is a generic utility, and will be used from namespaces `eprosima::fastdds`, `eprosima::fastrtps`, and `eprosima::fastrtps::rtps`.

I considered the `eprosima::utilities::collections`, as we have on `unordered_vector.hpp`, but prefered to use the generic namespace directly

richiware: Fine for me, but it could collide in the future with other products, especially because this product is the base of other ones.

richiware (dismissed):

src/cpp/utils/collections/sorted_vector_insert.hpp
MiguelCompany: An important thing to note is that it is an internal header. It will not be part of the public API, as the header is under the `src` folder.

Nevertheless, we have discussed this offline, and we're going to put the method on the `eprosima::utilities::collections` namespace to be consistent with other internal headers.

richiware (approved):

MiguelCompany:

PR #2205, opened 2021-09-15
[12199] Reload environment file information

JLBuenoLopez-eProsima:
This PR implements the final functionality. Monitoring the environment file and applying the changes when the watcher triggers.

It also includes the refactor of `SystemInfo::get_env` to use `std::string` instead of `char*`.

### DIFF
--- a/src/cpp/rtps/RTPSDomain.cpp
+++ b/src/cpp/rtps/RTPSDomain.cpp
@@ -113,7 +113,7 @@ RTPSParticipant* RTPSDomain::createParticipant(
     }
 
     // Only the first time, initialize environment file watch if the corresponding environment variable is set
-    if (m_RTPSParticipantIDs.empty())
+    if (!RTPSDomainImpl::file_watch_handle_)
     {
         if (!SystemInfo::get_environment_file().empty())
         {
@@ -549,6 +549,7 @@ void RTPSDomainImpl::file_watch_callback()
     // Ensure that all changes have been saved by the OS
     std::this_thread::sleep_for(std::chrono::seconds(1));
     // For all RTPSParticipantImpl registered in the RTPSDomain, call RTPSParticipantImpl::environment_file_has_changed
+    std::lock_guard<std::mutex> guard(RTPSDomain::m_mutex);
     for (auto participant : RTPSDomain::m_RTPSParticipants)
     {
         participant.second->environment_file_has_changed();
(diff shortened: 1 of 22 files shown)


src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -526,6 +544,17 @@ bool RTPSDomainImpl::should_intraprocess_between(
     return false;
 }
 
+void RTPSDomainImpl::file_watch_callback()
+{
+    // Ensure that all changes have been saved by the OS
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    // For all RTPSParticipantImpl registered in the RTPSDomain, call RTPSParticipantImpl::environment_file_has_changed
+    for (auto participant : RTPSDomain::m_RTPSParticipants)
```
MiguelCompany, line 552: This should be protected by the same mutex used on `create_participant`

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -2138,6 +2168,26 @@ DurabilityKind_t RTPSParticipantImpl::get_persistence_durability_red_line(
     return durability_red_line;
 }
 
+void RTPSParticipantImpl::environment_file_has_changed()
+{
+    RTPSParticipantAttributes patt;
```
MiguelCompany, line 2173:
```suggestion
    RTPSParticipantAttributes patt = m_att;
```
Otherwise, we may be inadvertently changing userData.

MiguelCompany (changes requested):

src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -107,6 +112,17 @@ RTPSParticipant* RTPSDomain::createParticipant(
         return nullptr;
     }
 
+    // Only the first time, initialize environment file watch if the corresponding environment variable is set
+    if (m_RTPSParticipantIDs.empty())
```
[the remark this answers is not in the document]
JLBuenoLopez-eProsima: It is enough I think checking if the `unique_ptr` has been set.

MiguelCompany (approved):
LGTM with green CI!

MiguelCompany:
I have cancelled the aarch64 job, as we still have to do the final PR into master.

Other failures are seeing on nightlies from time to time.

PR #2207, opened 2021-09-17
Fix complex types with collections bug in Dynamic Types[12517]

jparisu:


jparisu:
STATUS:
This fixes some dynamic types issues, but it does not solve them all.
This requires more study and implementation.

PR #2214, opened 2021-09-20
Fix DDS Late Joiners example

JLBuenoLopez-eProsima:
DDS Lata Joiners example was calling non implemented API `DataReader::get_matched_publications` which is always returning `ReturnCode_t::RETCODE_UNSUPPORTED`. Thus the previous implementation blocks in a `do-while` block waiting for this API to return a non-empty vector. This PR introduces a sleep that ensures that most times the discovery phase will finish and the example is going to work as expected.


IkerLuengo (approved):
LGTM

PR #2221, opened 2021-09-22
[11914] Feature/pkcs11 support

IkerLuengo:
Add support for PKCS#11 format URIs for private keys.


IkerLuengo:
This should have been a draft, as is a WIP

PR #2226, opened 2021-09-24
Fix Dynamic Types key parser from XML [12549]

jparisu:
This hotfix solves an error on the XML types parser for Dynamic Types.
The `key` was not being parsed correctly, and so the type could not handle keys. 


jparisu:
Fixing the key error it appears a new one where the type is not correctly sent in discvoery TypeLookup. The Subscriber does receive the type as if it has not key, and so the match is incompatible.

JLBuenoLopez-eProsima:
@jparisu I am closing old drafts and obsolete PRs. Could you look into this one? Even if a second bug was discovered fixing this one, it makes sense to merge these changes. If so, transform into PR and assign milestone v2.10.1. Thanks!

jparisu:
Whole Dynamic Types module is being refactored and fixed. Thus this does not make much sense, as it does not fix the whole bug.

PR #2227, opened 2021-09-24
Communication improvement sending only one GAP for all readers [12576]

richiware:
Currently, when there is a hole in the history a GAP message is sent for each reader. This PR improves this behavior by sending a single GAP for everyone (useful when readers are configured in multicast).


MiguelCompany (approved):
LGTM

MiguelCompany:
@richiprosima Please test mac and test linux

PR #2229, opened 2021-09-24
Avoid bad_node_size exception when cross building [12583]

MiguelCompany:
When foonathan/memory is cross-compiled, and the `node_size_dbg` tool cannot be executed with an emulator, some custom node types are used to calculate the size of the nodes being requested to the allocator.

After some reports of a `bad_node_size` exception being thrown, I investigated this deeply, and found the custom type was not compliant with certain std implementations (libstdc++).


foonathan:
This will permanently fix foonathan/memory#103, foonathan/memory#123.

MiguelCompany:
@mergifyio backport 2.3.x 2.2.x 2.1.x

IkerLuengo (approved):
LGTM

PR #2235, opened 2021-09-27
xTypes Example [12531]

jparisu:
New example using xTypes loading an .idl file to create dynamically the type.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,11 @@ endif()
 
 include_directories(thirdparty/nlohmann-json)
 
+# Not required packages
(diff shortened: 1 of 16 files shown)


jparisu:
STATUS:
This example requires a generic serializator and deserializator for xtypes objects that does not exist.
Thus, it is required to serialize and deserialize each xtypes type that is being used (what makes it useless).

JLBuenoLopez-eProsima:
@jparisu Does it make sense to keep this open?

jparisu:
It does not while there is no xTypes serializator. And knowing the status of xtypes package it will not be ready soon. Thus I close the PR.

PR #2240, opened 2021-09-30
Improve `DataWriterImpl::unregister_instance` performance and standard compliance [12622]

JLBuenoLopez-eProsima:
Only calculate the instance's key if the `handle` is `HANDLE_NIL` in Release.

In case of inconsistency between the `handle` and the instance's key, the standard establishes that it is a `RETCODE_PRECONDITION_NOT_MET`. Before this PR, `RETCODE_BAD_PARAMETER` is being returned.

Add DDS API unittest for `DataWriter::dispose` and `DataWriter::unregister_instance`.


MiguelCompany:

PR #2243, opened 2021-10-01
Fix buffers exhaution when compiled with statistics [12629]

juanlofer-eprosima:
Commit 14d33232bcd2252e6470291d583b9afa444c55f4 broke init_monitor_tests from Fast-DDS-statistics-backend.
The reason is that, when compiled with statistics, a new async flow controller is created but the same number of buffers are preallocated, leading to a resource exhaution and falling into a hangup state.

### DIFF
--- a/fastrtps.repos
+++ b/fastrtps.repos
@@ -10,7 +10,7 @@ repositories:
     fastrtps:
         type: git
         url: https://github.com/eProsima/Fast-DDS.git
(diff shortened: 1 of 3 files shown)


MiguelCompany (dismissed):
LGTM with greenish CI!

MiguelCompany (approved):
👍 Much nicer solution!

PR #2248, opened 2021-10-05
[foxy] Fix: gap not being sent [14675]

hexonxons:
* Port small SHM segment warning message
* Fix bug with gap not being sent with heartbeat message

This PR fixes a problem with my service test: https://github.com/ros2/rmw_fastrtps/issues/559
That test will eventually fail cuz of client won't receive response from service. Right now gaps will be sent after reader-writer match, but there could be a kind of race:
* Writer matches a reader in the service
* Writer sends heartbeat and gap messages
* Reader matches a writer in the client

In that case initial gap message will be lost and reader will be asking for sequence set from heartbeat forever.

Seems like that issue is already fixed in master branch.

### DIFF
--- a/include/fastdds/rtps/writer/ReaderProxy.h
+++ b/include/fastdds/rtps/writer/ReaderProxy.h
@@ -48,6 +48,7 @@ namespace rtps {
 class StatefulWriter;
 class TimedEvent;
 class RTPSReader;
+class RTPSGapBuilder;
 
 /**
  * ReaderProxy class that helps to keep the state of a specific Reader with respect to the RTPSWriter.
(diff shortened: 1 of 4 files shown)


JLBuenoLopez-eProsima:
ROS 2 Foxy is using `2.1.x` now and `2.0.x` is becoming EOL (end of life) this month, so I am going to redirect this PR to be merged into `2.1.x` and add the `2.1.2` milestone.

JLBuenoLopez-eProsima:
@hexonxons,

Would you mind rebasing your branch into `2.1.x` and drop all commits with the exception of the two of the original PR. If it is too much trouble, let me know so I can prepare a new PR replacing this one. Thanks for your contribution and sorry for the late reply!

hexonxons:
@JLBuenoLopez-eProsima Done

That fix was added in https://github.com/eProsima/Fast-DDS/commit/6b55651b6edfaf0a201fce26e1861ceef84696a4, but I'm unsure if it was fixed intentionally or accidentally. Also I can't find original commit when that code was changed.

hexonxons:
Also same fix for galactic: https://github.com/eProsima/Fast-DDS/pull/2519

richiware (approved):

PR #2249, opened 2021-10-05
Do not filter Error messages category

jparisu:
"I am no longer accepting the things I cannot change. I'm changing the things I cannot accept" - Angela Davis - :mechanical_arm: 


jparisu:
Finally there was not consensus about this useful feature.

PR #2254, opened 2021-10-08
Avoid sending GAP for no current history holes [12643]

richiware:
Current mechanism could send GAP about missing samples but they are not currently a hole in the history. This information already will be in the HEARTBEAT message.

This PR avoid to send GAP of this kind of samples which their sequence number is less than minimum sequence number in the history.


IkerLuengo (dismissed):
Failing tests are not related to the changes.
LGTM.

MiguelCompany (approved):
LGTM

MiguelCompany:

MiguelCompany:

PR #2261, opened 2021-10-13
Exported some function on Windows [12673]

richiware:
When generate source code with TypeObjects, on Windows the user application doesn't compile because some needed functions are not exported. This PR fixes this.


MiguelCompany:

PR #2262, opened 2021-10-13
Add support for Key annotation in TypeObjectFactory [12674]

richiware:
`TypeObjectFactory` currently only supports `@key`, but not `@Key`. This PR fixes this.


richiware:
> Changes look good to me, though "Key" is not a standard annotation. If we consider it as a valid extension, then a test for it should be added.

I've increased an already existing unit test. I've tested this addition fails on current master branch.

MiguelCompany (approved):
LGTM

PR #2280, opened 2021-10-19
[12529] Fix file watch initialization

JLBuenoLopez-eProsima:
If the `FASTDDS_ENVIRONMENT_FILE` environment variable is set but the file does not exist, Fast DDS breaks. This PR fixes this issue and adds also a test to check different Discovery Server configurations.

### DIFF
--- a/test/unittest/dds/participant/ParticipantTests.cpp
+++ b/test/unittest/dds/participant/ParticipantTests.cpp
@@ -663,6 +663,7 @@ TEST(ParticipantTests, RemoteServersListConfiguration)
     server.metatrafficUnicastLocatorList.push_back(locator);
     get_server_client_default_guidPrefix(1, server.guidPrefix);
     output.push_back(server);
+#ifndef __APPLE__
     std::this_thread::sleep_for(std::chrono::milliseconds(1100));
     file.open(filename);
     file << "{\"ROS_DISCOVERY_SERVER\": \"84.22.253.128:8888;192.168.1.133:64863;localhost:1234\"}";
@@ -670,6 +671,7 @@ TEST(ParticipantTests, RemoteServersListConfiguration)
     std::this_thread::sleep_for(std::chrono::milliseconds(1100));
     get_rtps_attributes(participant, attributes);
     EXPECT_EQ(attributes.builtin.discovery_config.m_DiscoveryServers, output);
+#endif // APPLE
     result_qos = participant->get_qos();
     EXPECT_EQ(ReturnCode_t::RETCODE_OK, participant->set_qos(result_qos));
     EXPECT_EQ(ReturnCode_t::RETCODE_OK, DomainParticipantFactory::get_instance()->delete_participant(participant));
@@ -748,7 +750,9 @@ TEST(ParticipantTests, RemoteServersListConfiguration)
     get_rtps_attributes(participant, attributes);
     EXPECT_EQ(attributes.builtin.discovery_config.m_DiscoveryServers, qos_output);
     // Capture log warning
+#ifndef __APPLE__
     helper_wait_for_at_least_entries(mockConsumer, 1);
+#endif // APPLE
     result_qos = participant->get_qos();
     EXPECT_EQ(ReturnCode_t::RETCODE_OK, participant->set_qos(result_qos));
     EXPECT_EQ(ReturnCode_t::RETCODE_OK, DomainParticipantFactory::get_instance()->delete_participant(participant));
@@ -789,6 +793,7 @@ TEST(ParticipantTests, RemoteServersListConfiguration)
     participant = DomainParticipantFactory::get_instance()->create_participant(0, qos);
     ASSERT_NE(nullptr, participant);
     // Try adding a new remote server
+#ifndef __APPLE__
     std::this_thread::sleep_for(std::chrono::milliseconds(1100));
     file.open(filename);
     file << "{\"ROS_DISCOVERY_SERVER\": \"172.17.0.5:4321;192.168.1.133:64863\"}";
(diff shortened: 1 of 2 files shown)


JLBuenoLopez-eProsima:
File watch is a feature not implemented yet for MacOS. I have fixed the new test so it also works in MacOS not running the checks related with the addition of dynamic servers.

jsantiago-eProsima (changes requested):
Just two little nitpicks.

As a side note, maybe this new test could have been split into a few smaller tests.

test/unittest/dds/participant/ParticipantTests.cpp
```diff
@@ -512,6 +527,364 @@ TEST(ParticipantTests, ChangePSMDomainParticipantQos)
...
+    Log::ClearConsumers();
+    MockConsumer* mockConsumer = new MockConsumer("RTPS_QOS_CHECK");
+    Log::RegisterConsumer(std::unique_ptr<LogConsumer>(mockConsumer));
+    Log::SetVerbosity(Log::Warning);
+
+    // Set environment variables
+    /**
+     * FASTDDS_ENVIRONMENT_FILE_ENV_VAR should be defined from the start because it is only loaded once (when the XML
+     * profiles are loaded). The file is not going to exist until a later case. However, the filewatch is initialized
+     * even though the file does not exist ()
```
jsantiago-eProsima, line 585: According to comments later in this test and this check:

https://github.com/eProsima/Fast-DDS/blob/4f6ca2a3dcdd57f0e8537b086c09df916af5ab98/src/cpp/rtps/RTPSDomain.cpp#L120

It looks like the filewatch is in fact NOT initialized if the file does not exist. If that is so please correct this comment.

jsantiago-eProsima (changes requested):
A last minute comment that was not included in the previous review

test/unittest/dds/participant/ParticipantTests.cpp
```diff
@@ -36,16 +44,23 @@
 #include <fastdds/dds/subscriber/Subscriber.hpp>
 #include <fastdds/dds/topic/qos/TopicQos.hpp>
 #include <fastdds/rtps/attributes/RTPSParticipantAttributes.h>
+#include <fastdds/rtps/attributes/ServerAttributes.h>
+#include <fastdds/rtps/common/Locator.h>
+#include <fastdds/rtps/participant/RTPSParticipant.h>
```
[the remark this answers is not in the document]
JLBuenoLopez-eProsima: It is not being used. Good catch!

test/unittest/dds/participant/ParticipantTests.cpp
JLBuenoLopez-eProsima: This comment does not apply any longer as the test has been split into smaller tests

test/unittest/dds/participant/ParticipantTests.cpp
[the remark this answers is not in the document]
JLBuenoLopez-eProsima: It is necessary because `RTPSParticipant::getRTPSParticipantAttributes` is defined within that header file. Without this include, the project does not build.

test/unittest/dds/participant/ParticipantTests.cpp
```diff
@@ -799,30 +894,44 @@ TEST(ParticipantTests, RemoteServersListConfiguration)
     file << "{\"ROS_DISCOVERY_SERVER\": \"172.17.0.5:4321;192.168.1.133:64863\"}";
```
jsantiago-eProsima, line 894: Shouldn't this fail since we already defined a Discovery Server via environment variable earlier and it is not included here?

jsantiago-eProsima (changes requested):

test/unittest/dds/participant/ParticipantTests.cpp
```diff
@@ -610,133 +585,249 @@ TEST(ParticipantTests, RemoteServersListConfiguration)
...
-    get_server_client_default_guidPrefix(1, server.guidPrefix);
-    output.push_back(server);
 #ifndef __APPLE__
     std::this_thread::sleep_for(std::chrono::milliseconds(1100));
     file.open(filename);
     file << "{\"ROS_DISCOVERY_SERVER\": \"84.22.253.128:8888;192.168.1.133:64863;localhost:1234\"}";
     file.close();
     std::this_thread::sleep_for(std::chrono::milliseconds(1100));
     get_rtps_attributes(participant, attributes);
+
```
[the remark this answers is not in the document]
JLBuenoLopez-eProsima: The attributes have indeed been modified. The variable `output` used to validate the expected attributes output is the one we need to update so the check is correct.

test/unittest/dds/participant/ParticipantTests.cpp
JLBuenoLopez-eProsima: That is just the issue this test intends to check: what happens when two servers with the same guidPrefix are being added, one through the environment variable and another through QoS. It seems that both are included though only one of them is pinged (as the test comment explains). Also, the attributes only holds the one set by QoS (which is not the one being pinged as far as I know checking with a wireshark capture) and it is against this server that the feature checks that no server has been removed. That is the reason why this test is called inconsistent. The main goal was checking that Fast DDS does not have a core and that any mishandle of the API is safe (even though the behavior can be undefined). That is the reason why in [our documentation](https://fast-dds.docs.eprosima.com/en/latest/fastdds/discovery/discovery_server.html#modifying-remote-servers-list-at-run-time) we have included a note stating that it is advised to use either the API or the environment variable but not both at the same time.

jsantiago-eProsima (approved):
LGTM.

MiguelCompany:
Just one more CI run.

MiguelCompany:

PR #2286, opened 2021-10-22
Adding content filter interfaces

MiguelCompany:
Depends on #2311, and should be rebased after that one is merged.


MiguelCompany:

include/fastdds/rtps/reader/RTPSReader.h
```diff
@@ -251,6 +252,12 @@ class RTPSReader
         return mp_history;
     }
 
+    RTPS_DllAPI void set_content_filter(
+            eprosima::fastdds::rtps::IReaderDataFilter* filter)
+    {
```
EduPonz, line 257: Maybe we can lock the mutex here. What'd happen if the user sets a `nullptr` filter while we are operating with it?

include/fastdds/rtps/writer/IReaderDataFilter.hpp
```diff
@@ -14,48 +14,7 @@
...
-
-};
-
-} /* namespace rtps */
-} /* namespace fastdds */
-} /* namespace eprosima */
-
-#endif /* _FASTDDS_RTPS_IREADERDATA_FILTER_H_ */
+// This file was moved as part of the implementation of content filtered topics
+#include <fastdds/rtps/IReaderDataFilter.hpp>
```
EduPonz, line 20: Can we mark this as deprecated/to-be-removed-by? I wouldn't want to have to keep this file here for ever

src/cpp/fastdds/topic/TopicDescriptionImpl.hpp
```diff
@@ -57,7 +56,10 @@ class TopicDescriptionImpl
         --num_refs_;
     }
 
+    virtual const std::string& get_rtps_topic_name() const = 0;
```
EduPonz, line 59: Should we add this to the design?

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -238,6 +240,13 @@ ReturnCode_t DataReaderImpl::enable()
         return ReturnCode_t::RETCODE_ERROR;
     }
 
+    auto topic_impl = topic_->get_impl();
```
EduPonz, line 243: You're getting the `topic_impl` twice (the other being the setting the property). Moving this line up we can do it just once

EduPonz (changes requested):
Partial review. Here are there we are missing includes for the "include what you use" principle, mainly std ones. Furthermore, you'd need to update statistics mock for this to compile with statistics

Reviews pending:
* `ContentFilter` interfaces
* `ContentFilteredTopic`
* `RTPSReader`s intergration

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -223,6 +226,8 @@ DomainParticipantImpl::~DomainParticipantImpl()
     {
         std::lock_guard<std::mutex> lock(mtx_topics_);
 
+        filtered_topics_.clear();
```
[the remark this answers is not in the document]
MiguelCompany: This is to be consistent with the rest of the method, and it is not harmful by itself.

include/fastdds/dds/domain/DomainParticipant.hpp
```diff
@@ -20,23 +20,23 @@
 #ifndef _FASTDDS_DOMAIN_PARTICIPANT_HPP_
 #define _FASTDDS_DOMAIN_PARTICIPANT_HPP_
 
-#include <fastdds/dds/topic/TypeSupport.hpp>
-#include <fastdds/dds/topic/Topic.hpp>
-#include <fastdds/rtps/common/Time_t.h>
-#include <fastrtps/types/TypeIdentifier.h>
+#include <utility>
```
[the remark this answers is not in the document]
MiguelCompany: Also `<functional>` and `<vector>`

include/fastdds/dds/domain/DomainParticipant.hpp
```diff
@@ -720,6 +734,72 @@ class DomainParticipant : public Entity
...
+     *                        unique within this DomainParticipant.
+     * @param filter_factory  Factory of content filters to be registered. Cannot be nullptr.
+     *
+     * @return RETCODE_BAD_PARAMETER if any parameter is nullptr, or the filter_name exceeds 255 characters.
+     * @return RERCODE_PRECONDITION_NOT_MET if the filter_name has been already registered.
+     * @return RETCODE_OK if the filter is correctly registered.
+     */
+    RTPS_DllAPI ReturnCode_t register_content_filter(
+            const char* filter_name,
+            const IContentFilterFactory* filter_factory);
```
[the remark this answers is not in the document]
MiguelCompany: We register a factory for a certain filter name/class. So I propose changing the method names (appending `_factory`), but either leave the parameter name as is, or change it to `filter_class_name`

include/fastdds/dds/domain/DomainParticipant.hpp
```diff
@@ -720,6 +734,72 @@ class DomainParticipant : public Entity
...
+
+    /**
+     * Lookup a custom content filter factory previously registered with register_content_filter.
```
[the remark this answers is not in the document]
MiguelCompany: I'm not sure about this. If we allow to get it, should we also allow to register/unregister it? This will allow the user to override the default filter factory, extending it. But it could potentially have a big and dangerous impact.

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -458,6 +463,104 @@ ReturnCode_t DomainParticipantImpl::delete_topic(
...
+    // Tell filter factory to compile the expression
+    IContentFilter* filter_instance = nullptr;
+    if (ReturnCode_t::RETCODE_OK != filter_factory->create_content_filter(related_topic->get_type_name().c_str(),
+            type_descriptor, filter_expression.c_str(), filter_parameters, filter_instance))
+    {
+        return nullptr;
+    }
+
+    ContentFilteredTopic* topic;
+    topic = new ContentFilteredTopic(name, related_topic, filter_expression, expression_parameters);
```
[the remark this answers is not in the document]
MiguelCompany: That would exceed the line length limit. I prefer having it in two lines to avoid breaking the line in the middle of the parameter list.

src/cpp/fastdds/domain/DomainParticipantImpl.cpp
```diff
@@ -458,6 +463,104 @@ ReturnCode_t DomainParticipantImpl::delete_topic(
...
+    {
+        if (it->second->get_impl()->is_referenced())
+        {
+            return ReturnCode_t::RETCODE_PRECONDITION_NOT_MET;
+        }
+        filtered_topics_.erase(it);
+        return ReturnCode_t::RETCODE_OK;
+    }
+
+    return ReturnCode_t::RETCODE_BAD_PARAMETER;
```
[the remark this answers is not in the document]
MiguelCompany: Should be `RETCODE_PRECONDITION_NOT_MET` => `if the topic does not belong to this participant`

include/fastdds/rtps/writer/IReaderDataFilter.hpp
MiguelCompany: I've added a `FASTDDS_TODO_BEFORE`. It will not warn the user about it, but it will fail when we get to v3.0.

include/fastdds/dds/domain/DomainParticipant.hpp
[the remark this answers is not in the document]
MiguelCompany: It was finally decided that the user cannot fiddle with `FASTDDS_SQLFILTER_NAME`.

MiguelCompany:
@EduPonz I think I addressed all your comments. Please review.

include/fastdds/dds/topic/ContentFilteredTopic.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+constexpr const char* const sqlfilter_name = "DDSSQL";
+
+/**
+ * Specialization of TopicDescription that allows for content-based subscriptions
+ * @ingroup FASTDDS_MODULE
+ */
+class ContentFilteredTopic : public TopicDescription
+{
+    friend class ContentFilteredTopicImpl;
```
EduPonz, line 46: Is this friend necessary?

include/fastdds/dds/topic/ContentFilteredTopic.hpp
```diff
@@ -0,0 +1,103 @@
...
+    friend class DomainParticipantImpl;
+
+    RTPS_DllAPI ContentFilteredTopic(
```
EduPonz, line 49: I'd be explicit about `private` section

src/cpp/rtps/reader/StatefulReader.cpp
```diff
@@ -623,6 +632,10 @@ bool StatefulReader::processDataFragMsg(
             {
                 mp_history->completed_change(work_change);
                 pWP->received_change_set(work_change->sequenceNumber);
+                if (data_filter_ && !data_filter_->is_relevant(*work_change, m_guid))
+                {
+                    mp_history->remove_change(work_change);
```
EduPonz, line 637: Shouldn't we return without notifying?

EduPonz (changes requested):

src/cpp/rtps/reader/StatefulReader.cpp
MiguelCompany: No, because we could have filtered a change that is just before another one that should be notified.

include/fastdds/dds/topic/ContentFilteredTopic.hpp
MiguelCompany: I don't think so

src/cpp/fastdds/topic/ContentFilteredTopic.cpp
```diff
@@ -0,0 +1,107 @@
...
+ContentFilteredTopic::~ContentFilteredTopic()
+{
+    impl_->related_topic->get_impl()->dereference();
+    impl_->filter_factory->delete_content_filter(impl_->filter_class_name.c_str(), impl_->filter_instance);
+    delete impl_;
+}
+
+Topic* ContentFilteredTopic::get_related_topic() const
+{
+    return impl_->related_topic;
```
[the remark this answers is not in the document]
MiguelCompany: I don't think it makes sense for an implementation class.

include/fastdds/dds/domain/DomainParticipant.hpp
```diff
@@ -720,6 +734,72 @@ class DomainParticipant : public Entity
             const std::string& type_name,
             std::function<void(const std::string& name, const fastrtps::types::DynamicType_ptr type)>& callback);
 
+    /**
+     * Register a custom content filter factory, which can be used to create a ContentFilteredTopic.
+     *
+     * DDS specifies a SQL-like content filter for use by content filtered topics.
+     * If this filter does not meet your filtering requirements, you can register a custom filter.
```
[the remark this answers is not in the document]
EduPonz: Please address this
```suggestion
     * If this filter does not meet your filtering requirements, you can register a custom filter factory.
```

EduPonz (changes requested):

EduPonz (approved):
LGTM

PR #2311, opened 2021-11-10
Avoid payload reservation and copy when change is going to be rejected by history [13120]

MiguelCompany:
This PR adds a mechanism on ReaderHistory to check if a new change will be accepted into it. This allows the code on `processDataMsg` to avoid copying the change information (and, possibly, its payload) for a sample that is going to be rejected

### DIFF
--- a/src/cpp/rtps/reader/StatefulReader.cpp
+++ b/src/cpp/rtps/reader/StatefulReader.cpp
@@ -456,7 +456,8 @@ bool StatefulReader::processDataMsg(
 
             size_t unknown_missing_changes_up_to = pWP ? pWP->unknown_missing_changes_up_to(change->sequenceNumber) : 0;
             bool will_never_be_accepted = false;
-            if (!mp_history->can_change_be_added_nts(change, unknown_missing_changes_up_to, will_never_be_accepted))
+            if (!mp_history->can_change_be_added_nts(change->writerGUID, change->serializedPayload.length,
+                    unknown_missing_changes_up_to, will_never_be_accepted))
             {
                 if (will_never_be_accepted && pWP)
                 {
--- a/src/cpp/rtps/reader/StatefulReader.cpp
+++ b/src/cpp/rtps/reader/StatefulReader.cpp
@@ -562,6 +562,17 @@ bool StatefulReader::processDataFragMsg(
                     IDSTRING "Trying to add fragment " << incomingChange->sequenceNumber.to64long() << " TO reader: " <<
                     getGuid().entityId);
 
+            size_t changes_up_to = pWP->unknown_missing_changes_up_to(incomingChange->sequenceNumber);
+            bool will_never_be_accepted = false;
+            if (!mp_history->can_change_be_added_nts(incomingChange->writerGUID, sampleSize, changes_up_to, will_never_be_accepted))
+            {
+                if (will_never_be_accepted)
+                {
+                    pWP->irrelevant_change_set(incomingChange->sequenceNumber);
+                }
+                return false;
+            }
+
             CacheChange_t* change_to_add = incomingChange;
 
             CacheChange_t* change_created = nullptr;
@@ -595,7 +606,6 @@ bool StatefulReader::processDataFragMsg(
             // If this is the first time we have received fragments for this change, add it to history
             if (change_created != nullptr)
             {
-                size_t changes_up_to = pWP->unknown_missing_changes_up_to(change_created->sequenceNumber);
                 if (!change_received(change_created, pWP, changes_up_to))
                 {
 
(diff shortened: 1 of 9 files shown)


MiguelCompany:

src/cpp/rtps/reader/StatefulReader.cpp
```diff
@@ -446,58 +446,63 @@ bool StatefulReader::processDataMsg(
             logInfo(RTPS_MSG_IN,
                     IDSTRING "Trying to add change " << change->sequenceNumber << " TO reader: " << getGuid().entityId);
 
-            // Ask the pool for a cache change
-            CacheChange_t* change_to_add = nullptr;
-            if (!change_pool_->reserve_cache(change_to_add))
+            size_t unknown_missing_changes_up_to = pWP ? pWP->unknown_missing_changes_up_to(change->sequenceNumber) : 0;
+            if (mp_history->can_change_be_added_nts(change, unknown_missing_changes_up_to))
```
richiware, line 450: Before this PR, when `change_received` fails, `false` is returned. But now, if `can_change_be_added_nts` fails, `true` is returned. Is this correct?

src/cpp/rtps/reader/StatelessReader.cpp
```diff
@@ -398,65 +398,70 @@ bool StatelessReader::processDataMsg(
     {
         logInfo(RTPS_MSG_IN, IDSTRING "Trying to add change " << change->sequenceNumber << " TO reader: " << m_guid);
 
-        // Ask the pool for a cache change
-        CacheChange_t* change_to_add = nullptr;
-        if (!change_pool_->reserve_cache(change_to_add))
+        // Check rejection by history
+        if (mp_history->can_change_be_added_nts(change, 0) &&
```
richiware, line 402: Same comment than in `StatefulReader`.

test/mock/rtps/ReaderHistory/fastdds/rtps/history/ReaderHistory.h
```diff
@@ -72,6 +72,13 @@ class ReaderHistory
         return ret;
     }
 
+    virtual bool can_change_be_added_nts(
```
richiware, line 75: No plan for a test were this funtion fails?

src/cpp/rtps/reader/StatefulReader.cpp
MiguelCompany: Fixed on e17467fc61f10f5f4b1c1e38852d72123cf1bd3a

src/cpp/rtps/reader/StatelessReader.cpp
MiguelCompany: Fixed on e17467fc61f10f5f4b1c1e38852d72123cf1bd3a

test/mock/rtps/ReaderHistory/fastdds/rtps/history/ReaderHistory.h
MiguelCompany: I think it should be implicitly checked by some of the tests using KEEP_ALL policies

src/cpp/rtps/reader/StatefulReader.cpp
```diff
@@ -569,7 +594,8 @@ bool StatefulReader::processDataFragMsg(
             // If this is the first time we have received fragments for this change, add it to history
             if (change_created != nullptr)
```
richiware, line 595: Why don't use `can_change_be_added_nts` when accepting first fragment of a CacheChange_t?

src/cpp/rtps/reader/StatelessReader.cpp
```diff
@@ -481,6 +500,14 @@ bool StatelessReader::processDataFragMsg(
...
         if (writer.guid == writer_guid)
         {
+            // Always assert liveliness on scope exit
+            auto assert_liveliness_lambda = [&lock, this, &writer_guid](void*)
+                    {
+                        lock.unlock(); // Avoid deadlock with LivelinessManager.
+                        assert_writer_liveliness(writer_guid);
+                    };
+            std::unique_ptr<void, decltype(assert_liveliness_lambda)> p{ this, assert_liveliness_lambda };
+
```
richiware, line 510: As in `StatefulReader`. why don't use new function here?

src/cpp/rtps/reader/StatefulReader.cpp
MiguelCompany: The `incomingChange` received on the argument has the payload size of the fragment, so we cannot pass it to `can_change_be_added_nts`. And calling it on the line where you have put your comment is not necessary, since `change_received` is being called.

richiware: Sorry, I wrote the comment where Github allows me. I cannot write it in code without changes.

This function receives the `sampleSize`. Cannot be used? Extend your new function to support it.

MiguelCompany: Complete refactor of the new function to just accept the necessary input

richiware (approved):

PR #2328, opened 2021-11-25
Fix read after free on security code when remote participant is gone before an endpoint is matched [13035]

richiware:
When a remote writer is discovered (SecurityManager::discovered_writer), the cryptohandle for this remote writer is
stored in local_reader->second.associated_writers.
But the writer is not matched with the local reader after the key exchange process finish.
If meantime the participant is unregistered, all crypthandle of its remote endpoints are removed.
But the local_reader still maintains a pointer, because this pointer is only removed when the reader unmatched with the
remote writer.
Then, when the local participant is deleted, SecurityManager access to this invalid reference.

This PR fixes this.


MiguelCompany (approved):
LGTM

MiguelCompany:
Unrelated failures

MiguelCompany:
@mergifyio backport 2.3.x

PR #2336, opened 2021-12-02
Add participant locators if new network interfaces are found [12753]

JLBuenoLopez-eProsima:
Currently only network interfaces found when launching Fast DDS are taken into account. This PR adds new participant locators if new network interfaces are detected. In order to trigger the check, the user must call `DomainParticipant::set_qos`. If some user is using the RTPS layer directly, the feature can also be triggered by calling `RTPSParticipant::update_attributes`.

This PR includes both the implementation and the Blackbox tests at RTPS and DDS level. It also completes test API, both in the Entities as in the Transport layer.

Related PRs:
* eProsima/Fast-DDS-docs#301

### DIFF
--- a/test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
+++ b/test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
@@ -15,6 +15,7 @@
 #include <chrono>
 #include <cstdlib>
 #include <ctime>
+#include <memory>
 #include <string>
 
 #include <gtest/gtest.h>
@@ -22,9 +23,12 @@
 #include "BlackboxTests.hpp"
 #include "PubSubParticipant.hpp"
 #include "PubSubReader.hpp"
+#include "PubSubWriter.hpp"
 
 #include <fastdds/dds/core/policy/QosPolicies.hpp>
 #include <fastdds/rtps/common/Locator.h>
+#include <fastdds/rtps/transport/test_UDPv4TransportDescriptor.h>
+#include <rtps/transport/test_UDPv4Transport.h>
 #include <utils/SystemInfo.hpp>
 
 // Regression test for redmine issue 11857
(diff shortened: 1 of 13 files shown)


test/blackbox/common/RTPSWithRegistrationWriter.hpp
```diff
@@ -400,6 +407,16 @@ class RTPSWithRegistrationWriter
         return *this;
     }
 
+    bool is_matched() const
```
EduPonz, line 410: I think it might be useful in the future to have a `get_matched()` instead, and then make the comparison in the tests; same goes for the reader

test/blackbox/common/RTPSBlackboxTestsBasic.cpp
```diff
@@ -568,6 +569,64 @@ TEST_P(RTPS, RTPSAsReliableTransientLocalTwoWritersConsecutives)
...
+    writer.disable_builtin_transport().add_user_transport_to_pparams(test_transport).init();
+    ASSERT_TRUE(writer.isInitialized());
+
+    // no discovery
+    writer.wait_discovery(std::chrono::seconds(3));
+    reader.wait_discovery(std::chrono::seconds(3));
+
+    EXPECT_FALSE(writer.is_matched());
+    EXPECT_FALSE(reader.is_matched());
+
```
EduPonz, line 605: I'm thinking on whether we should send here and wait in the reader for a while, then reload the interfaces and wait on the reader again. First time nothing will be received, second everything. What do you think?

test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
```diff
@@ -168,3 +172,61 @@ TEST(DDSDiscovery, AddDiscoveryServerToList)
...
+    datawriter.disable_builtin_transport().add_user_transport_to_pparams(test_transport).init();
+    ASSERT_TRUE(datawriter.isInitialized());
+
+    // no discovery
+    datawriter.wait_discovery(std::chrono::seconds(3));
+    datareader.wait_discovery(std::chrono::seconds(3));
+
+    EXPECT_FALSE(datawriter.is_matched());
+    EXPECT_FALSE(datareader.is_matched());
+
```
EduPonz, line 210: Same as in the other test, should we send before reloading the interfaces?

EduPonz (changes requested):

test/blackbox/common/RTPSBlackboxTestsBasic.cpp
JLBuenoLopez-eProsima: If there is no matching there is not going to be any communication. So if the reader is `VOLATILE` as it is now (default) even though the interface is enabled later and discovery succeeds, no data is going to be received in the reader.

JLBuenoLopez-eProsima: Data is going to be sent and the reader is going to be `TRANSIENT_LOCAL` to access the already sent data

test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
```diff
@@ -168,3 +172,65 @@ TEST(DDSDiscovery, AddDiscoveryServerToList)
     server_2.wait_discovery(std::chrono::seconds::zero(), 2, true);
 }
 
+/**
+ * This test checks the addition of network interfaces at run-time.
```
EduPonz, line 176: You'll need to update the description. Also, some steps in the tests are commented and some aren't. Same goes for the RTPS test

EduPonz (changes requested):

EduPonz (approved):
LGTM

MiguelCompany:

MiguelCompany:

MiguelCompany:

MiguelCompany:

EduPonz (approved):

PR #2339, opened 2021-12-03
Fix oss-fuzz 40243 [13161]

MiguelCompany:
Fixed a null-dereference on XML parser when a type has no `name` attribute


src/cpp/rtps/xmlparser/XMLDynamicParser.cpp
```diff
@@ -606,6 +606,10 @@ XMLP_ret XMLParser::parseXMLBitmaskDynamicType(
     }
 
     const char* name = p_root->Attribute(NAME);
+    if (!name)
```
jsantiago-eProsima, line 609: I would test against nullptr just for consistency with the other test in this code block.

jsantiago-eProsima (changes requested):
Small nitpick

src/cpp/rtps/xmlparser/XMLDynamicParser.cpp
MiguelCompany: Changed

jsantiago-eProsima (approved):
LGTM

MiguelCompany:

PR #2245, opened 2021-10-04
Adds profile loading from XML string [12821]

SamuelWHalodi:
Implements #2179 

Signed-off-by: Samuel Wilhelmsson <samuel@halodi.com>

This is a remake of a previous PR had to be closed because of a bad merge.

Related documentation PR: eProsima/Fast-DDS-docs#303

PR #2345, opened 2021-12-10
[12821] Hotfix: add xml string feature to versions

JLBuenoLopez-eProsima:
Only last commit is relevant. Merge after #2245.


EduPonz (approved):
LGTM

PR #2356, opened 2021-12-16
Ensure correct boost singleton destruction order [13288]

MiguelCompany:
The first time an interprocess synchronization object is created by boost, a singleton is instantiated and its destructor is registered with std::atexit(&atexit_work).

This PR ensures that singleton is created before the DomainParticipantFactory instance so that it is destroyed after all the participants have been deleted.

### DIFF
--- a/src/cpp/fastdds/domain/DomainParticipantFactory.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantFactory.cpp
@@ -32,7 +32,7 @@
 #include <statistics/fastdds/domain/DomainParticipantImpl.hpp>
 #include <utils/SystemInfo.hpp>
 
- // We include boost through this internal header, to ensure we use our custom boost config file
+// We include boost through this internal header, to ensure we use our custom boost config file
 #include <utils/shared_memory/SharedMemSegment.hpp>
 #include <boost/interprocess/sync/interprocess_mutex.hpp>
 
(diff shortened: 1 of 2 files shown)


src/cpp/fastdds/domain/DomainParticipantFactory.cpp
```diff
@@ -101,6 +102,26 @@ DomainParticipantFactory::~DomainParticipantFactory()
...
 DomainParticipantFactory* DomainParticipantFactory::get_instance()
 {
+    /*
+     * The first time an interprocess synchronization object is created by boost, a singleton is instantiated and
+     * its destructor is registered with std::atexit(&atexit_work).
+     *
+     * We need to ensure that the boost singleton is destroyed after the instance of DomainParticipantFactory, to
+     * ensure that the interprocess objects keep working until all the participants are destroyed.
+     *
+     * We achieve this behavior by having an static instance of an auxiliary struct that instantiates a synchornization
```
JLBuenoLopez-eProsima, line 112:
```suggestion
     * We achieve this behavior by having an static instance of an auxiliary struct that instantiates a synchronization
```

JLBuenoLopez-eProsima (commented):
Only a typo

JLBuenoLopez-eProsima (approved):
:ok_hand:

JLBuenoLopez-eProsima (approved):
LGTM

MiguelCompany:

PR #2364, opened 2021-12-20
Fix locks used in MockConsumer [13303]

MiguelCompany:
The class used on unit tests for checking behavior regarding the log messages is wrongly used.
A reference to its condition_variable is obtained and then used with a different mutex.

This PR fixes this by adding a generic `wait` method receiving a predicate, as long as a `wait_for_at_least_entries` method, which is the most common use in the unit tests.

The `cv()` getter and `ConsumedEntriesSize_nts()` methods are removed, to ensure we don't have incorrect behaviour.

### DIFF
--- a/test/unittest/statistics/dds/StatisticsDomainParticipantMockTests.cpp
+++ b/test/unittest/statistics/dds/StatisticsDomainParticipantMockTests.cpp
@@ -42,11 +42,7 @@ public:
     void helper_block_for_at_least_entries(
             uint32_t amount)
     {
-        std::unique_lock<std::mutex> lck(*mutex_);
-        mock_consumer_->cv().wait(lck, [this, amount]
-                {
-                    return mock_consumer_->ConsumedEntriesSize_nts() >= amount;
-                });
+        mock_consumer_->wait_for_at_least_entries(amount);
     }
 
     eprosima::fastdds::dds::MockConsumer* mock_consumer_;
--- a/test/unittest/statistics/dds/StatisticsDomainParticipantMockTests.cpp
+++ b/test/unittest/statistics/dds/StatisticsDomainParticipantMockTests.cpp
@@ -47,21 +47,6 @@ public:
 
     eprosima::fastdds::dds::MockConsumer* mock_consumer_;
 
-    mutable std::mutex* mutex_;
-
-protected:
-
-    void SetUp() override
-    {
-        mutex_ = new std::mutex();
-    }
-
-    void TearDown() override
-    {
-        delete mutex_;
-        mutex_ = nullptr;
-    }
-
 };
 
 class DomainParticipantImplTest : public DomainParticipantImpl
--- a/test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
+++ b/test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
@@ -141,21 +141,6 @@ public:
 
     eprosima::fastdds::dds::MockConsumer* mock_consumer_;
 
-    mutable std::mutex* mutex_;
-
-protected:
-
-    void SetUp() override
-    {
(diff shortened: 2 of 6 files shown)


test/unittest/statistics/dds/StatisticsDomainParticipantMockTests.cpp
```diff
@@ -42,11 +42,7 @@ class StatisticsDomainParticipantMockTests : public ::testing::Test
     void helper_block_for_at_least_entries(
             uint32_t amount)
     {
-        std::unique_lock<std::mutex> lck(*mutex_);
```
richiware, line 45: I think this mutex is only used here. Remove it.

test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
```diff
@@ -136,11 +136,7 @@ class StatisticsDomainParticipantTests : public ::testing::Test
     void helper_block_for_at_least_entries(
             uint32_t amount)
     {
-        std::unique_lock<std::mutex> lck(*mutex_);
```
richiware, line 139: I think this mutex is only used here. Remove it.

test/unittest/xmlparser/XMLElementParserTests.cpp
```diff
@@ -61,11 +61,7 @@ class XMLParserTests : public ::testing::Test
     void helper_block_for_at_least_entries(
             uint32_t amount)
     {
-        std::unique_lock<std::mutex> lck(*xml_mutex_);
```
richiware, line 64: I think this mutex is only used here. Remove it.

test/unittest/xmlparser/XMLEndpointParserTests.cpp
```diff
@@ -59,11 +59,7 @@ class XMLEndpointParserTests : public ::testing::Test
     void helper_block_for_at_least_entries(
             uint32_t amount)
     {
-        std::unique_lock<std::mutex> lck(*xml_mutex_);
```
richiware, line 62: I think this mutex is only used here. Remove it.

test/unittest/xmlparser/XMLParserTests.hpp
```diff
@@ -68,11 +68,7 @@ class XMLParserTests : public ::testing::Test
     void helper_block_for_at_least_entries(
             uint32_t amount)
     {
-        std::unique_lock<std::mutex> lck(*xml_mutex_);
```
richiware, line 71: I think this mutex is only used here. Remove it.

richiware (changes requested):

test/unittest/statistics/dds/StatisticsDomainParticipantMockTests.cpp
MiguelCompany: done

test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
MiguelCompany: done

test/unittest/xmlparser/XMLElementParserTests.cpp
MiguelCompany: done

test/unittest/xmlparser/XMLEndpointParserTests.cpp
MiguelCompany: done

test/unittest/xmlparser/XMLParserTests.hpp
MiguelCompany: done

richiware (approved):

PR #2365, opened 2021-12-20
Avoid logWarning when environment file filename is empty [13302]

EduPonz:
This PR avoids a `logWarning` that was issued whenever `DomainParticipantFactory::load_profiles()` was called for the first time without setting `FASTDDS_ENVIRONMENT_FILE` environment variable or setting it to `""`. It also adds a regression test and a test that checks that the `logWarning` is indeed issued if `FASTDDS_ENVIRONMENT_FILE` is set to an non-existing file.

### DIFF
--- a/test/blackbox/common/BlackboxTestsPubSubBasic.cpp
+++ b/test/blackbox/common/BlackboxTestsPubSubBasic.cpp
@@ -99,7 +99,9 @@ public:
  * @param expected_logs The number of expected logs (error or warning) matching
  *                      ".*does not exist. File watching not initialized."
  */
-void env_file_warning(const char* env_file_name, size_t expected_logs)
+void env_file_warning(
+        const char* env_file_name,
+        size_t expected_logs)
 {
     /* Set environment variable */
 #ifdef _WIN32
--- a/test/blackbox/common/BlackboxTestsPubSubBasic.cpp
+++ b/test/blackbox/common/BlackboxTestsPubSubBasic.cpp
@@ -26,6 +26,7 @@
 
 #include <gtest/gtest.h>
 
+#include <cstdlib>
 #include <string>
 #include <tuple>
 #include <vector>
@@ -127,8 +128,7 @@ void env_file_warning(const char* env_file_name, size_t expected_logs)
     EXPECT_EQ(helper_consumer->ConsumedEntries().size(), expected_logs);
 
     /* Clean-up */
-    helper_consumer->ClearEntries(); // This calls to ClearConsumers, which deletes the registered consumer
-    Log::Reset();
+    Log::Reset();  // This calls to ClearConsumers, which deletes the registered consumer
 }
 
 TEST_P(PubSubBasic, PubSubAsNonReliableHelloworld)
@@ -809,7 +809,7 @@ TEST_P(PubSubBasic, ReliableTransientLocalTwoWritersConsecutives)
 /*
  * Check that setting FASTDDS_ENVIRONMENT_FILE to an unexisting file issues 1 logWarning
  */
-TEST_P(PubSubBasic, EnvFileWarningWrongFile)
+TEST(PubSubBasic, EnvFileWarningWrongFile)
 {
     env_file_warning("unexisting_file", 1);
 }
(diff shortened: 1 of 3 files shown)


test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -81,6 +86,51 @@ class PubSubBasic : public testing::TestWithParam<std::tuple<communication_type,
...
+ * Sets environment variable @c FASTDDS_ENVIRONMENT_FILE to @c env_file_name, and then initializes a DomainParticipant,
+ * using a custom log consumer to check how many logWarnings matching
+ * ".*does not exist. File watching not initialized." are issued, comparing that number with @c expected_logs.
+ * Finally, it resets the log module
+ *
+ * @param env_file_name The value to assign to @c FASTDDS_ENVIRONMENT_FILE
+ * @param expected_logs The number of expected logs (error or warning) matching
+ *                      ".*does not exist. File watching not initialized."
+ */
+void env_file_warning(const char* env_file_name, size_t expected_logs)
```
MiguelCompany, line 101: Linters complain with this.

```suggestion
void env_file_warning(
        const char* env_file_name,
        size_t expected_logs)
```

test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -81,6 +86,51 @@ class PubSubBasic : public testing::TestWithParam<std::tuple<communication_type,
...
+{
+    /* Set environment variable */
+#ifdef _WIN32
+    _putenv_s("FASTDDS_ENVIRONMENT_FILE", env_file_name);
+#else
+    setenv("FASTDDS_ENVIRONMENT_FILE", env_file_name, 1);
+#endif // _WIN32
```
MiguelCompany, line 108: Include the relevant headers for these functions.

test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -81,6 +86,51 @@ class PubSubBasic : public testing::TestWithParam<std::tuple<communication_type,
...
+    reader.init();
+    EXPECT_TRUE(reader.isInitialized());
+
+    /* Check logs */
+    Log::Flush();
+    EXPECT_EQ(helper_consumer->ConsumedEntries().size(), expected_logs);
+
+    /* Clean-up */
+    helper_consumer->ClearEntries(); // This calls to ClearConsumers, which deletes the registered consumer
+    Log::Reset();
```
MiguelCompany, line 131: If the consumer is going to be deleted by the `Log::Reset()` call, why should we call `ClearEntries`?
```suggestion
    Log::Reset(); // This calls to ClearConsumers, which deletes the registered consumer
```

test/blackbox/common/BlackboxTestsPubSubBasic.cpp
```diff
@@ -756,6 +806,22 @@ TEST_P(PubSubBasic, ReliableTransientLocalTwoWritersConsecutives)
...
+ */
+TEST_P(PubSubBasic, EnvFileWarningWrongFile)
+{
+    env_file_warning("unexisting_file", 1);
+}
+
+/*
+ * Check that setting FASTDDS_ENVIRONMENT_FILE to an empty string issues 0 logWarning
+ */
+TEST_P(PubSubBasic, EnvFileWarningEmpty)
```
MiguelCompany, line 820: Do we really need these tests to be `TEST_P`? I think we could use just `TEST`

test/blackbox/common/mock/BlackboxMockConsumer.h
```diff
@@ -42,6 +42,12 @@ class BlackboxMockConsumer : public LogConsumer
         return mEntriesConsumed;
     }
 
+    void ClearEntries()
```
MiguelCompany, line 45: As commented before, I don't know if this new method is necessary. I don't mind having it for the future, but then it may be nice to follow naming style rules: i.e. `clear_entries`

test/blackbox/common/mock/BlackboxMockConsumer.h
```diff
@@ -42,6 +42,12 @@ class BlackboxMockConsumer : public LogConsumer
...
+    {
+        std::unique_lock<std::mutex> guard(mMutex);
+        mEntriesConsumed.clear();
```
MiguelCompany, line 48: We are modifying the entries, so we should notify the condition.

```suggestion
        mEntriesConsumed.clear();
        cv_.notify_all();
```

MiguelCompany (changes requested):

MiguelCompany (approved):
LGTM

PR #2368, opened 2021-12-21
Avoid null dereference on log message

MiguelCompany:
Method `ReadTakeCommand::check_datasharing_validity` could receive 'nullptr' on `wp`.

As that parameter was only used on the `logWarning` call to obtain the GUID of the writer, and it could be obtained from the cache change, I just removed that parameter completely


jsantiago-eProsima (approved):
LGTM

MiguelCompany:

PR #2371, opened 2021-12-22
Fix #2370

### LINKED ISSUE #2370: CMake:  Should VERSION_SCRIPT_SUPPORT_FLAGS always be PRIVATE?
https://github.com/eProsima/Fast-DDS/commit/2a655a22af6dd757ced7220e17f9db01e3b1eddd#diff-e3052ec6d01ba5f9ae8d014fd3bdd8653f7e759e704a77bbfdb99ada532923daR517

Is there a reason `${VERSION_SCRIPT_SUPPORT_FLAGS}` should have PUBLIC visibility when building a shared library?  When I build my library which links to fastrtps and I set in my project `target_link_libraries( ${PROJECT_NAME} PUBLIC fastrtps )`, the `-Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libfastrtps.version` linker command is pulled into my build, and destroys my library's symbol table, leading to `undefined reference` errors in my library (despite being linked properly).  I lost a day tracking this down.

When I change the linked source code line to `PRIVATE`, the linker flags are not included in my build, and linking is successful.  I'm not sure it makes sense to have this linker command be exported as `PUBLIC`, unless I'm missing something.  Thanks

MiguelCompany:
This should fix #2370 by changing the privacy of the linker option to always private

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -514,7 +514,7 @@ elseif(BUILD_SHARED_LIBS)
     unset(OLD_CMAKE_REQUIRED_FLAGS)
     if(HAS_VERSION_SCRIPT_SUPPORT)
         message(STATUS "Using linker version script to select exported symbols")
-        target_link_options(${PROJECT_NAME} ${PRIVACY} ${VERSION_SCRIPT_SUPPORT_FLAGS})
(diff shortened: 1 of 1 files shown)


EduPonz (approved):
I've checked that this indeed solves it

PR #2380, opened 2021-12-24
[13340] Bugfix/windows installer/static linking

MiguelBarro:
CMake always makes static libraries transitive targets even if they are marked as PRIVATE, that is, CMake will never bundle Fast-DDS static library with OpenSSL, tinyxml or pkcs11.

People have found ways to work around it through custom CMake commands [see here](https://cristianadam.eu/20190501/bundling-together-static-libraries-with-cmake/).

We took an easier workaround on the third parties by including the actual sources into the Fast-DDS project.

The Bugfix plainly removes the Fast-DDS dependency because it's only useful on the tests. Note the OpenSSL engine loads the pkcs11 module dynamically thus no linking is required.


jsantiago-eProsima (approved):
LGTM

MiguelCompany:
Closing in favor of #2381, as we have a mirror action from `master` into `2.5.x`, so merges into master are automatically backported.

PR #2381, opened 2021-12-24
Move libp11 dependency to tests only

MiguelBarro:
CMake always makes static libraries transitive targets even if they are marked as PRIVATE, that is, CMake will never bundle Fast-DDS static library with OpenSSL, tinyxml or pkcs11.

People have found ways to work around it through custom CMake commands [see here](https://cristianadam.eu/20190501/bundling-together-static-libraries-with-cmake/).

We took an easier workaround on the third parties by including the actual sources into the Fast-DDS project.

The Bugfix plainly removes the Fast-DDS dependency because it's only useful on the tests. Note the OpenSSL engine loads the pkcs11 module dynamically thus no linking is required.


jsantiago-eProsima (dismissed):
LGTM

MiguelCompany (changes requested):
It seems `target_link_libraries` also adds the include directory of the linked target, so removing it is making the build fail with `Cannot open include file: 'libp11.h': No such file or directory (compiling source file C:\workspace\fastdds-manual-windows\b64998d6\src\fastrtps\src\cpp\rtps\security\SecurityPluginFactory.cpp)`

I wonder why it is building correctly on other platforms, though.

MiguelBarro:
The problem is in the jenkins windows image. I will relaunch the windows testing once it is fixed.

MiguelBarro:
@richiprosima Please darling be so gentle to test Windows for me 😎

src/cpp/security/authentication/PKIDH.cpp
```diff
@@ -258,27 +258,32 @@ static EVP_PKEY* load_private_key(
...
-        exception = _SecurityException_(std::string("PKCS11 URIs require libp11 ") + file);
-        return nullptr;
-#endif // HAVE_LIBP11
+
+        key = pkidh.pkcs11_provider->load_private_key(certificate, file, password, exception);
+
+        if ( nullptr == key )
+        {
+            exception = _SecurityException_(std::string("PKCS11 URIs require libp11 ") + file);
+        }
```
MiguelCompany, line 279: Method `pkcs11_provider->load_private_key` is already filling out `exception` in case of error, so there is no need to change the exception message here.

```suggestion
```

test/blackbox/CMakeLists.txt
```diff
@@ -205,9 +205,14 @@ endif()
 
 # OpenSSL on Windows requires a hint on which config file to load
 if(WIN32 AND OPENSSL_FOUND)
-    get_filename_component(OPENSSL_DIR "${OPENSSL_INCLUDE_DIR}" DIRECTORY)
-    set(OPENSSL_CONF "${OPENSSL_DIR}/bin/cnf/openssl.cnf")
-    unset(OPENSSL_DIR)
+
+    find_package(LibP11)
+
+    if(LibP11_FOUND)
```
MiguelCompany, line 211: Why is this condition necessary?
Do we need to set OPENSSL_CONF if LibP11 is not found?

MiguelCompany (changes requested):

src/cpp/security/authentication/PKIDH.cpp
MiguelBarro: You're right. Upstream more complete information is available.

test/blackbox/CMakeLists.txt
MiguelBarro: PKCS11 testing requires such a complex configuration and dependency settings (on windows at least) that will (probably) only suit our CI.
I'm going to subordinate the windows configuration file settings to the presence of the `EPROSIMA_BUILD`  flag to prevent an override of user-specific Open-SSL settings.
Although, is quite probable that the PKCS11 testing will fail on a user machine unless [this settings](https://fast-dds.docs.eprosima.com/en/v2.5.0/installation/sources/sources_windows.html?highlight=softhsm#libp11-and-softhsm-libraries) are enforced.

MiguelCompany (changes requested):
Changes seem ok now, but the build failed on Mac, as we don't have LibP11 installed on the CI.

MiguelCompany:
@richiprosima Please test Mac to check if I installed libp11 correctly

MiguelBarro:
@richiprosima Please test Mac again because we have modified CI's setup since the last push 🛠

MiguelCompany:
@richiprosima Please test Mac and test Windows

MiguelCompany:
@richiprosima Please test Mac

MiguelCompany (approved):
I think it is safe to merge this. CI has been correct for all the related tests.

PR #2389, opened 2022-01-07
Fix fill local IP [13390]

rsanchez15:
This PR fix the fill local IP functions defining the IP kind before filling the locator with the given IP.

### DIFF
--- a/src/cpp/rtps/transport/TCPv4Transport.cpp
+++ b/src/cpp/rtps/transport/TCPv4Transport.cpp
@@ -328,7 +328,7 @@ bool TCPv4Transport::compare_locator_ip_and_port(
 void TCPv4Transport::fill_local_ip(
         Locator& loc) const
 {
-    loc.kind = LOCATOR_KIND_TCPv4;
+    loc.kind = kind();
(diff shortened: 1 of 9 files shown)


src/cpp/rtps/transport/TCPv4Transport.cpp
```diff
@@ -328,8 +328,8 @@ bool TCPv4Transport::compare_locator_ip_and_port(
 void TCPv4Transport::fill_local_ip(
         Locator& loc) const
 {
-    IPLocator::setIPv4(loc, "127.0.0.1");
     loc.kind = LOCATOR_KIND_TCPv4;
```
EduPonz, line 331: In general,

I suggest using the `kind()` getter inhereted from `TransportInterface`

EduPonz (changes requested):
I'd really appreciate a TCPv6 communication blackbox test to check that it works at least locally

EduPonz (approved):
LGTM with green CI

EduPonz:
@richiprosima please test Linux

PR #2396, opened 2022-01-10
Change log verbosity in error mutation_took_place_on_meta [13422]

jparisu:


src/cpp/rtps/RTPSDomain.cpp
```diff
@@ -192,7 +192,7 @@ RTPSParticipant* RTPSDomain::createParticipant(
                 PParam.builtin.metatrafficUnicastLocatorList))
     {
         // we do not log an error because the library may use participant creation as a trial for server existence
```
MiguelCompany, line 194: The comment would become invalid, so it should be removed

MiguelCompany (changes requested):

src/cpp/rtps/RTPSDomain.cpp
jparisu: :+1:

EduPonz (approved):
LGTM

PR #2403, opened 2022-01-12
Fix not caught exception in XMLDynamicParser [13462]

richiware:
XMLDynamicParser uses `stoi()` without a try/catch statement. With a wrong XML attribute value a exception is launched without being caught.

This will fix oss-fuzz 43525


MiguelCompany (approved):
LGTM

MiguelCompany:
@mergifyio backport 2.3.x

PR #2405, opened 2022-01-12
Enable memory protection on DataSharing readers [13457]

### LINKED ISSUE #2385: Shared Memory data protection [13457]
Hi,

I got some security questions ( @MiguelCompany ) .

Regarding data-sharing/ zero copy transport.
The reader get the pointer to the actual shared memory region same as the writer, so:

According to the documentation, as I understands it, a reader can modify the sample :
>Important: If the samples “returned” by this method are loaned from the middleware (see take for more information on memory loaning), it is important that their contents not be changed. Because the memory in which the data is stored belongs to the middleware, any modifications made to the data will be seen the next time the same samples are read or taken; the samples will no longer reflect the state that was received from the network.

I also tried to run a subscriber only application and using `virt2phys` I see this:

```
7f8b91478000-7f8b9147e000 rw-s 00000000 00:1a 231                        /dev/shm/fast_datasharing_01.0f.86.eb.8c.8b.23.0c.01.00.00.00_0.0.1.4
7f8b882b4000-7f8b887fb000 rw-s 00000000 00:1a 224                        /dev/shm/fast_datasharing_01.0f.86.eb.96.86.1d.b7.01.00.00.00_0.0.1.3
```

The application get "rw" access to the topic.
I understand the MW need access to the "metadate/header" of the data in order to control it but the user data is vulnerable and can be mistakably override by a read only application.

according to what I read on boost/interprocess the shared memory segments can be opened as read only access mode.

Does `FastDDS` API supply some kind of protection reader overriding samples?

a misbehaved subscriber has direct access to the sample even if only a reader is created, he can update the data even without FastDDS API because the virtual address that mapped to the shared memory belongs to him and currently I see it as "rw" access.
theoretically a data overrun inside the reader only process can change the topic data for other readers.
Is that correct?
Can we possibly block this behavior?

I also tried running the zeroCopy example with 1 publisher and 1 subscriber and saw that 2 shared memory data files were created.
Can you explain why? I was expecting only 1 since both the reader and the write work with the same shared memory segment.
image

Weird questions, I know...
But that's the world we live in :) 

Thanks
Nir

MiguelCompany:
This PR adds several safety protection mechanisms on DataReader entities using data-sharing.

First, the readers will open the shared history on read-only mode.
Second, a new FASTDDS_CONST_SEQUENCE macro to define a sequence that only has the `const` version of the `[]` operator.

Should fix #2385

### DIFF
--- a/examples/C++/DDS/ZeroCopyExample/LoanableHelloWorldSubscriber.cxx
+++ b/examples/C++/DDS/ZeroCopyExample/LoanableHelloWorldSubscriber.cxx
@@ -126,7 +126,7 @@ void LoanableHelloWorldSubscriber::SubListener::on_subscription_matched(
 void LoanableHelloWorldSubscriber::SubListener::on_data_available(
         DataReader* reader)
 {
-    FASTDDS_SEQUENCE(DataSeq, LoanableHelloWorld);
+    FASTDDS_CONST_SEQUENCE(DataSeq, LoanableHelloWorld);
 
     DataSeq data;
     SampleInfoSeq infos;
--- a/include/fastdds/dds/core/LoanableSequence.hpp
+++ b/include/fastdds/dds/core/LoanableSequence.hpp
@@ -22,6 +22,7 @@
 #include <cassert>
 #include <cstdint>
 #include <vector>
+#include <type_traits>
 
 #include <fastdds/dds/core/LoanableTypedCollection.hpp>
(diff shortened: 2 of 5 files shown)


JLBuenoLopez-eProsima (approved):
LGTM!

CI failures seem unrelated.

MiguelCompany:

PR #2407, opened 2022-01-17
Remove UDP transport from the shared memory example

Nir-Az:
The shared memory example is adding 2 transport layers, shared memory + UDP.
The UDP will not be used if a shared memory transport layer conditions are met and this is a unique shared memory example.
I find it confusing as a user that the shared memory example also configure UDP, it looks like a mandatory step for shared memory and it's not.

Consider using this PR to remove it and make the example clearer.


EduPonz (approved):
LGTM

EduPonz:
Thanks @Nir-Az

PR #2416, opened 2022-01-20
Fixing wrong usage of condition_variable [13542]

richiware:
In a blackbock tests a condition_variable is being notified locking a mutex but for waiting process another mutex is used.


MiguelCompany (approved):
LGTM

MiguelCompany:
Unrelated failures

PR #2431, opened 2022-01-25
Fix flaky test: incompatible qos conditions [13582]

richiware:
Sometimes the `DDSStatus.IncompatibleQosConditions` test fails due to `PubSubReader::incompatible_qos()` not taking into account `total_count_changes`, and instead only increasing the variable `times_incompatible_qos_` by one each time the Waitset returns. This PR corrects `PubSubReader::incompatible_qos()` so it uses `times_incompatible_qos_`.

### DIFF
--- a/test/blackbox/api/dds-pim/PubSubReader.hpp
+++ b/test/blackbox/api/dds-pim/PubSubReader.hpp
@@ -721,7 +721,7 @@ public:
     }
 
     void incompatible_qos(
-            eprosima::fastdds::dds::OfferedIncompatibleQosStatus status)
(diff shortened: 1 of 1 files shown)


MiguelCompany (changes requested):
I think [this failing test](http://jenkins.eprosima.com:8080/job/fastdds_windows/label=windows-secure,toolset=v142/1814/testReport/junit/projectroot.test/blackbox/BlackboxTests_DDS_PIM_DDSStatus_IncompatibleQosGetters_Intraprocess/) is related with the changes here.

richiware:
I was checking the failed test and I think it is not related with my changes.

PR #2448, opened 2022-01-31
Fix inserting minimum CacheChange_t in GAP message [13671]

richiware:
This PR fixes a bug which occurs when the `CacheChange_t` with minimum sequence number in the Writer's history is irrelevant for a new reader (for example because reader is volatile). This change is not added in the GAP message because of a `>` instead of a `>=`.


MiguelCompany (changes requested):
This needs a regression test

MiguelCompany (approved):
LGTM.

richiware:
@mergifyio backport 2.4.x

PR #2468, opened 2022-02-07
[13780] Fix dynamic network interfaces

JLBuenoLopez-eProsima:
This PR fixes the dynamic network interfaces feature.


JLBuenoLopez-eProsima:
CI failures seem unrelated

src/cpp/rtps/transport/UDPTransportInterface.cpp
```diff
@@ -61,6 +61,8 @@ UDPTransportInterface::UDPTransportInterface(
     : TransportInterface(transport_kind)
     , mSendBufferSize(0)
     , mReceiveBufferSize(0)
+    , first_time_open_output_channel_(true)
+    , whitelisted_(false)
```
MiguelCompany, line 65: After adding it to UDPSenderResource, it seems this whitelisted_ flag is not used.

```suggestion
    , first_time_open_output_channel_(true)
```

src/cpp/rtps/transport/UDPTransportInterface.cpp
```diff
@@ -698,6 +674,33 @@ bool UDPTransportInterface::fillUnicastLocator(
     return true;
 }
 
+void UDPTransportInterface::remove_known_network_interfaces(
+        const SendResourceList& sender_resource_list,
+        std::vector<IPFinder::info_IP>& locNames,
+        bool return_loopback)
+{
+    get_ips(locNames, return_loopback);
```
MiguelCompany, line 682: I would rename this method to something like `get_unknown_network_interfaces`, and clear the locNames inside it, instead of forcing the caller to have it empty before calling the method.

```suggestion
void UDPTransportInterface::get_unknown_network_interfaces(
        const SendResourceList& sender_resource_list,
        std::vector<IPFinder::info_IP>& locNames,
        bool return_loopback)
{
    locNames.clear();
    get_ips(locNames, return_loopback);
```

src/cpp/rtps/transport/UDPTransportInterface.h
```diff
@@ -175,6 +176,11 @@ class UDPTransportInterface : public TransportInterface
     uint32_t mReceiveBufferSize;
     eprosima::fastdds::statistics::rtps::OutputTrafficManager statistics_info_;
 
+    //! First time open output channel flag: open unicast and localhost socket when not whitelisted
+    bool first_time_open_output_channel_;
+    //! Whitelist feature is being used
+    bool whitelisted_;
```
MiguelCompany, line 182: I think this can be removed

```suggestion
```

src/cpp/rtps/transport/UDPTransportInterface.h
```diff
@@ -259,7 +265,19 @@ class UDPTransportInterface : public TransportInterface
...
+            bool whitelisted,
             const std::chrono::microseconds& timeout);
+
+    /**
+     * @brief Remove already opened network interfaces
+     *
+     * @param sender_resource_list List of SenderResources already registered in the transport
+     * @param locNames List of available network interfaces
+     */
+    void remove_known_network_interfaces(
```
MiguelCompany, line 277: Update doxydoc (including missing param `return_loopback`), the purpose is `Return list of not yet open network interfaces`

test/unittest/transport/UDPv4Tests.cpp
```diff
@@ -148,11 +148,19 @@ TEST_F(UDPv4Tests, send_and_receive_between_ports)
                 LocatorList_t locator_list;
                 locator_list.push_back(multicastLocator);
 
-                Locators locators_begin(locator_list.begin());
-                Locators locators_end(locator_list.end());
-
-                EXPECT_TRUE(send_resource_list.at(0)->send(message, 5, &locators_begin, &locators_end,
-                        (std::chrono::steady_clock::now() + std::chrono::microseconds(100))));
+                bool sended = false;
```
MiguelCompany, line 151:
```suggestion
                bool sent = false;
```

MiguelCompany: Here and below

src/cpp/rtps/transport/UDPv4Transport.cpp
```diff
@@ -553,6 +554,49 @@ void UDPv4Transport::SetSocketOutboundInterface(
...
+            {
+                std::vector<IPFinder::info_IP> locNames;
+                get_ipv4s_unique_interfaces(locNames, true);
+                for (const auto& infoIP : locNames)
+                {
+                    auto ip = asio::ip::address_v4::from_string(infoIP.name);
+                    try
+                    {
+                        channelResource->socket()->set_option(ip::multicast::join_group(
+                                    ip::address_v4::from_string(DEFAULT_METATRAFFIC_MULTICAST_ADDRESS), ip));
```
MiguelCompany, line 573: What will happen when a multicast input channel different from the default one has been open, and `update_network_interfaces` is called?

MiguelCompany (changes requested):

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -1252,6 +1253,26 @@ void RTPSParticipantImpl::update_attributes(
...
+            m_network_Factory.RegisterTransport(nullptr);
+            createSenderResources(m_att.builtin.metatrafficMulticastLocatorList);
+            createSenderResources(m_att.builtin.metatrafficUnicastLocatorList);
+
+            // Update default locators
+            for (auto locator : m_att.defaultUnicastLocatorList)
+            {
+                local_participant_proxy_data->default_locators.add_unicast_locator(locator);
+            }
+
```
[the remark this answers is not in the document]
JLBuenoLopez-eProsima: Yes, we should create the sender resources. `UDPTransportInterface::OpenOutputChannel` takes charge of registering the locators into the Fast DDS Statistics Module, so these new locators would not be added into the statistics. But other than this reason, the `OpenOutputChannel` logic does not depend on the locator so calling it once would be enough. Maybe a refactor could be considered for `2.6.0`.

src/cpp/rtps/transport/UDPTransportInterface.cpp
JLBuenoLopez-eProsima: Good catch. It was added because it was required for a previous solution. Now that the `AnyAddressEndpoint` is out bounded to the localhost interface, is no longer necessary.

src/cpp/rtps/transport/UDPv4Transport.cpp
JLBuenoLopez-eProsima: I do not think we should be worried by that specific case. This feature has been designed to work only when the locators are generated automatically by Fast DDS. Being that the case, if the user has opened a different multicast input channel (or several ones), the feature would not work because `RTPSParticipantImpl::internal_metatraffic_locators_` flag would not be set. I have move the call to this function in `RTPSParticipantImpl::update_attributes` so is called only when the default unicast metatraffic locators have changed.

MiguelCompany (approved):
LGTM

JLBuenoLopez-eProsima:
Windows warnings are solved in #2487. Other CI failures are unrelated

PR #2469, opened 2022-02-07
Fix infinite loop in TCP header read [13721]

juanlofer-eprosima:
Signed-off-by: Juan López Fernández <juanlopez@eprosima.com>

### DIFF
--- a/src/cpp/rtps/transport/TCPTransportInterface.cpp
+++ b/src/cpp/rtps/transport/TCPTransportInterface.cpp
@@ -975,14 +975,14 @@ bool TCPTransportInterface::Receive(
         {
             if (ec != asio::error::eof)
             {
-                logWarning(DEBUG, "Error reading TCP header: " << ec.message());
+                logWarning(DEBUG, "Failed to read TCP header: " << ec.message());
             }
             close_tcp_socket(channel);
             success = false;
         }
         else if (!channel->connection_status())
         {
-            logWarning(DEBUG, "Error reading TCP header: channel disconnected while reading.");
+            logWarning(DEBUG, "Failed to read TCP header: channel disconnected while reading.");
             success = false;
         }
         else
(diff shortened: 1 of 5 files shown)


src/cpp/rtps/transport/TCPTransportInterface.cpp
```diff
@@ -972,6 +980,11 @@ bool TCPTransportInterface::Receive(
             close_tcp_socket(channel);
             success = false;
         }
+        else if (!channel->connection_status())
+        {
+            logWarning(DEBUG, "Error reading TCP header: channel disconnected while reading.");
```
juanlofer-eprosima, line 985: Should the log category be different?

JLBuenoLopez-eProsima: Currently there are no too many log levels supported. If you don't want to promote this log to the Error level (even though the message says "Error") then a Warning seems to be correct. If the Warning level is kept, I suggest changing the log message:

```suggestion
            logWarning(DEBUG, "Failed to read TCP header: channel disconnected while reading.");
```

JLBuenoLopez-eProsima (changes requested):
I think these changes require a regression test before being considered for merging.

src/cpp/rtps/transport/TCPTransportInterface.cpp
juanlofer-eprosima: Yes I believe it should be a warning, though I was referring to the `DEBUG` part.

JLBuenoLopez-eProsima: Other Errors/Warnings in this module are being registered under the `DEBUG` label. Thus I will keep consistency. Either change all other logs to another new label (maybe `TCP_DEBUG`) or keep this label as it is currently.

juanlofer-eprosima:
Changes moved to #2470

PR #2473, opened 2022-02-08
Fix clock used by TimedConditionVariable

richiware:
When compiled with `STRICT_REALTIME`, a custom condition variable is used: TimedConditionVariable. This uses internally a `pthread_cond_t`  which by default uses `CLOCK_REALTIME`. But FastDDS uses `steady_clock` (`CLOCK_MONOTONIC`) and therefore `wait_until()`function return immediately. This PR fixes this configuring `pthread_cond_t` to use `CLOCK_MONOTONIC`.


MiguelCompany:

MiguelCompany (approved):

MiguelCompany:
Unrelated failures. New test checked locally

PR #2476, opened 2022-02-09
[13824] add check_xml_static_discovery(std::string& xml)

inesluno:
Add a member function `ReturnCode_t DomainParticipantFactory::check_xml_static_discovery(std::string& xml_file)` that can be used to check Static Discovery XML by calling `XMLEndpointParser::loadXMLFile()`

Signed-off-by: Ines Luño <inesluno@eprosima.com>


src/cpp/fastdds/domain/DomainParticipantFactory.cpp
```diff
@@ -386,6 +387,18 @@ ReturnCode_t DomainParticipantFactory::load_XML_profiles_string(
     return ReturnCode_t::RETCODE_OK;
 }
 
+ReturnCode_t DomainParticipantFactory::check_xml_static_discovery(
+    std::string& xml_file)
+{
+    XMLEndpointParser parser;
+    if (XMLP_ret::XML_ERROR == parser.loadXMLFile(xml_file))
```
EduPonz, line 394: Right now, the parser only returns either error or ok, but that may change in the future. In that case, I'd only return ok whenever the parser does so.
```suggestion
    if (XMLP_ret::XML_OK != parser.loadXMLFile(xml_file))
```

EduPonz (changes requested):
1. Please address warnings
2. Uncrustify failed
3. I'd separate the Doxygen related changes into a separate PR that we can merge for v2.5.1
4. I'm missing tests for the new API

EduPonz (changes requested):
Please, document this new API in [here](https://fast-dds.docs.eprosima.com/en/latest/fastdds/discovery/static.html#static-discovery-settings)

src/cpp/fastdds/domain/DomainParticipantFactory.cpp
inesluno: ok, committed

EduPonz:
Closed in favor of #2817

PR #2485, opened 2022-02-10
[13364] Fix memory problem when ciphering payload

### LINKED ISSUE #2379: Not enough memory to cipher payload [13364]
There is an error when try to send a big message (near 64 Kb boundary). It seems that encryption overhead is not considered when packing submessages inside `RTPSMessageGroup`. Such error occurs when the encryption plugin tries to encode the message, but it is too late. All of this can lead to message retransmission or even inability to send the message at all.

## Expected Behavior
It should be possible send messages of any size. If the submessage does not fit into the buffer then it should be split into smaller chunks.

## Current Behavior
There are the following error logs:
```
2021-12-23 14:06:46.067 [SECURITY_CRYPTO Error] Not enough memory to cipher payload -> Function serialize_SecureDataBody
2021-12-23 14:06:46.067 [RTPS_WRITER Error] Error encoding rtps message. -> Function send

```
## Steps to Reproduce
1. Compile the test program (attached).
2. Copy certs directory where you want to run the program.
3. Run the reader instance: ./big_message_test
4. Run the writer instance: ./big_message_test writer

## System information
- Fast-RTPS version: 2.4.0
- OS: Ubuntu 20.04

## Additional context
There is some attempt to limit a message size in `DataMsg::addSubmessageData()`:
```
    if (size32 <= std::numeric_limits<uint16_t>::max())
    {
        submessage_size = static_cast<uint16_t>(size32);
        octet* o = reinterpret_cast<octet*>(&submessage_size);
        if (msg->msg_endian == DEFAULT_ENDIAN)
        {
            msg->buffer[submessage_size_pos] = *(o);
            msg->buffer[submessage_size_pos + 1] = *(o + 1);
        }
        else
        {
            msg->buffer[submessage_size_pos] = *(o + 1);
            msg->buffer[submessage_size_pos + 1] = *(o);
        }

        *is_big_submessage = false;
    }

```
But this code does not consider an encryption overhead and the fact that buffer size can less than a hardcoded value. Right now the buffer size is less than 65535, because of memory alignment.
Also in the function `RTPSMessageGroup::insert_submessage()`:
```
    if (!append_message(full_msg_, submessage_msg_))
    {
        // Retry
        flush_and_reset();
        add_info_dst_in_buffer(full_msg_, destination_guid_prefix);

        if (!append_message(full_msg_, submessage_msg_))
        {
            logError(RTPS_WRITER, "Cannot add RTPS submesage to the CDRMessage. Buffer too small");
            return false;
        }
    }

```
[big_message_test.zip](https://github.com/eProsima/Fast-DDS/files/7769261/big_message_test.zip)

JLBuenoLopez-eProsima:
This closes #2379

### DIFF
--- a/src/cpp/rtps/messages/RTPSMessageGroup.cpp
+++ b/src/cpp/rtps/messages/RTPSMessageGroup.cpp
@@ -74,18 +74,29 @@ static bool data_exceeds_limitation(
 }
 
 static bool append_message(
+        RTPSParticipantImpl* participant,
         CDRMessage_t* full_msg,
         CDRMessage_t* submsg)
 {
-#ifndef FASTDDS_STATISTICS
-    return CDRMessage::appendMsg(full_msg, submsg);
-#else
+    static_cast<void>(participant);
+
+    uint32_t extra_size = 0;
+
+#if HAVE_SECURITY
+    // Avoid full message growing over estimated extra size for RTPS encryption
+    extra_size += participant->calculate_extra_size_for_rtps_message();
+#endif  // HAVE_SECURITY
+
+#ifdef FASTDDS_STATISTICS
     // Keep room for the statistics submessage by reducing max_size while appending submessage
-    full_msg->max_size -= eprosima::fastdds::statistics::rtps::statistics_submessage_length;
+    extra_size += eprosima::fastdds::statistics::rtps::statistics_submessage_length;
+#endif  // FASTDDS_STATISTICS
+
+    full_msg->max_size -= extra_size;
     bool ret_val = CDRMessage::appendMsg(full_msg, submsg);
-    full_msg->max_size += eprosima::fastdds::statistics::rtps::statistics_submessage_length;
+    full_msg->max_size += extra_size;
+
     return ret_val;
-#endif // FASTDDS_STATISTICS
 }
 
 bool sort_changes_group (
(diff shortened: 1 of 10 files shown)


JLBuenoLopez-eProsima:
This PR breaks several security tests:

* Communication tests in several platforms
* Blackbox tests in Windows platform

richiware (changes requested):
After a meeting we decided this approach is not the better one. The calculation has to be done when the `CacheChange_t` is added in the `WriterHistory`, as it is done now. According to @MiguelCompany  we have to take into account the size of the DataFrag header when calculating the size of each fragment.

MiguelCompany:
@richiware This has been rebased, and I have changed the last commit (the one that fixes the issue), with a nicer solution.

Would you mind reviewing this?

EduPonz:
@mergifyio rebase

EduPonz:

MiguelCompany:
@mergifyio rebase

MiguelCompany:
@richiprosima please test mac

MiguelCompany:

EduPonz (approved):

EduPonz:
@richiprosima please test windows test mac

EduPonz:

EduPonz:

EduPonz:
@MiguelCompany do we need to backport this?

MiguelCompany:
@mergifyio backport 2.12.x 2.11.x 2.10.x 2.6.x

PR #2502, opened 2022-02-15
Suppress OpenSSL 3.0 warnings.

clalancette:
This is a temporary fix until the code can be refactored to
use real OpenSSL 3.0 APIs.

Note that this should also be forward-ported to the master branch, as the same warning is there.

### DIFF
--- a/src/cpp/security/authentication/PKIDH.cpp
+++ b/src/cpp/security/authentication/PKIDH.cpp
@@ -16,6 +16,10 @@
  * @file PKIDH.cpp
  */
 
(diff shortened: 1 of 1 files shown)


MiguelCompany (approved):
LGTM.

The forward port would not be direct, as there is another file with OpenSSL 3.0 warnings, but it will be as easy as this one (we will take care of it).

Thank you @clalancette !

MiguelCompany:
Failures are unrelated.

MiguelCompany:
@mergifyio backport master 2.4.x 2.1.x

PR #2506, opened 2022-02-16
[10843] Correctly protect endpoints collections

### LINKED ISSUE #1819: Segmentation fault in RTPSDomainImpl::find_local_reader [10842]

JLBuenoLopez-eProsima:
This PR closes #1819. It also replaces #1823.

I have tried the regression test proposed in that PR but it does not seem to provoke the data race condition and the segmentation fault. However, the test is prone to failure due to discovery issues. Thus, I have not introduced it in this PR (if the reviewer considers it is of interest I can update this PR). There is no regression test because forcing the failure is not straight forward nor easy. However, the benefits of correctly protecting the access to the collections are too great compare to a regression test. My opinion is that this fix should be merged.

### DIFF
--- a/src/cpp/rtps/writer/StatefulWriter.cpp
+++ b/src/cpp/rtps/writer/StatefulWriter.cpp
@@ -493,7 +493,7 @@ bool StatefulWriter::intraprocess_heartbeat(
     bool returned_value = false;
 
     std::lock_guard<RecursiveTimedMutex> guardW(mp_mutex);
-    RTPSReader* reader = RTPSDomainImpl::find_local_reader(reader_proxy->guid());
+    RTPSReader* reader = reader_proxy->local_reader();
 
     if (reader)
     {
--- a/src/cpp/rtps/writer/StatefulWriter.cpp
+++ b/src/cpp/rtps/writer/StatefulWriter.cpp
@@ -493,7 +493,7 @@ bool StatefulWriter::intraprocess_heartbeat(
     bool returned_value = false;
 
     std::lock_guard<RecursiveTimedMutex> guardW(mp_mutex);
-    RTPSReader* reader = reader_proxy->local_reader();
+    RTPSReader* reader = RTPSDomainImpl::find_local_reader(reader_proxy->guid());
 
     if (reader)
     {
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
@@ -460,6 +460,7 @@ void RTPSParticipantImpl::disable()
         block.disable();
     }
 
+    std::unique_lock<std::recursive_mutex> lock(*mp_mutex);
     while (m_userReaderList.size() > 0)
     {
         deleteUserEndpoint(static_cast<Endpoint*>(*m_userReaderList.begin()));
@@ -469,6 +470,7 @@ void RTPSParticipantImpl::disable()
     {
         deleteUserEndpoint(static_cast<Endpoint*>(*m_userWriterList.begin()));
     }
+    lock.unlock();
 
     delete(mp_builtinProtocols);
     mp_builtinProtocols = nullptr;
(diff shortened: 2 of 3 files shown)


JLBuenoLopez-eProsima:
CI failures seem unrelated

src/cpp/rtps/writer/StatefulWriter.cpp
```diff
@@ -493,7 +493,7 @@ bool StatefulWriter::intraprocess_heartbeat(
     bool returned_value = false;
 
     std::lock_guard<RecursiveTimedMutex> guardW(mp_mutex);
-    RTPSReader* reader = RTPSDomainImpl::find_local_reader(reader_proxy->guid());
+    RTPSReader* reader = reader_proxy->local_reader();
```
MiguelCompany, line 496: I think this should be left untouched.

There is a corner case where the proxy could still have a pointer to an already deleted reader.

MiguelCompany (changes requested):

MiguelCompany (approved):

MiguelCompany:
@richiprosima Pleas test aarch64

JLBuenoLopez-eProsima:
@richiprosima please test aarch64

PR #2509, opened 2022-02-17
Encapsulate FastDDS library compile options

Nir-Az:
As a user of `FastDDS` library I do not wish that consuming `FastDDS` will affect my target compile warnings..
This should be set as private at the library level.

If dependent targets wish to act the same as `FastDDS` they should set their compile flags the same way..

My motivation is that after integrating `FastDDS` I started getting lots of unrequested warnings from my project.
Please consider changing it.

Thanks


MiguelCompany:
@Nir-Az I rebased this onto master, and refactored your changes to use `add_compile_options` on the main CMakeLists. This will make the options local to the build of Fast DDS, and avoid polluting the compile options of the dependent targets.

MiguelCompany:
@MiguelBarro would you mind reviewing my changes?

MiguelBarro (approved):
Clever workaround.
Instead of associating the settings to a target and propagating to dependents, associate to a directory and propagate to subdirs.
This way the examples and tests (which are beneath in the folder tree) inherit them but no user-dependent target will.
🧠👏

Nir-Az:
Sounds like a great solution to distribute the flags to all sub-projects, thanks.

MiguelCompany:
@richiprosima Please test Windows

PR #2521, opened 2022-02-24
[13259] Content filter implementation

MiguelCompany:
This PR adds reader-side content-filter implementation.

Co-Authored-By: Tomoya Fujita <Tomoya.Fujita@sony.com>

PR #2525, opened 2022-02-24
[13955] Generate TypeSupport using Fast DDS-Gen 2.1.1

JLBuenoLopez-eProsima:
This PR is built over #2521


MiguelCompany (changes requested):
Still some warnings on windows.

And it will also be great to have the linter issues solved

JLBuenoLopez-eProsima:
Windows warnings fixed in 99f5d752e3654f183316eede861fe18bcf225d2d.

Linters have not been fixed because there are auto generated files. Fast DDS Gen should either add the disabling macro for Uncrustify or comply with our code style. Otherwise, any modification introduced to these files is going to be lose when the files are updated in the next Fast DDS Gen release.

MiguelCompany (approved):

PR #2535, opened 2022-02-28
Make $<INSTALL_INTERFACE... use INCLUDE_INSTALL_DIR, and install headers to include/${PROJECT_NAME}

sloretz:
Same as https://github.com/eProsima/Fast-CDR/pull/120

The first commit is a bug fix. The exported targets currently always say the include directory is `include/` regardless of `INCLUDE_INSTALL_DIR`, so this fixes it by using `${INCLUDE_INSTALL_DIR}` instead.

The second commit is an enhancement supporting https://github.com/ros2/ros2/issues/1150


sloretz:
Thanks for the dockerfile, I made a few edits to make it build with these PRs.

<details><summary>Updated Dockerfile</summary>

```Dockerfile
FROM ubuntu:20.04

# Needed for a dependency that forces to set timezone
ENV TZ=Europe/Madrid
RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone

SHELL ["/bin/bash", "-c"]

WORKDIR /fastdds_ws

# Install build and run dependencies
RUN apt update && \
    apt install --yes --no-install-recommends \
        git \
        build-essential \
        cmake \
        libssl-dev \
        libasio-dev \
        libtinyxml2-dev \
        openjdk-8-jre-headless

ARG fastcdr_repo=https://github.com/eProsima/Fast-CDR.git
ARG fastcdr_version=master
ARG fastdds_repo=https://github.com/eProsima/Fast-DDS.git
ARG fastdds_version=master

# Install foonathan/memory
RUN git clone https://github.com/eProsima/foonathan_memory_vendor.git --depth 1 && \
    mkdir -p foonathan_memory_vendor/build && \
    cd foonathan_memory_vendor/build && \
    cmake .. -DCMAKE_INSTALL_PREFIX=/usr/local/ -DBUILD_SHARED_LIBS=ON && \
    cmake --build . --target install && \
    cd ../..

# Install Fast CDR
RUN git clone ${fastcdr_repo} fastcdr -b ${fastcdr_version} --depth 1 && \
    cd fastcdr && \
    mkdir -p build && \
    cd build && \
    cmake .. && \
    cmake --build . --target install --parallel 16 && \
    cd ../..

# Install Fast DDS
RUN git clone ${fastdds_repo} fastdds -b ${fastdds_version} && \
    cd fastdds && \
    mkdir -p build && \
    cd build && \
    cmake .. && \
    cmake --build . --target install --parallel 16 && \
    cd ../..

COPY test_fastdds_link.cpp .

# Compile simple example -- hardcoding the include directory to /usr/local/include
# RUN g++ test_fastdds_link.cpp -o test_fastdds_link -l fastcdr -l fastrtps -L /usr/local/lib -I /usr/local/include

# Compile simple example -- hardcoding the include directories to /usr/local/include/fastcdr and /usr/local/include/fastrtps
RUN g++ test_fastdds_link.cpp -o test_fastdds_link -l fastcdr -l fastrtps -L /usr/local/lib -I /usr/local/include/fastcdr -I /usr/local/include/fastrtps

# Run simple example
RUN LD_LIBRARY_PATH=/usr/local/lib ./test_fastdds_link
```
</details>

> This means, that upgrading Fast DDS would break the compilation of such program, which is something that cannot happen unless increasing major version.

That's true, if someone hard-codes the include directory then this PR would break them. I agree this change would require bumping the major version. I think adding a pkg-config file would make this easier on non-CMake users in the future ([See this example on CycloneDDS](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/PkgConfig.pc.in)).

> One option would be to only change the installation directory if certain new CMake option is passed when building the library.

Adding a CMake option makes sense. I'll add the same one I did for `urdfdom` to this PR: https://github.com/ros/urdfdom/pull/167 `APPEND_PROJECT_NAME_TO_INCLUDEDIR`.

> Furthermore, compiling ROS 2 from sources would become more cumbersome, as users would need to pass that option to avoid colcon issues. What do you think?

It becomes cumbersome in that the option would need to be passed, but that's doable. I think ROS 2 is already passing `--cmake-args -DSECURITY=ON` [for Fast-DDS](https://github.com/eProsima/Fast-DDS/blob/ba01ca6a87975c3daccce74c9af90d7621f6fdf5/CMakeLists.txt#L161), so the same solution could be done here.

sloretz:
@EduPonz I've updated this PR and eProsima/Fast-CDR#120 to have a CMake option `APPEND_PROJECT_NAME_TO_INCLUDEDIR` (matching ros/urdfdom#167) which defaults to `OFF`. When `ON`, `${PROJECT_NAME}` is appended to `INCLUDE_INSTALL_DIR`.

sloretz:
Friendly ping @MiguelBarro . Any thoughts on this PR now that eProsima/Fast-CDR#120 has been merged?

EduPonz:
Hi @sloretz ,

Sorry for the late reply. Our Quality Declaration enforces that all new features must be documented, would you mind submitting a PR to [Fast-DDS-docs](https://github.com/eProsima/Fast-DDS-docs/)? In particular, you'd need to document the new option [here](https://github.com/eProsima/Fast-DDS-docs/blob/master/docs/installation/configuration/cmake_options.rst). In any case, we are not going to block the merging of these PR until the documentation is ready; once we get @MiguelBarro approval I think we are good to go.

I'd like to take the opportunity to ask how are you planning on activating the flag on ROS 2 builds, specially from sources. I think [this](https://github.com/colcon/colcon-metadata-repository/blob/master/fastrtps.meta) would probably be a good place, although extending the [ROS 2 installation guide](https://docs.ros.org/en/rolling/Installation/Ubuntu-Development-Setup.html) would be required so that users [update their colcon metadata](https://colcon.readthedocs.io/en/released/reference/verb/metadata.html). In fact, I think this step is already relevant without the feature in this PR, since right now there is a mismatch between building from sources or getting the binary installation (regarding security) that is dangerous when compiling Fast DDS in an overlay.

MiguelBarro (approved):
LGTM

sloretz:
> would you mind submitting a PR to [Fast-DDS-docs](https://github.com/eProsima/Fast-DDS-docs/)?

Sure! https://github.com/eProsima/Fast-DDS-docs/pull/355

> I'd like to take the opportunity to ask how are you planning on activating the flag on ROS 2 builds, specially from sources. I think [this](https://github.com/colcon/colcon-metadata-repository/blob/master/fastrtps.meta) would probably be a good place, although extending the [ROS 2 installation guide](https://docs.ros.org/en/rolling/Installation/Ubuntu-Development-Setup.html) would be required so that users [update their colcon metadata](https://colcon.readthedocs.io/en/released/reference/verb/metadata.html). In fact, I think this step is already relevant without the feature in this PR, since right now there is a mismatch between building from sources or getting the binary installation (regarding security) that is dangerous when compiling Fast DDS in an overlay.

That's a good question, and great suggestions. I opened colcon/colcon-metadata-repository#7 to set the option for Fast-DDS and Fast-CDR. One idea is to patch the release repo to set the CMake option to cover the case when using Debian packages: https://github.com/ros2-gbp/fastcdr-release/issues/9 . I also opened a PR to make the ROS 2 CI jobs set the option, which includes the packaging jobs: https://github.com/ros2/ci/pull/629

EduPonz:
Lots of thanks @sloretz for taking care of it so promptly!

PR #2581, opened 2022-03-16
Fix publishing on physical data topic [14124]

EduPonz:
## Description

#2171 introduced a separation between initialization and enabling of `BuiltinProtocols`.
This had the side effect of the statistics' physical data not to be published, since the data was generated and added to the writer history only once on the creation of the corresponding `DataWriter`, which was performed prior to enabling the `DomainParticipantImpl`.
This lead to the `RTPSWriter` not to being created, as there was not `RTPSParticipant` (created when enabling the `DomainParticipantImpl`).
This PR fixes this behaviour by always enabling the `DomainParticipantImpl` before creating the statistics `DataWriter`s.

Furthermore, #2171 also introduced a regression in which removing a not enabled `RTPSParticipant` results in accessing a `nullptr` and therefore into a segmentation fault. This PR fixes this regression by checking whether the `PDP` instance is enabled within the call to `PDP::stopParticipantAnnouncement()`, which is the member function accessing to a `TimedEvent*`, which in turn is set to `nullptr` on non-enabled `PDP`s.
Moreover, the PR adds a regression test to ensure that it is always safe to remove a not enabled `RTPSParticipant`.

## Contributor Checklist
- [x] *N/a* New feature has been added to the `versions.md` file (if applicable).
- [x] *N/a* New feature has been documented/Current behavior is correctly described in the documentation. 

## Reviewer Checklist

### DIFF
--- a/include/fastdds/rtps/builtin/discovery/participant/PDP.h
+++ b/include/fastdds/rtps/builtin/discovery/participant/PDP.h
@@ -105,13 +105,6 @@ public:
      */
     bool enable();
 
-    /**
-     * @brief Check whether this PDP is enabled
-     *
-     * @return true if already enabled; false otherwise
-     */
-    bool is_enabled();
-
     virtual bool init(
             RTPSParticipantImpl* part) = 0;
 
@@ -415,7 +408,7 @@ protected:
     //!To protect callbacks (ParticipantProxyData&)
     std::mutex callback_mtx_;
     //!Tell if object is enabled
-    std::atomic<bool> enabled_ = {false};
+    std::atomic<bool> enabled_ {false};
 
     /**
      * Adds an entry to the collection of participant proxy information.
--- a/include/fastdds/rtps/builtin/discovery/participant/PDP.h
+++ b/include/fastdds/rtps/builtin/discovery/participant/PDP.h
@@ -21,6 +21,7 @@
 #define _FASTDDS_RTPS_PDP_H_
 #ifndef DOXYGEN_SHOULD_SKIP_THIS_PUBLIC
 
+#include <atomic>
 #include <mutex>
 #include <functional>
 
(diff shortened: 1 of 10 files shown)


include/fastdds/rtps/builtin/discovery/participant/PDP.h
```diff
@@ -402,7 +415,7 @@ class PDP
     //!To protect callbacks (ParticipantProxyData&)
     std::mutex callback_mtx_;
     //!Tell if object is enabled
-    bool enable_ = false;
+    std::atomic<bool> enabled_ = {false};
```
MiguelCompany, line 418:
```suggestion
    std::atomic<bool> enabled_ {false};
```

src/cpp/rtps/builtin/BuiltinProtocols.cpp
```diff
@@ -308,7 +308,10 @@ void BuiltinProtocols::stopRTPSParticipantAnnouncement()
 
     if (mp_PDP)
     {
-        mp_PDP->stopParticipantAnnouncement();
+        if (mp_PDP->is_enabled())
```
MiguelCompany, line 311: I would put this check inside `PDP::stopParticipantAnnouncement`. And then, I don't think we will need `PDP::is_enabled`.

MiguelCompany (changes requested):
Just two minor comments.

src/cpp/rtps/builtin/discovery/participant/PDP.cpp
```diff
@@ -554,7 +549,10 @@ void PDP::announceParticipantState(
 
 void PDP::stopParticipantAnnouncement()
 {
-    resend_participant_info_event_->cancel_timer();
+    if (enabled_)
```
MiguelCompany, line 552:
```suggestion
    if (resend_participant_info_event_)
```

src/cpp/rtps/builtin/discovery/participant/PDP.cpp
```diff
@@ -554,7 +549,10 @@ void PDP::announceParticipantState(
...
+    {
+        resend_participant_info_event_->cancel_timer();
+    }
 }
 
 void PDP::resetParticipantAnnouncement()
```
MiguelCompany, line 558: I think we should also protect this method

MiguelCompany (changes requested):

MiguelCompany (approved):
LGTM with green CI

MiguelCompany:
Unrelated CI failures.

PR #2536, opened 2022-03-01
Remove constructor exports for factory created entities [13947]

MiguelBarro:
## Description

Added missing exports required by the SWIG wrapper used in the python bindings.

## Contributor Checklist
- [x] Commit messages follow the project guidelines. 
- [x] The code follows the style guidelines of this project. 
- [x] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added. 
- [x] n/a Any new/modified methods have been properly documented using Doxygen. 
- [x] Fast DDS test suite has been run locally. 
- [ ] Changes are ABI compatible. 
- [x] Changes are API compatible. 
- [x] Documentation builds and tests pass locally. 
- [x] New feature has been added to the `versions.md` file (if applicable).
- [x] n/a New feature has been documented/Current behavior is correctly described in the documentation. 

## Reviewer Checklist
- [x] Check contributor checklist is correct.
- [x] Check CI results: changes do not issue any warning.
- [x] Check CI results: failing tests are unrelated with the changes.

PR #2585, opened 2022-03-18
[13791] Several doxydoc improvements

### LINKED ISSUE #2609: [14256] reader is_sample_valid() return always false when DataSharing=On

### Expected behavior

when using `DataSharing` without using the `loan` API.
`is_sample_valid()` should return true on the reading samples

### Current behavior

`is_sample_valid()` return  false

### Steps to reproduce

Usage:
On reader creation:
`rqos.data_sharing().automatic();`

On reader usage:
```
        if( _reader->wait_for_unread_message( { 1, 0 } ) )
        {
            devices data;
            SampleInfo info;
            bool device_update_detected = false;
            // Process all the samples until no one is returned
            while( ReturnCode_t::RETCODE_OK == _reader->take_next_sample( &data, &info ) )
            {
                if( info.valid_data && _reader->**is_sample_valid**( &data, &info ) )
                {
. . .
```

setting `rqos.data_sharing().off` fix it.

@MiguelBarro , I think it had to do with `is_sample_valid()` implementation, treating `DataSharing` only as loan messages API.
See here: https://github.com/eProsima/Fast-DDS/blob/e8581dbe3e58905c108017eaa14ed225bcd8f76f/src/cpp/rtps/reader/RTPSReader.cpp#L397

As it seems the RTPS 2.3 extra header does not apply on DataSharing only, right?
 `is_sample_valid()` is only useful when using `loan` API?

### Fast DDS version/commit

Release 2.6.0

### Platform/Architecture

Windows 10 Visual Studio 2019

### Transport layer

### Additional context

_No response_

### XML configuration file

_No response_

### Relevant log output

_No response_

### Network traffic capture

_No response_

MiguelCompany:
## Description
Adding missing doxydoc on ContentFilteredTopic methods.
Also removing some doc, as we don't need destructor or get_impl to be shown.

Closes #2609 

## Contributor Checklist
- [x] *N/A* Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added. 
- [ ] Documentation builds and tests pass locally.  <-- Leaving this to eProsima/Fast-DDS-docs#361
- [x] *N/A* New feature has been added to the `versions.md` file (if applicable).

## Reviewer Checklist

### DIFF
--- a/include/fastdds/dds/topic/ContentFilteredTopic.hpp
+++ b/include/fastdds/dds/topic/ContentFilteredTopic.hpp
@@ -67,14 +67,60 @@ public:
      */
     RTPS_DllAPI Topic* get_related_topic() const;
 
+    /**
+     * @brief Get the filter expression.
+     *
+     * This operation returns filter expression associated with this ContentFilteredTopic.
+     * It will return the @c filter_expression specified on the last successful call to @c set_expression or,
+     * if that method is never called, the expression specified when the ContentFilteredTopic was created.
+     *
+     * @return the @c filter_expression.
+     */
     RTPS_DllAPI const std::string& get_filter_expression() const;
 
+    /**
+     * @brief Get the expression parameters.
+     *
+     * This operation returns expression parameters associated with this ContentFilteredTopic.
+     * These will be the @c expression_parameters specified on the last successful call to @c set_expression or
+     * @c set_expression_parameters.
+     * If those methods have never been called, the expression parameters specified when the ContentFilteredTopic
+     * was created will be returned.
+     *
+     * @param [out] expression_parameters  The expression parameters currently associated with the
+     *                                     ContentFilteredTopic.
+     *
+     * @return RETCODE_OK
+     */
     RTPS_DllAPI ReturnCode_t get_expression_parameters(
             std::vector<std::string>& expression_parameters) const;
 
+    /**
+     * @brief Set the expression parameters.
+     *
+     * This operation changes expression parameters associated with this ContentFilteredTopic.
+     *
+     * @param [in] expression_parameters  The expression parameters to set.
+     *
+     * @return RETCODE_OK             if the expression parameters where correctly updated.
+     * @return RETCODE_BAD_PARAMETER  if the expression parameters do not match with the current @c filter_expression.
+     */
     RTPS_DllAPI ReturnCode_t set_expression_parameters(
             const std::vector<std::string>& expression_parameters);
 
+    /**
+     * @brief Set the filter expression and the expression parameters.
+     *
+     * This operation changes the filter expression and the expression parameters associated with this
+     * ContentFilteredTopic.
+     *
+     * @param [in] filter_expression      The filter expression to set.
+     * @param [in] expression_parameters  The expression parameters to set.
+     *
+     * @return RETCODE_OK             if the expression and parameters where correctly updated.
+     * @return RETCODE_BAD_PARAMETER  if @c filter_expression is not valid for this ContentFilteredTopic.
+     * @return RETCODE_BAD_PARAMETER  if the expression parameters do not match with the @c filter_expression.
+     */
     RTPS_DllAPI ReturnCode_t set_filter_expression(
             const std::string& filter_expression,
             const std::vector<std::string>& expression_parameters);
--- a/include/fastdds/dds/topic/ContentFilteredTopic.hpp
+++ b/include/fastdds/dds/topic/ContentFilteredTopic.hpp
@@ -55,9 +55,6 @@ private:
 
 public:
 
-    /**
-     * @brief Destructor
-     */
     RTPS_DllAPI virtual ~ContentFilteredTopic();
 
     /**
@@ -131,10 +128,6 @@ public:
      */
     virtual DomainParticipant* get_participant() const override;
 
-    /**
-     * @brief Getter for the TopicDescriptionImpl
-     * @return pointer to TopicDescriptionImpl
-     */
     TopicDescriptionImpl* get_impl() const override;
 
 protected:
--- a/include/fastdds/dds/topic/ContentFilteredTopic.hpp
+++ b/include/fastdds/dds/topic/ContentFilteredTopic.hpp
@@ -128,7 +128,7 @@ public:
      * @brief Getter for the DomainParticipant
      * @return DomainParticipant pointer
      */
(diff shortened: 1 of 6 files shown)


include/fastdds/dds/topic/ContentFilteredTopic.hpp
```diff
@@ -55,26 +56,70 @@ class ContentFilteredTopic : public TopicDescription
 
 public:
 
-    /**
-     * @brief Destructor
-     */
     RTPS_DllAPI virtual ~ContentFilteredTopic();
```
JLBuenoLopez-eProsima, line 59: Following #2536, I think the `RTPS_DllAPI` should be removed. However, in order to merge this PR, we cannot do it right now. Maybe issuing another PR with this change and setting the milestone to `2.7.0` is the way to go.

include/fastdds/dds/topic/ContentFilteredTopic.hpp
```diff
@@ -55,26 +56,70 @@ class ContentFilteredTopic : public TopicDescription
...
+     * This operation returns expression parameters associated with this ContentFilteredTopic.
+     * These will be the @c expression_parameters specified on the last successful call to @c set_expression or
+     * @c set_expression_parameters.
+     * If those methods have never been called, the expression parameters specified when the ContentFilteredTopic
+     * was created will be returned.
+     *
+     * @param [out] expression_parameters  The expression parameters currently associated with the
+     *                                     ContentFilteredTopic.
+     *
+     * @return RETCODE_OK
```
JLBuenoLopez-eProsima, line 92: Is this the only possible return code?

include/fastdds/dds/topic/ContentFilteredTopic.hpp
```diff
@@ -85,10 +130,6 @@ class ContentFilteredTopic : public TopicDescription
      */
     virtual DomainParticipant* get_participant() const override;
```
JLBuenoLopez-eProsima, line 131:
```suggestion
    DomainParticipant* get_participant() const override;
```

`override` is enough to indicate that the method is `virtual`.

Also, this method is not being exported in Windows (`RTPS_DllAPI`) neither in the base clase (`TopicDescription`) nor here, so it is only part of the public API in non-Windows platforms. I think we should fix and export this method if it is intended to be used by the user and if not, remove the doxygen documentation so it does not appear in the API reference (and consider making the method protected).

include/fastdds/dds/topic/ContentFilteredTopic.hpp
```diff
@@ -85,10 +130,6 @@ class ContentFilteredTopic : public TopicDescription
...
 
-    /**
-     * @brief Getter for the TopicDescriptionImpl
-     * @return pointer to TopicDescriptionImpl
-     */
     TopicDescriptionImpl* get_impl() const override;
```
JLBuenoLopez-eProsima, line 133: Maybe we should consider making the base method and the override implementations protected.

include/fastdds/dds/topic/TopicDescription.hpp
```diff
@@ -65,21 +65,10 @@ class TopicDescription
...
     virtual TopicDescriptionImpl* get_impl() const = 0;
 
 protected:
 
-    /**
-     * @brief Constructor using topic name and data type name
-     *
-     * @param name Name for the topic
-     * @param type_name Data type name
-     */
```
JLBuenoLopez-eProsima, line 82: I think this can be kept because being a protected method Doxygen does not include it in the API reference. Same applies to the destructor.

include/fastdds/dds/topic/Topic.hpp
```diff
@@ -75,9 +75,6 @@ class Topic : public DomainEntity, public TopicDescription
 
 public:
 
-    /**
-     * @brief Destructor
-     */
     virtual ~Topic();
```
JLBuenoLopez-eProsima, line 78: Maybe we should consider making this destructor protected.

JLBuenoLopez-eProsima (changes requested):
Some considerations. I think maybe a PR to be merged in `2.7.0` can be prepared with some of the suggestions.

include/fastdds/dds/topic/ContentFilteredTopic.hpp
MiguelCompany: Yes

include/fastdds/dds/topic/ContentFilteredTopic.hpp
MiguelCompany: I will do another PR removing the dll export.

include/fastdds/dds/topic/ContentFilteredTopic.hpp
MiguelCompany: Having the implementation getter public makes no harm, as the user does not have the implementation class declaration.

Making this a protected method will break API, and require adding several friend classes.

include/fastdds/dds/topic/ContentFilteredTopic.hpp
MiguelCompany: Removed the virtual on 55c498d, will do another PR adding the DLL export.

include/fastdds/dds/topic/Topic.hpp
MiguelCompany: That would break API. We may do it on a PR targetted for v3.0.

include/fastdds/dds/topic/TopicDescription.hpp
MiguelCompany: I think the arguments have self explanatory names, so I don't think the doxydoc adds useful information

MiguelCompany:
@JLBuenoLopez-eProsima I think I attended all your suggestions

JLBuenoLopez-eProsima (approved):
LGTM. I would like to see the PR adding the Dll exports or at least a ticket opened in Redmine to ensure that we do not forget it.

include/fastdds/dds/topic/ContentFilteredTopic.hpp
MiguelCompany: See #2611

include/fastdds/dds/topic/ContentFilteredTopic.hpp
MiguelCompany: See #2611

PR #2565, opened 2022-03-09
MessageReceiver refactor to fix deadlocks [14069] <feature/tsan/fixes>

MiguelBarro:
## Description

This refactor promotes the mutex to shared mutex in order to prevent deadlocks due to interaction with participant mutexes.

A) std::mutex MessageReceives::mtx_
B) std::recursive_mutex* RTPSParticipantImpl::mp_mutex 
C) std::mutex RTPSParticipantImpl::m_receiverResourcelistMutex

`MessageReceiver::associateEndpoint(C > A)` takes A while C is taken at `RTPSParticipantImpl::assignEndpoint2LocatorList()`

`EDP::pairing_writer_proxy_with_any_local_reader(A > B)` takes B while A is taken in `MessageReceiver::process_data_message_without_security()`

`RTPSParticipantImpl::deleteUserEndpoint(B > C)` takes C while B is taken in `RTPSParticipantImpl::disable()`

This is the first use of `shared_mutex` in Fast-DDS. Because it's a C++17 feature a public domain implementation is included in our utils folder to support old compilers.

## Contributor Checklist
- [x] Commit messages follow the project guidelines. 
- [x] The code follows the style guidelines of this project. 
- [x] [Sanitizer] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added. 
- [ ] NA Any new/modified methods have been properly documented using Doxygen. 
- [x] Fast DDS test suite has been run locally. 
- [x] Changes are ABI compatible. 
- [x] Changes are API compatible. 
- [ ] NA Documentation builds and tests pass locally. 
- [ ] NA New feature has been added to the `versions.md` file (if applicable).
- [ ] NA New feature has been documented/Current behavior is correctly described in the documentation. 

## Reviewer Checklist
- [x] Check contributor checklist is correct.
- [ ] Check CI results: changes do not issue any warning.
- [x] Check CI results: failing tests are unrelated with the changes.

PR #2600, opened 2022-03-24
 [14237] Fixing flow controller TSan lock-order-inversion reports <feature/tsan/fixes>

MiguelBarro:
## Description
This addresses the most common deadlock TSan report in Fast-DDS test suite (reported 317 times).
Now all `DataWriters` use a default asynchronous flow controller.
The associated asynchronous thread traverses the collection of pending changes to send.
For each change the `RTPSMessageSenderInterface` associated with the `DataWriter` is set into the `FlowController` auxiliary `RTPSMessageGroup`.
+ Before calling the `RTPSMessageGroup` setter the new `RTPSMessageSenderInterface` is locked (A).
+ The `RTPSMessageGroup` setter flushes pending messages before change the `RTPSMessageSenderInterface` locking the former `RTPSMessageSenderInterface` (B).

TSan reports that if the next change in the collection is delivered to A then we will lock B &rarr; A.
Note that this is not an issue because, despite been possible to have several `FlowController`s (each one with a devoted asynchronous thread), only one of them can be associated with a `DataWriter/RTPSMessageSenderInterface`. That makes an actual deadlock impossible.

Mitigation:
+ Add to the TSan's suppressions list. Not very elegant though.
+ Flush() the `FlowController`'s message group after each cache processing.
+ Don't lock the `RTPSMessageSenderInterface` before calling the `RTPSMessageGroup` setter.

I've chosen the last option:
+ It's safe because the `RTPSMessageSenderInterface` is not modified inside the setter.
+ Manually flushing may multiply the number of messages if several consecutive changes have the same destination.

Merge after https://github.com/eProsima/Fast-DDS/pull/2565

## Contributor Checklist
- [ ] NA Any new/modified methods have been properly documented using Doxygen. 
- [ ] NA Documentation builds and tests pass locally. 

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
@@ -447,11 +447,13 @@ void RTPSParticipantImpl::enable()
 
 void RTPSParticipantImpl::disable()
 {
-    // Ensure that other participants will not accidentally discover this one
-    if (mp_builtinProtocols && mp_builtinProtocols->mp_PDP)
+    if (nullptr == mp_builtinProtocols)
     {
-        mp_builtinProtocols->stopRTPSParticipantAnnouncement();
+        return;
     }
+
+    // Ensure that other participants will not accidentally discover this one
+    stopRTPSParticipantAnnouncement();
 
     // Disable Retries on Transports
     m_network_Factory.Shutdown();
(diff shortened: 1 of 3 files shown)


src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -447,12 +447,14 @@ void RTPSParticipantImpl::enable()
...
-    if (mp_builtinProtocols && mp_builtinProtocols->mp_PDP)
+    if (nullptr == mp_builtinProtocols)
     {
-        mp_builtinProtocols->stopRTPSParticipantAnnouncement();
+        return;
     }
 
+    // Ensure that other participants will not accidentally discover this one
+    stopRTPSParticipantAnnouncement();
+
```
jsantiago-eProsima, line 457: Removing the `mp_builtinProtocols->mp_PDP` check will log an error when trying to disable a participant with `DiscoveryProtocol::None`. Maybe we should only call `stopRTPSParticipantAnnouncement` with a valid `mp_builtinProtocols->mp_PDP` here (as it was done before).

jsantiago-eProsima (changes requested):

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
MiguelBarro: Absolutely right. I introduced the fixed in ec657ef

jsantiago-eProsima:
LGTM with green CI

PR #2350, opened 2021-12-15
Add new exchange format to Static Discovery [13266]

richiware:
This PR adds a new exchange format of Static Discovery's information which reduces the network bandwidth usage.

Documentation: eProsima/Fast-DDS-docs#358

PR #2628, opened 2022-04-10
Refs 14322 Refactor SecurityManager synchronization <feature/tsan/fixes>

MiguelBarro:
## Description

Refactor Guidelines:
+ What to do with const-use of members and references to the plugins?
  - the pointer members are initialized on `init()` which is called from participant constructor.
  - the pointer members are removed on `destroy()` which is called from ~SecurityManager() from the participant destructor. 
  - meanwhile `discovered_participant` and `on_process_handshake` are called from transport threads and leads to a myriad of member calls. The issue here is that destroy is called concurrently with those transport calls and leads to a race.
 
  The initial approach was to promote them to smart pointers able to guarantee unique getters and setters [see](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4058.pdf)
      but this 2014 proposal of an atomic<unique_ptr<T>> had no success. Though on 2017 the standard admitted
        `__cpp_lib_atomic_shared_ptr (c++20) -> atomic<shared_ptr<T>>`
   
   Finally a sentry based approach was favoured. The sentry requirements are:
  + It protects the plugins and other pointer based ancillary that doesn't require sync per-se but only `nullptr` checking.
  + It uses RAII strategy, thus it must be a method that returns a temporary object. While this object is *alive* the pointers cannot be modified.
  + The `SecurityManager::destroy()` method should wait till all the transport driven calls are done.
  Note that all methods call from `destroy()` must be sentry free to allow their execution.
  Note also that the `destroy()` must always called from a sentry free stack.

+ Collections must follow C++ const conventions. For example:
   ```c++
          std::map<GUID_t, DiscoveredParticipantInfo> discovered_participants_;
   ```
  in C++ philosophy this implies that the collection elements are like members and cannot be modified from `SecurityManager` const methods. This obviously doesn't fit in the actual use of them. Enconding/decoding operations should be const for the SecurityManager because it doesn't change its state but these members state. I'm going to promote them to:
   ```c++
          std::map<GUID_t, unique_ptr<DiscoveredParticipantInfo>> discovered_participants_;
   ```
+ Promote current mutex to `shared_mutex`. The write lock to protect changes on members (like collections update) and references (`new`/`delete`).
+ Collection members update must resort to their own sync methods. This was the premise before seeing how  
      `DiscoveredParticipantInfo` guarantees atomical update of its `AuthenticationInfo` members by using `unique_ptr<>` in
      the getters and setters (though the getters and setters must use another sync mechanism).  
+ Note that the previous mutex was recursive. Now care must be taken to avoid reentrancy.

After testing these changes it was detected that the sentry approach cannot prevent data removal due to discovery callbacks.
That is when a participant or endpoint was removed nothing prevent its encoding keys from being removed while other threads were using them to encode/decode a message.
In order to prevent this a new handle strategy was enforced:
+ Reference counting was added to the keys and other info required for encoding/decoding. This way even if the participant or endpoint info is removed only when the encode/decode operations are finished the actual removal would take place.
+ The handles cannot be created arbitrarily as before but class factories are assigned. This way the heap allocations and deallocations are prevented outside of the controlled factory code. Using this strategy the reference counting block uniqueness is assured.

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
+++ b/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
@@ -144,25 +144,78 @@ public:
             SecurityException& exception) = 0;
 
     /**
-     * Releases resources associated with a DataWriter. The Crypto Handle becomes unusable after this
+     * Releases resources associated with a DataWriter. The Crypto Handle may become unusable after this
      * @param datawriter_crypto_handle Belonging to the DataWriter that awaits termination
      * @param exception (out) Security exception
      * @return TRUE is successful
      */
     virtual bool unregister_datawriter(
-            DatawriterCryptoHandle* datawriter_crypto_handle,
+            std::shared_ptr<DatawriterCryptoHandle>& datawriter_crypto_handle,
             SecurityException& exception) = 0;
 
     /**
-     * Releases resources associated with a DataReader. The Crypto Handle becomes unusable after this
+     * Convenient override for raw pointers arguments.
+     * @param datawriter_crypto_handle Belonging to the DataWriter that awaits termination
+     * @param exception (out) Security exception
+     * @return TRUE is successful
+     */
+    bool unregister_datawriter(
+            DatawriterCryptoHandle* datawriter_crypto_handle,
+            SecurityException& exception)
+    {
+
+        if (nullptr == datawriter_crypto_handle)
+        {
+            return false;
+        }
+
+        try
+        {
+            auto temp = datawriter_crypto_handle->shared_from_this();
+            return unregister_datawriter(temp, exception);
+        }
+        catch(std::bad_weak_ptr&)
+        {
+            return false;
+        }
+    }
+
+    /**
+     * Releases resources associated with a DataReader. The Crypto Handle may become unusable after this
      * @param datareader_crypto_handle Belonging to the DataReader that awaits termination
      * @param exception (out) Security exception
      * @return TRUE is successful
      */
     virtual bool unregister_datareader(
-            DatareaderCryptoHandle* datareader_crypto_handle,
+            std::shared_ptr<DatareaderCryptoHandle>& datareader_crypto_handle,
             SecurityException& exception) = 0;
 
+    /**
+     * Convenient override for raw pointers arguments.
+     * @param datareader_crypto_handle Belonging to the DataWriter that awaits termination
+     * @param exception (out) Security exception
+     * @return TRUE is successful
+     */
+    bool unregister_datareader(
+            DatareaderCryptoHandle* datareader_crypto_handle,
+            SecurityException& exception)
+    {
+
+        if (nullptr == datareader_crypto_handle)
+        {
+            return false;
+        }
+
+        try
+        {
+            auto temp = datareader_crypto_handle->shared_from_this();
+            return unregister_datareader(temp, exception);
+        }
+        catch(std::bad_weak_ptr&)
+        {
+            return false;
+        }
+    }
 
 };
 
--- a/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
+++ b/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
@@ -174,7 +174,7 @@ public:
             auto temp = datawriter_crypto_handle->shared_from_this();
             return unregister_datawriter(temp, exception);
         }
-        catch(std::bad_weak_ptr&)
+        catch (std::bad_weak_ptr&)
         {
             return false;
         }
@@ -211,7 +211,7 @@ public:
             auto temp = datareader_crypto_handle->shared_from_this();
             return unregister_datareader(temp, exception);
         }
-        catch(std::bad_weak_ptr&)
+        catch (std::bad_weak_ptr&)
         {
             return false;
         }
--- a/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
+++ b/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
@@ -68,7 +68,7 @@ public:
             const ParticipantCryptoHandle& local_participant_crypto_handle,
             const IdentityHandle& remote_participant_identity,
             const PermissionsHandle& remote_participant_permissions,
-            const SharedSecretHandle& shared_secret,
+            const SecretHandle& shared_secret,
             SecurityException& exception) = 0;
 
     /**
@@ -99,7 +99,7 @@ public:
     virtual DatareaderCryptoHandle* register_matched_remote_datareader(
             DatawriterCryptoHandle& local_datawriter_crypto_handle,
             ParticipantCryptoHandle& remote_participant_crypto,
-            const SharedSecretHandle& shared_secret,
+            const SecretHandle& shared_secret,
             const bool relay_only,
             SecurityException& exception) = 0;
 
@@ -130,7 +130,7 @@ public:
     virtual DatawriterCryptoHandle* register_matched_remote_datawriter(
             DatareaderCryptoHandle& local_datareader_crypto_handle,
             ParticipantCryptoHandle& remote_participant_crypt,
-            const SharedSecretHandle& shared_secret,
+            const SecretHandle& shared_secret,
             SecurityException& exception) = 0;
 
     /**
--- a/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
+++ b/include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
@@ -22,6 +22,8 @@
 #include <fastdds/rtps/security/accesscontrol/EndpointSecurityAttributes.h>
 #include <fastdds/rtps/security/accesscontrol/ParticipantSecurityAttributes.h>
 
+#include <memory>
+
 namespace eprosima {
 namespace fastrtps {
 namespace rtps {
@@ -45,7 +47,7 @@ public:
      * @param exception (out) Security exception
      * @return ParticipantCryptoHandle with generated key material
      */
-    virtual ParticipantCryptoHandle* register_local_participant(
+    virtual std::shared_ptr<ParticipantCryptoHandle> register_local_participant(
             const IdentityHandle& participant_identity,
             const PermissionsHandle& participant_permissions,
             const PropertySeq& participant_properties,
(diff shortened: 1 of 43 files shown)


MiguelBarro:
@richiprosima Please be so gentle to test Windows for me 😎

test/unittest/rtps/security/SecurityTests.cpp
```diff
@@ -39,8 +39,10 @@ void SecurityTest::initialization_ok()
             WillOnce(DoAll(SetArgPointee<0>(stateless_reader_), Return(true))).
             WillOnce(DoAll(SetArgPointee<0>(volatile_reader_), Return(true)));
 
-    ASSERT_TRUE(manager_.init(security_attributes_, participant_properties_, security_activated_));
-    ASSERT_TRUE(!security_activated_ || manager_.create_entities());
+    security_activated_ = manager_.init(security_attributes_, participant_properties_);
+    ASSERT_TRUE(security_activated_);
+    ASSERT_TRUE(manager_.is_security_initialized());
+    ASSERT_TRUE(!manager_.is_security_initialized() || manager_.create_entities());
```
jsantiago-eProsima, line 45:
```suggestion
    ASSERT_TRUE(manager_.create_entities());
```
`!manager_.is_security_initialized()` can't evaluate `true` or else the previous ASSERT would fail and abort the test.

test/unittest/rtps/security/SecurityTests.cpp
```diff
@@ -60,8 +62,10 @@ void SecurityTest::initialization_auth_ok()
     EXPECT_CALL(participant_, createReader_mock(_, _, _, _, _, _, _)).Times(1).
             WillOnce(DoAll(SetArgPointee<0>(stateless_reader_), Return(true)));
 
-    ASSERT_TRUE(manager_.init(security_attributes_, participant_properties_, security_activated_));
-    ASSERT_TRUE(!security_activated_ || manager_.create_entities());
+    security_activated_ = manager_.init(security_attributes_, participant_properties_);
+    ASSERT_TRUE(security_activated_);
+    ASSERT_TRUE(manager_.is_security_initialized());
+    ASSERT_TRUE(!manager_.is_security_initialized() || manager_.create_entities());
```
jsantiago-eProsima, line 68:
```suggestion
    ASSERT_TRUE(manager_.create_entities());
```
`!manager_.is_security_initialized()` can't evaluate `true` or else the previous ASSERT would fail and abort the test.

src/cpp/rtps/security/SecurityManager.h
```diff
@@ -148,70 +192,124 @@ class SecurityManager
             const PropertyPolicy& reader_properties,
             EndpointSecurityAttributes& security_attributes);
 
+    /**
+     * Returns a handle to the identity token used in handshake
```
jsantiago-eProsima, line 196: I think this is misleading since the actual return type is bool. Maybe use something like "Retrieves a handle". Same comment for the rest of the instances here.

src/cpp/rtps/security/SecurityManager.cpp
```diff
@@ -566,21 +543,18 @@ bool SecurityManager::restore_discovered_participant_info(
...
 
-    std::unique_lock<std::mutex> lock(mutex_);
+    shared_lock<shared_mutex> _(mutex_);
+
     auto dp_it = discovered_participants_.find(remote_participant_key);
 
     if (dp_it != discovered_participants_.end())
     {
-        dp_it->second.set_auth(auth_ptr);
+        dp_it->second->set_auth(auth_ptr);
```
jsantiago-eProsima, line 552: This collection is protected for reading only. This might be a data race.

src/cpp/rtps/security/SecurityManager.cpp
```diff
@@ -3972,7 +4172,7 @@ void SecurityManager::resend_handshake_message_token(
                 }
             }
 
-            dp_it->second.set_auth(remote_participant_info);
+            dp_it->second->set_auth(remote_participant_info);
```
jsantiago-eProsima, line 4175: This collection is locked for read only. This might be a data race.

jsantiago-eProsima (changes requested):

src/cpp/rtps/security/SecurityManager.cpp
jsantiago-eProsima: I realized this is not the case. Resolving conversation.

src/cpp/rtps/security/common/SharedSecretHandle.cpp
```diff
@@ -22,29 +22,16 @@ using namespace eprosima::fastrtps::rtps::security;
...
-const std::vector<uint8_t>* SharedSecretHelper::find_data_value(const SharedSecret& sharedsecret, const std::string& name)
+const std::vector<uint8_t>* SharedSecretHelper::find_data_value(
+        const SecretHandle& secret,
+        const std::string& name)
 {
     const std::vector<uint8_t>* returnedValue = nullptr;
+    const SharedSecret& sharedsecret = **SharedSecretHandle::narrow(secret);
 
-    for(auto property = sharedsecret.data_.begin(); property != sharedsecret.data_.end(); ++property)
+    for (auto property = sharedsecret.data_.begin(); property != sharedsecret.data_.end(); ++property)
```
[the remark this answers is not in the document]
MiguelBarro: You are right. `narrow` may return a `nil` if the `Handle` type doesn't match.

include/fastdds/rtps/security/cryptography/CryptoKeyFactory.h
```diff
@@ -138,29 +140,82 @@ class CryptoKeyFactory
      * @return TRUE is successful
      */
     virtual bool unregister_participant(
-            ParticipantCryptoHandle* participant_crypto_handle,
+            std::shared_ptr<ParticipantCryptoHandle>& participant_crypto_handle,
             SecurityException& exception) = 0;
 
     /**
-     * Releases resources associated with a DataWriter. The Crypto Handle becomes unusable after this
+     * Releases resources associated with a DataWriter. The Crypto Handle may become unusable after this
```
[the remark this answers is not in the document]
MiguelBarro: This method clear the framework references to the writer keys.
That is, the participant key keeps `shared_ptrs` to its readers and writers keys. Calling this method the participant reset that ptr.
If some decoding is undergone while this method is called the handle is still valid because it will be clear when the last stack reference is gone.
And that's the brief explanation on the use of *may become* and not *becomes*.

include/fastdds/rtps/security/cryptography/CryptoTransform.h
```diff
@@ -22,6 +22,8 @@
 #include <fastdds/rtps/common/CDRMessage_t.h>
 #include <fastdds/rtps/common/SerializedPayload.h>
 
+#include <memory>
+
```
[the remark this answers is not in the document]
MiguelBarro: acknowledge

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -325,8 +325,10 @@ RTPSParticipantImpl::RTPSParticipantImpl(
 
 #if HAVE_SECURITY
     // Start security
-    if (!m_security_manager.init(security_attributes_, PParam.properties,
-            m_is_security_active))
+    // TODO(Ricardo) Get returned value in future.
```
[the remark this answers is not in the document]
MiguelBarro: Nope. The *TODOer* decided here (https://github.com/eProsima/Fast-DDS/pull/2350) to withdraw from the TODO.
So I think is good to remove it.

test/unittest/rtps/security/SecurityValidationRemoteTests.cpp
```diff
@@ -37,7 +37,7 @@ TEST_F(SecurityTest, discovered_participant_validation_remote_identity_ok)
 {
     initialization_auth_ok();
 
-    MockIdentityHandle remote_identity_handle;
+    auto& remote_identity_handle = get_handle<MockIdentityHandle>();
```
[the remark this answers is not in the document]
MiguelBarro: It must because is generated with `SecurityTest::get_handle()`. Only the generated with `SecurityTest::get_sh_ptr()` don't need to return.

test/unittest/rtps/security/SecurityValidationRemoteTests.cpp
```diff
@@ -62,7 +62,7 @@ TEST_F(SecurityTest, discovered_participant_validation_remote_identity_pending_h
 {
     initialization_ok();
 
-    MockIdentityHandle remote_identity_handle;
+    auto& remote_identity_handle = get_handle<MockIdentityHandle>();
```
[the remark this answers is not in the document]
MiguelBarro: done

test/unittest/rtps/security/SecurityValidationRemoteTests.cpp
```diff
@@ -81,7 +81,7 @@ TEST_F(SecurityTest, discovered_participant_validation_remote_identity_pending_h
...
```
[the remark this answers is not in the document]
MiguelBarro: done

src/cpp/security/cryptography/AESGCMGMAC_KeyFactory.cpp
```diff
@@ -215,8 +254,15 @@ ParticipantCryptoHandle* AESGCMGMAC_KeyFactory::register_matched_remote_particip
     bool is_origin_auth =
             (plugin_attrs & PLUGIN_PARTICIPANT_SECURITY_ATTRIBUTES_FLAG_IS_RTPS_ORIGIN_AUTHENTICATED) != 0;
 
+
```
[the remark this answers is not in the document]
MiguelBarro: done

test/unittest/rtps/security/SecurityTests.cpp
MiguelBarro: right

test/unittest/rtps/security/SecurityTests.cpp
MiguelBarro: done

src/cpp/rtps/security/SecurityManager.h
MiguelBarro: done

jsantiago-eProsima (approved):
LGTM with green CI.

PR #2646, opened 2022-04-20
[14385] Avoid data race on UDPChannelResource

MiguelCompany:
## Description

This PR moves the close operation over the UDP socket from `UDPChannelResource::release` to the destructor of `UDPChannelResource`.

As the destructor is called after calling `UDPChannelResource::clear`, the receiving thread would have been already joined, so the thread sanitizer should not complain.

As a bonus, I introduced a small change on `TCPChannelResourceBasic::disconnect` that I think is responsible for leaving TCP sockets on a FIN_2 state sometimes.

@Mergifyio backport 2.5.x 2.3.x 2.1.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/transport/UDPChannelResource.cpp
+++ b/src/cpp/rtps/transport/UDPChannelResource.cpp
@@ -45,6 +45,9 @@ UDPChannelResource::UDPChannelResource(
 UDPChannelResource::~UDPChannelResource()
 {
     message_receiver_ = nullptr;
+
+    asio::error_code ec;
+    socket()->close(ec);
 }
 
 void UDPChannelResource::perform_listen_operation(
--- a/src/cpp/rtps/transport/UDPChannelResource.cpp
+++ b/src/cpp/rtps/transport/UDPChannelResource.cpp
@@ -120,8 +120,11 @@ void UDPChannelResource::release()
     // in Windows and Linux anyways, which is what we want.
     asio::error_code ec;
     socket()->shutdown(asio::socket_base::shutdown_type::shutdown_receive, ec);
-    // On OSX shutdown does not unblock the listening thread, but close does.
+
+#if defined(__APPLE__)
+    // On OSX shutdown does not seem to unblock the listening thread, but close does.
     socket()->close();
+#endif
 }
 
 } // namespace rtps
--- a/src/cpp/rtps/transport/UDPChannelResource.cpp
+++ b/src/cpp/rtps/transport/UDPChannelResource.cpp
@@ -124,7 +124,7 @@ void UDPChannelResource::release()
 #if defined(__APPLE__)
     // On OSX shutdown does not seem to unblock the listening thread, but close does.
     socket()->close();
-#endif
+#endif // if defined(__APPLE__)
 }
 
 } // namespace rtps
--- a/src/cpp/rtps/transport/TCPChannelResourceBasic.cpp
+++ b/src/cpp/rtps/transport/TCPChannelResourceBasic.cpp
@@ -115,11 +115,6 @@ void TCPChannelResourceBasic::disconnect()
                     try
                     {
                         socket->cancel();
-
-                        // This method was added on the version 1.12.0
-#if ASIO_VERSION >= 101200 && (!defined(_WIN32_WINNT) || _WIN32_WINNT >= 0x0603)
-                        socket->release();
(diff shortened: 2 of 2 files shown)


src/cpp/rtps/transport/UDPChannelResource.cpp
```diff
@@ -45,6 +45,9 @@ UDPChannelResource::UDPChannelResource(
 UDPChannelResource::~UDPChannelResource()
 {
     message_receiver_ = nullptr;
+
+    asio::error_code ec;
+    socket()->close(ec);
```
jsantiago-eProsima, line 50: Why not use close() instead? We are not performing any error checking here.

jsantiago-eProsima (changes requested):

xbreak:
In `UDPChannelResource::Receive()` with the call `socket()->receive_from(...)`: What happens now if there is nothing to receive?
When does call fail so the thread can detect that it should exit? The call has no timeout after all, and what previously triggered a read error due to closing the socket, no longer happens because thread is joined first?

MiguelCompany:
> In `UDPChannelResource::Receive()` with the call `socket()->receive_from(...)`: What happens now if there is nothing to receive? When does call fail so the thread can detect that it should exit? The call has no timeout after all, and what previously triggered a read error due to closing the socket, no longer happens because thread is joined first?

The trick is using `shutdown`, which happens to unblock the `receive_from` call, which returns an error

xbreak:
> 


> > In `UDPChannelResource::Receive()` with the call `socket()->receive_from(...)`: What happens now if there is nothing to receive? When does call fail so the thread can detect that it should exit? The call has no timeout after all, and what previously triggered a read error due to closing the socket, no longer happens because thread is joined first?
> 
> The trick is using `shutdown`, which happens to unblock the `receive_from` call, which returns an error

Is that not also a data race then as [ASIO document it to be thread _unsafe_](https://www.boost.org/doc/libs/1_76_0/doc/html/boost_asio/reference/ip__udp/socket.html)?

MiguelCompany:
> Is that not also a data race then as [ASIO document it to be thread _unsafe_](https://www.boost.org/doc/libs/1_76_0/doc/html/boost_asio/reference/ip__udp/socket.html)?

The docs do not explicitly mention `shutdown`, and in my local testing, the thread sanitizer did not complain. I did investigate it a bit further. and I think the race is on the socket descriptor. Calling `close` will mark the descriptor as ready to be reused, so if another socket is open, the same descriptor could be returned by the OS, and the `recvfrom` may receive data from the new socket. As the `shutdown` call is not freeing the resource (i.e. the descriptor) there is no race in this case.

xbreak:
> The docs do not explicitly mention `shutdown`, and in my local testing, the thread sanitizer did not complain. I did investigate it a bit further. and I think the race is on the socket descriptor. Calling `close` will mark the descriptor as ready to be reused, so if another socket is open, the same descriptor could be returned by the OS, and the `recvfrom` may receive data from the new socket. As the `shutdown` call is not freeing the resource (i.e. the descriptor) there is no race in this case.

It may or may not be safe for the tested platform and version, but from what I read it's _explicitly unsafe_ as socket is documented as _thread unsafe_ with the exception of a few enumerated operations. `shutdown` is not listed as safe and assuming so could break things for different OS or ASIO version.

The limitations of the synchronous interface is well known and most often the recommendation seems to be to use the async version to support cancellation or timeouts (see e.g. https://github.com/boostorg/asio/blob/develop/example/cpp11/timeouts/blocking_udp_client.cpp). This may come with an performance penalty though.

Excerpt from https://www.boost.org/doc/libs/1_76_0/doc/html/boost_asio/reference/ip__udp/socket.html:

> Thread Safety
> Distinct objects: Safe.
> Shared objects: **Unsafe.**
> Synchronous send, send_to, receive, receive_from, and connect operations are **thread safe with respect to each other,** if the underlying operating system calls are also thread safe. This means that it is permitted to perform concurrent calls to these synchronous operations on a single socket object. **Other synchronous operations, such as open or close, are not thread safe.**

JLBuenoLopez-eProsima (approved):
LGTM

MiguelCompany:
@xbreak I opened chriskohlhoff/asio#1046 so the ASIO maintainers tell us whether the `shutdown` operation is thread-safe.

In the meantime, we are going to merge and backport this PR, since it solves the data race on the supported platforms.

PR #2654, opened 2022-04-26
[14430] Fix freezes in the RTPS level when transferring large files

sfod:
Signed-off-by: Igor Abdrakhimov <abdrahimov@arrival.com>

## Description

When sending large files (>16 MB), the transfer may hang on the RTPS level. This happens because of how `ChangeForReader_t` handles fragments.

### Details

The issue appears when on `ChangeForReader_t::markFragmentsAsUnsent` call, if `unsentFragments` contains only one fragment and this fragment is less than `unsent_fragments_.base_` by >255 fragments.
Then on a `ChangeForReader_t::markFragmentsAsSent` call, `unsent_fragments_` becomes empty. So, `base_update` and `add_range` won't be called for `unsent_fragments_`.
Which leads to `ChangeForReader_t::get_next_unsent_fragment` reporting that all fragments are sent.

Now, a hang will happen if a Reader request for an unsent change which is greater than the previously requested fragment by >255 fragments. Surprisingly, it happens more than I'd expect.

### Notes

While fixing this issue, a couple of other things caught my attention.

1. When I transfer ~100 MB of data, a Reader always requests for few fragments starting from the 6th fragment. This happens because  of this statement in `StatefulReader::processDataFragMsg`:
    ```
    work_change->setFragmentSize(change_to_add->getFragmentSize(), true);
    ```
    It takes ~2 ms on my system. During this period, a socket buffer fills up and few UDP packets are lost. Increasing a buffer might help, but on some systems and for larger data this period might be much higher, so I don't think it's a good solution.
    This issue may seem minor, but it causes the next issue.

2. If an unsent fragment is less than the last sent fragment by >255 fragments, a Writer will resend all fragments again, starting with the undelivered one. On my local experiments, 10% of losses lead to 250% of data being transferred because of this issue.
Considering that the incoming data is handled synchronously, any delay (caused by system overload, for example) can cause a significant increase of the network traffic.

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/include/fastdds/rtps/writer/ChangeForReader.h
+++ b/include/fastdds/rtps/writer/ChangeForReader.h
@@ -133,7 +133,7 @@ public:
         if (unsent_fragments_.empty())
         {
             base = highest_sent_fragment_ + 1u;
-            max = highest_sent_fragment_ + 1u;
+            max = highest_sent_fragment_;
         }
         else
         {
@@ -141,6 +141,11 @@ public:
             max = unsent_fragments_.max();
             assert(!unsent_fragments_.is_set(base));
             base = unsent_fragments_.min();
+
+            if (max < highest_sent_fragment_)
+            {
+                max = highest_sent_fragment_;
+            }
         }
 
         if (max < change_->getFragmentCount())
--- a/include/fastdds/rtps/writer/ChangeForReader.h
+++ b/include/fastdds/rtps/writer/ChangeForReader.h
@@ -123,6 +123,8 @@ public:
     {
         unsent_fragments_.remove(sentFragment);
 
+        // We only use the running window mechanism during the first stage, until all fragments have been delivered
+        // once, and we consider the whole change as delivered.
         if (!delivered_ && !unsent_fragments_.empty() && (unsent_fragments_.max() < change_->getFragmentCount()))
         {
             FragmentNumber_t base = unsent_fragments_.base();
@@ -141,9 +143,30 @@ public:
     void markFragmentsAsUnsent(
             const FragmentNumberSet_t& unsentFragments)
     {
-        if (delivered_ && unsent_fragments_.empty())
+        // Ignore NACK_FRAG messages during the first stage, until all fragments have been delivered once, and we
+        // consider the whole change as delivered.
+        if (delivered_)
         {
-            unsent_fragments_ = unsentFragments;
+            if (unsent_fragments_.empty())
+            {
+                // Current window is empty, so we can set it to the received one.
+                unsent_fragments_ = unsentFragments;
+            }
+            else
+            {
+                // Update window to send the lowest possible requested fragments first.
+                FragmentNumber_t other_base = unsentFragments.base();
+                if (other_base < unsent_fragments_.base())
+                {
+                    unsent_fragments_.base_update(other_base);
+                }
+                unsentFragments.for_each(
+                    [this](
+                        FragmentNumber_t element)
+                    {
+                        unsent_fragments_.add(element);
+                    });
+            }
         }
     }
 
--- a/include/fastdds/rtps/writer/ChangeForReader.h
+++ b/include/fastdds/rtps/writer/ChangeForReader.h
@@ -150,16 +150,25 @@ public:
     {
         unsent_fragments_.remove(sentFragment);
 
-        if (!unsent_fragments_.empty() && unsent_fragments_.max() < change_->getFragmentCount())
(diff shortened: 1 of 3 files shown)


MiguelCompany:
@richiprosima Please be so gentle to test mac again here

sfod:
@MiguelCompany Yes, your understanding of the issue is correct. And you're right, the Reader eventually sends `NACK_FRAG`, but because of the `base_` is too small, no fragments are marked as unsent on the Writer side.

Regarding the fix you're suggesting, that was my first idea also. However, with this fix, `ChangeForReader_t` will report that the change is delivered after it resends the 6th fragment (I'm referring to the example you provided). It may work just fine now, but I can easily imagine how things might break because of some unrelated changes in the future (for example, handling of a delivered flag changes, or `ChangeForReader_t` reused for another functionality). In other words, the cause of the issue lies in the `markFragmentsAsSent` method, and I believe it should be fixed there.

Saying that, if you think that changing the `markFragmentsAsUnsent` is a proper way to fix the issue, I'll gladly do this. I just wanted to share my view on things.

MiguelCompany:
@sfod I pushed some commits with what I think is the correct solution.

1. Commit 633bf80b2b361015dcee35547cb8ad1270c4d6f5 with the change I proposed.
2. Commit 5a2b1a054b7e1df2a946bd6a2bf88022d5c05637 that extends your that, when `unsent_changes_` becomes empty,  resumes sending from the highest fragment ever sent
3. An additional commit to remove the custom destructor and copy constructors, so the default ones are always used.

sfod:
@MiguelCompany I tried to fix the issue with resending all fragments, but I ended wit a messy solution. Maybe this problem worth a separate PR with its own tests?

MiguelCompany:
@sfod
> Maybe this problem worth a separate PR with its own tests?

When fixing a bug, I take the opportunity to reconsider design decisions. I've been rethinking this a bit more, and I think we should proceed the following way:

We need two stages:
1. An initial stage where we use the running window mechanism until all fragments are sent at least once. During this stage we will ignore calls to `markFragmentsAsUnsent`. This has the advantage that all fragments are sent to all readers at the same time during this stage, saving CPU time on the writer, as the same RTPS datagram can be sent to all readers.
2. After all fragments have been sent once, we only accept calls to `markFragmentsAsUnsent` when `unsent_fragments_` is empty. 

If you agree with my proposal, would you mind adding the corresponding tests to ReaderProxyTests?
I will think of a black-box test and add it [here](https://github.com/eProsima/Fast-DDS/blob/master/test/blackbox/common/BlackboxTestsPubSubFragments.cpp)

MiguelCompany:
> What do you think?

🤔 I think you are right. We should go for the first behavior you describe.

```
1 - 2 - 3 - 4 - 5 - 6 - 7 - 8 - 9 - 3 - 4 - 10 - 11 - 12
                                  NACK_FRAG arrived
```

I think I know how to do it. Would you mind updating the tests to reflect that as the expected behavior? This way we will follow TDD 😄

sfod:
@MiguelCompany Sure thing, I'll come up with new tests today :smiley:

MiguelCompany:
> I'll come up with new tests today

Great! I went ahead and implemented a black-box test, as long as the solution. I also checked the wireshark capture, and I think it works as expected.

sfod:
OK, I've fixed ReaderProxyTests. I didn't add new tests, as it seems the existing ones cover the logic well.

MiguelCompany:
@sfod Thank you. As I have contributed with some code. I would like input from another colleague.

@richiware Would you mind reviewing this?

MiguelCompany:
@sfod We talked about this internally, and decided to go with the two-stages approach.

Would you mind updating the tests to consider that?

sfod:
@MiguelCompany I've fixed the tests according to the new logic.
May I ask what are the reasons that you chose the two-stage approach in the end?

MiguelCompany:
> I've fixed the tests according to the new logic

Thank you!

> May I ask what are the reasons that you chose the two-stage approach in the end?

We don't want a lossy reader to make non-lossy readers wait for new samples. That's why we handle retransmissions differently. If the writer has both a new sample and a retransmission pending, we give priority to the new samples.

MiguelCompany:
@richiware Would you mind reviewing this?

richiware (approved):

MiguelCompany:
@mergifyio backport 2.1.x 2.3.x 2.5.x

PR #2682, opened 2022-05-17
Refs 14322 Refactor SecurityManager synchronization <master>

MiguelBarro:
## Description

For sanitizer CI see https://github.com/eProsima/Fast-DDS/pull/2628

Refactor Guidelines:
+ What to do with const-use of members and references to the plugins?
  - the pointer members are initialized on `init()` which is called from participant constructor.
  - the pointer members are removed on `destroy()` which is called from ~SecurityManager() from the participant destructor. 
  - meanwhile `discovered_participant` and `on_process_handshake` are called from transport threads and leads to a myriad of member calls. The issue here is that destroy is called concurrently with those transport calls and leads to a race.
 
  The initial approach was to promote them to smart pointers able to guarantee unique getters and setters [see](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4058.pdf)
      but this 2014 proposal of an atomic<unique_ptr<T>> had no success. Though on 2017 the standard admitted
        `__cpp_lib_atomic_shared_ptr (c++20) -> atomic<shared_ptr<T>>`
   
   Finally a sentry based approach was favoured. The sentry requirements are:
  + It protects the plugins and other pointer based ancillary that doesn't require sync per-se but only `nullptr` checking.
  + It uses RAII strategy, thus it must be a method that returns a temporary object. While this object is *alive* the pointers cannot be modified.
  + The `SecurityManager::destroy()` method should wait till all the transport driven calls are done.
  Note that all methods call from `destroy()` must be sentry free to allow their execution.
  Note also that the `destroy()` must always called from a sentry free stack.

+ Collections must follow C++ const conventions. For example:
   ```c++
          std::map<GUID_t, DiscoveredParticipantInfo> discovered_participants_;
   ```
  in C++ philosophy this implies that the collection elements are like members and cannot be modified from `SecurityManager` const methods. This obviously doesn't fit in the actual use of them. Enconding/decoding operations should be const for the SecurityManager because it doesn't change its state but these members state. I'm going to promote them to:
   ```c++
          std::map<GUID_t, unique_ptr<DiscoveredParticipantInfo>> discovered_participants_;
   ```
+ Promote current mutex to `shared_mutex`. The write lock to protect changes on members (like collections update) and references (`new`/`delete`).
+ Collection members update must resort to their own sync methods. This was the premise before seeing how  
      `DiscoveredParticipantInfo` guarantees atomical update of its `AuthenticationInfo` members by using `unique_ptr<>` in
      the getters and setters (though the getters and setters must use another sync mechanism).  
+ Note that the previous mutex was recursive. Now care must be taken to avoid reentrancy.

After testing these changes it was detected that the sentry approach cannot prevent data removal due to discovery callbacks.
That is when a participant or endpoint was removed nothing prevent its encoding keys from being removed while other threads were using them to encode/decode a message.
In order to prevent this a new handle strategy was enforced:
+ Reference counting was added to the keys and other info required for encoding/decoding. This way even if the participant or endpoint info is removed only when the encode/decode operations are finished the actual removal would take place.
+ The handles cannot be created arbitrarily as before but class factories are assigned. This way the heap allocations and deallocations are prevented outside of the controlled factory code. Using this strategy the reference counting block uniqueness is assured.

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/.github/workflows/thread-sanitizer.yaml
+++ b/.github/workflows/thread-sanitizer.yaml
@@ -89,13 +89,10 @@ jobs:
           cd ws
           git clone https://github.com/eProsima/Fast-DDS.git
           cd Fast-DDS
-<<<<<<< HEAD
-=======
           git fetch origin ${{ github.ref }}:merge-branch
           git checkout merge-branch
           echo "Modified files against destination branch"
           git diff --name-only HEAD~1
->>>>>>> b14962a01... Fixed unknown ref error during checkout in workflow execution (#2644)
           git remote add efork https://github.com/MiguelBarro/Fast-DDS.git
           git fetch efork
           git merge --no-edit efork/bugfix/add_sanitizer_support | Out-Null
--- a/.github/workflows/thread-sanitizer.yaml
+++ b/.github/workflows/thread-sanitizer.yaml
@@ -21,7 +21,7 @@ jobs:
     name: Sanitizer Evaluation
     runs-on: ubuntu-latest
     steps:
-      - name: Download the sanitazers cmake module and apply some fixes
+      - name: Download the sanitizers cmake module and apply some fixes
         run: |
           git config --global user.email "dummy@mail.com"
           git config --global user.name "dummy"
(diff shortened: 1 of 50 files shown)


MiguelCompany:
@richiprosima Please test mac

MiguelCompany:

jsantiago-eProsima (approved):
LGTM

Issue #1906, opened 2021-04-09
SIGSEGV in removeRTPSReader on application exit [13602]


StefanFabian:
I have some very simple tests for a ros2 package I am working on.
This involves publishing and subscribing messages in the same process.
After all tests passed, the executable crashes with a segmentation fault when the node is destructed.

## System information

- Fast-RTPS version: 1.2.4
- OS: Ubuntu 20.04
- ROS2: Foxy

## Additional resources
- Stack Trace:
```
eprosima::fastrtps::rtps::RTPSDomain::removeRTPSReader(eprosima::fastrtps::rtps::RTPSReader*) 0x00007ffff6c9e8ff
eprosima::fastrtps::SubscriberImpl::~SubscriberImpl() 0x00007ffff6cc5b06
eprosima::fastrtps::SubscriberImpl::~SubscriberImpl() 0x00007ffff6cc5dcd
eprosima::fastrtps::ParticipantImpl::removeSubscriber(eprosima::fastrtps::Subscriber*) 0x00007ffff6cab4a4
eprosima::fastrtps::Domain::removeSubscriber(eprosima::fastrtps::Subscriber*) 0x00007ffff6ca216b
rmw_fastrtps_shared_cpp::destroy_subscription(char const*, CustomParticipantInfo*, rmw_subscription_t*) 0x00007ffff7239683
rmw_fastrtps_shared_cpp::__rmw_destroy_subscription(char const*, rmw_node_t const*, rmw_subscription_t*) 0x00007ffff7237052
rcl_subscription_fini 0x00007ffff7d4db53
<unknown> 0x00007ffff7f3f005
<unknown> 0x00007ffff7f3f7d5
rclcpp::SubscriptionBase::~SubscriptionBase() 0x00007ffff7f3dcb2
std::_Sp_counted_base<(__gnu_cxx::_Lock_policy)2>::_M_release shared_ptr_base.h:155
rclcpp::TimeSource::detachNode() 0x00007ffff7f443ed
rclcpp::TimeSource::~TimeSource() 0x00007ffff7f44531
rclcpp::node_interfaces::NodeTimeSource::~NodeTimeSource() 0x00007ffff7f02543
rclcpp::node_interfaces::NodeTimeSource::~NodeTimeSource() 0x00007ffff7f0298d
rclcpp::Node::~Node() 0x00007ffff7eda420
[...]

PR #2739, opened 2022-06-09
Move DomainParticipant deletion from Impl to delete_participant()

### LINKED ISSUE #937: Problems when creating/destructing participant in others static objects
In ros2, we're migrating from creating one `Participant` per `Node` to one per `Context` (see https://github.com/ros2/rmw_fastrtps/pull/312).
The [default context](https://github.com/ros2/rclcpp/blob/fd3655c26cc363c128402eaab7f563bdd17a3b7c/rclcpp/include/rclcpp/contexts/default_context.hpp#L39) has for example [static storage](https://github.com/ros2/rclcpp/blob/fd3655c26cc363c128402eaab7f563bdd17a3b7c/rclcpp/src/rclcpp/contexts/default_context.cpp#L25).
The destructor of that object, ends up calling: https://github.com/eProsima/Fast-RTPS/blob/a181a67b94e88839755e83a52ec72af4c9145ac6/include/fastrtps/Domain.h#L145
As the `Domain` object is a singleton, and destruction order of static objects in different transitional units can't be ensured, it sometimes crash (because it's used an already destructed `Domain`).

I didn't try, but the same bug can happen with https://github.com/eProsima/Fast-RTPS/blob/master/include/fastdds/dds/domain/DomainParticipantFactory.hpp.
And AFAIK, the same problem is common in other DDS vendors.

## Expected Behavior

Have a way to construct/destroy Participants in objects with static storage.

## Current Behavior

It's not possible.

## Steps to Reproduce

I don't have a simple repro ....
I can share detailed steps if needed.

## System information

- Fast-RTPS version: 1.9.x
- OS: Ubuntu 18.04
- Network interfaces: 
                      
- ROS2: Eloquent

## Additional context

The only solution that I have in mind, is to provide a way to the user to create a `Domain` (or `DomainParticipantFactory`). In that way, the object can be stored in another transitional unit and  destruction order can be ensured. The solution to the commented ROS 2 problem is still complicated, but possible if that's available.
We're also considering to stop having a default context, but I just want to rise awareness to the issue.

maloel:
Signed-off-by: Eran <librealsense.eran@gmail.com>

## Description
See #2537, #1906, etc.:

The singleton destructor of the `DomainParticipantFactory` is not the problem. Rather, the fact that it deletes all participants is. This PR makes it so it deletes their `impl_` field, setting it to `nullptr` thereby invalidating existing participants so they do not crash, for example when someone then tries to `delete_participant()`.

Fixes #937 

This was tested locally where we see the issue. In the above scenario, the Factory is destroyed before our object, which then sees that the participant is no longer `is_valid()` and does not try to call the factory to `delete_participant()`. This, of course, has the side-effect of leaving the participant alive on process exit (so is a "memory leak") but this is much better than the crash that currently occurs and that cannot be avoided if things need destruction on exit.

## Contributor Checklist
- [ ] Changes are ABI compatible. 

## Reviewer Checklist

### DIFF
--- a/include/fastdds/dds/domain/DomainParticipant.hpp
+++ b/include/fastdds/dds/domain/DomainParticipant.hpp
@@ -95,6 +95,13 @@ public:
      */
     virtual ~DomainParticipant();
 
+    /**
+     * This returns whether the participant is a valid object to be interacted with.
+     * 
+     * @return true if the participant is still valid, false otherwise
+     */
+    bool is_valid() const { return( nullptr != impl_ ); }
+
     // Superclass methods
 
     /**
--- a/src/cpp/fastdds/domain/DomainParticipant.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipant.cpp
@@ -46,28 +46,38 @@ DomainParticipant::~DomainParticipant()
 ReturnCode_t DomainParticipant::get_qos(
         DomainParticipantQos& qos) const
 {
+    if (!impl_)
+        return ReturnCode_t::RETCODE_ALREADY_DELETED;
     return impl_->get_qos(qos);
 }
 
 const DomainParticipantQos& DomainParticipant::get_qos() const
 {
+    if (!impl_)
+        throw std::runtime_error( "participant was deleted" );
     return impl_->get_qos();
 }
 
 ReturnCode_t DomainParticipant::set_qos(
         const DomainParticipantQos& qos) const
 {
+    if (!impl_)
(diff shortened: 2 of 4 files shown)


JLBuenoLopez-eProsima:
Hi @maloel,

I am going to close this PR because it is not going to be merged following [this comment](https://github.com/eProsima/Fast-DDS/discussions/2537#discussioncomment-2984869). Thanks anyway for your contribution!

PR #2833, opened 2022-07-12
[15184] Log race-condition bugfix

jparisu:

The idea was to solve a race condition in Fast DDS Log `Flush`.

## Description
After `Flush` it may happen that not all the logs has been consumed by consumers.

**BUT** creating a test to get this race condition, I have encountered another bug that makes a new Consumer get old logs.

## Contributor Checklist

## Reviewer Checklist


MiguelBarro:
This line here:
https://github.com/eProsima/Fast-DDS/blob/7d4e28a4cbd0d670393d27de8bacb8f3291d5294/src/cpp/fastdds/log/Log.cpp#L184
was not only a `pop()` it was actually a barrier to prevent the `Flush` operation from return without processing the last entry in the foreground queue. See:
https://github.com/eProsima/Fast-DDS/blob/7d4e28a4cbd0d670393d27de8bacb8f3291d5294/src/cpp/fastdds/log/Log.cpp#L139-L149

This can be seen analyzing the new `LogTests.flush_info` test:

1. The entry is logged in the background queue (if the compilation flags are right 😄).
2. `Flush` is performed and the first wait takes place.
   The log thread loops swapping the queues and RELEASING the lock before delivering to the consumers.
   The `FrontAndPop()` operation clears the foreground queue but the consumer didn't process the entry yet.  
   The former behavior was to `pop()` after consumer processing thus the foreground queue was not empty.
3. An spurious wakeup takes place in the main thread. Now `resources_.logs.BothEmpty() == true` and no wait is performed.
   This means `Log::Flush()` returns without the last entry being delivered.

This means that the last element on the queue may not be flushed. But again all the above requires AN SPURIOUS WAKEUP.
But according to the standard we cannot trust an spurious wakeup may not happen and strongly advises to use wait predicates always.

It would be great to fall back to the old behavior and add a comment in the `pop()` statement explaining his function as a barrier.
I've been trying to keep the `FrontAndPop()` in place but that means adding a devoted variable for the special case and makes things more cumbersome (though maybe somebody may find a workaround 😃).

jparisu:
> It would be great to fall back to the old behavior and add a comment in the `pop()` statement explaining his function as a barrier.
> I've been trying to keep the `FrontAndPop()` in place but that means adding a devoted variable for the special case and makes things more cumbersome (though maybe somebody may find a workaround smiley).

That is totally true.
Hoewever:
1. The test presented in this PR still fails, and has nothing to do (a priori) with this.
2. There is still a race condition, much less common, but still possible: *makes pop, flush finishes, enters a new log -> while continues and flush has finished*.
    In my opinion the solution would be avoid unlocking and locking again a mutex in a secured code. Inside DBQueue is still blocking, so it actually does not achieve anything with this unlock, apart from logic problems and race conditions.

MiguelBarro (approved):
LGTM
Especially my commits.

MiguelBarro:

jparisu (commented):
LGTM (Barro commits)

MiguelBarro:
@richiprosima Please let's test linux again

MiguelCompany (approved):

PR #2898, opened 2022-08-02
[15341] Protect access to reader listeners

MiguelCompany:
## Description

This should fix dataraces involving access to the listener on `RTPSReader` and `DataReaderImpl`.

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/fastdds/subscriber/DataReaderImpl.cpp
+++ b/src/cpp/fastdds/subscriber/DataReaderImpl.cpp
@@ -1639,13 +1639,16 @@ fastrtps::TopicAttributes DataReaderImpl::topic_attributes() const
 DataReaderListener* DataReaderImpl::get_listener_for(
         const StatusMask& status)
 {
-    std::lock_guard<std::mutex> _(listener_mutex_);
-
-    if (listener_ != nullptr &&
-            user_datareader_->get_status_mask().is_active(status))
     {
-        return listener_;
+        std::lock_guard<std::mutex> _(listener_mutex_);
+
+        if (listener_ != nullptr &&
+                user_datareader_->get_status_mask().is_active(status))
+        {
+            return listener_;
+        }
     }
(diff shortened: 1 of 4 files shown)


src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -1637,6 +1639,8 @@ fastrtps::TopicAttributes DataReaderImpl::topic_attributes() const
 DataReaderListener* DataReaderImpl::get_listener_for(
         const StatusMask& status)
 {
+    std::lock_guard<std::mutex> _(listener_mutex_);
```
jsantiago-eProsima, line 1642: Maybe it would be best to keep the final return clause outside the scope of this mutex.

jsantiago-eProsima (changes requested):
[This](https://github.com/eProsima/Fast-DDS/blob/c2a209e89c566f71293882fae6b0f9724334149b/src/cpp/rtps/reader/RTPSReader.cpp#L203) should be protected as well.

MiguelCompany:
> [This](https://github.com/eProsima/Fast-DDS/blob/c2a209e89c566f71293882fae6b0f9724334149b/src/cpp/rtps/reader/RTPSReader.cpp#L203) should be protected as well.

I have mixed feelings about this.
On one hand, I feel like protecting `RTPSReader::getListener()` should be done.
On the other hand, it is called all throughout the RTPS layer code, and the TSan job does not seem to be complaining. This means there are other protections in place whenever the method is called.

My fear is that taking the reader mutex may add ABBA deadlocks...

jsantiago-eProsima (approved):

jsantiago-eProsima:
I'm of the same mind regarding GetListener. Let's just keep it as is for now.

MiguelCompany:

PR #3040, opened 2022-10-26
Fix datarace on DomainParticipant disable [16040]

juanlofer-eprosima:
Signed-off-by: Juan López Fernández <juanlopez@eprosima.com>

## Description

Avoid dataraces on RTPS listener callbacks triggered while disabling a DomainParticipant (participant). The listener's pointer to participant is unset without protection in ``DomainParticipantImpl::disable``. Unsetting this reference is not required as it is enough to unset the (DomainParticipant)listener, which is thread safe. 

An alternative solution would be to unset the RTPS listener beforehand (before unsetting the DomainParticipant's listener). This solution may prevent other dataraces, but should be thoroughly analyzed.

Although in principle not required, DomainParticipant getters within RTPS listener callbacks are now also protected.

**Note** that for this case (`DomainParticipantImpl` disable before destruction when deleting a `DomainParticipant` via factory) it is possible that the `DomainParticipantListener` is set to `nullptr` (from `DomainParticipantImpl::disable`) in the middle of a RTPS listener callback, but no undefined behaviour should occur as the user should not be able to destroy the listener object at least until the delete operation is over, and this is a blocking operation waiting for the reception and event threads to join (thus assuring all callback operations are concluded by then).
However, this may not be the case when setting the `DomainParticipantListener` in other situations, see #3089 

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.cpp
@@ -159,7 +159,6 @@ void DomainParticipantImpl::disable()
     {
         participant->set_listener(nullptr);
     }
-    rtps_listener_.participant_ = nullptr;
 
     // The function to disable the DomainParticipantImpl is called from
     // DomainParticipantFactory::delete_participant() and DomainParticipantFactory destructor.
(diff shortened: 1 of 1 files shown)


EduPonz:

EduPonz:
Closing in favor of #3089

PR #3048, opened 2022-10-27
[16038] Fix SegFault in delete_contained_entities() when compiled with statistics

Mario-DL:
## Description

If compiled with statistics, when a participant calls delete_contained_entities() followed by a call to Factory::delete_participant(), a SegFault was produced.

Proposed solution changes:
Making ParticipantImpl delete_contained_entities() virtual so that the statistics participant can set the builtin_publishers to nullptr. 

@Mergifyio backport 2.7.x 2.6.x 2.3.x

## Contributor Checklist
- [x] Fast DDS test suite has been run locally. amd64 

## Reviewer Checklist

### DIFF
--- a/src/cpp/fastdds/domain/DomainParticipantImpl.hpp
+++ b/src/cpp/fastdds/domain/DomainParticipantImpl.hpp
@@ -332,7 +332,7 @@ public:
 
     DomainId_t get_domain_id() const;
 
-    ReturnCode_t delete_contained_entities();
+    virtual ReturnCode_t delete_contained_entities();
 
     ReturnCode_t assert_liveliness();
 
--- a/test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
+++ b/test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
@@ -698,6 +698,48 @@ TEST_F(StatisticsDomainParticipantTests, EnableStatisticsDataWriterFailureIncomp
 #endif // FASTDDS_STATISTICS
 }
 
+/**
+ * This test checks that a participant is correctly deleted by the factory
+ * after calling delete_contained_entities() method
+ * 1. Create a statistics participant, register type
+ * 2. Create a sample topic
+ * 3. Perform a delete_contained_entities() in the statistics participant
+ * 4. Delete the participant
+ */
+TEST_F(StatisticsDomainParticipantTests, DeleteParticipantAfterDeleteContainedEntitiesFailure)
+{
+#ifdef FASTDDS_STATISTICS
+
+    eprosima::fastdds::dds::TypeSupport count_type(new EntityCountPubSubType);
+
+    // 1. Create DomainParticipant with statistics
+    eprosima::fastdds::dds::DomainParticipant* participant =
+            eprosima::fastdds::dds::DomainParticipantFactory::get_instance()->
+                    create_participant(0, eprosima::fastdds::dds::PARTICIPANT_QOS_DEFAULT);
+    ASSERT_NE(participant, nullptr);
+
+    // Register type
+    participant->register_type(count_type);
+
+    // 2. Create a sample topic
+    participant->create_topic(HEARTBEAT_COUNT_TOPIC,
+                    count_type->getName(), eprosima::fastdds::dds::TOPIC_QOS_DEFAULT);
+
+    DomainParticipant* statistics_participant = DomainParticipant::narrow(participant);
+    ASSERT_NE(statistics_participant, nullptr);
+
+    EXPECT_EQ(ReturnCode_t::RETCODE_OK, statistics_participant->enable_statistics_datawriter(HEARTBEAT_COUNT_TOPIC,
+            STATISTICS_DATAWRITER_QOS));
+
+    // 3. Perform a delete_contained_entities() in the statistics participant
+    EXPECT_EQ(statistics_participant->delete_contained_entities(), ReturnCode_t::RETCODE_OK);
+
+    // 4. Delete the participant
+    EXPECT_EQ(eprosima::fastdds::dds::DomainParticipantFactory::get_instance()->
+                    delete_participant(participant), ReturnCode_t::RETCODE_OK);
+#endif // FASTDDS_STATISTICS
+}
+
 } // namespace dds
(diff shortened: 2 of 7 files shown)


src/cpp/fastdds/domain/DomainParticipantImpl.hpp
```diff
@@ -332,7 +332,7 @@ class DomainParticipantImpl
 
     DomainId_t get_domain_id() const;
 
-    ReturnCode_t delete_contained_entities();
+    virtual ReturnCode_t delete_contained_entities();
```
jparisu, line 335: Does this break ABI compatibility?

jparisu (changes requested):
NOTE: Notice that this feature is not tested in actual tests due to not compile with STATISTICS.
CI must be updated in order to check it, or add a test where in CMakeLists it is manually specified the option `FASTDDS_STATISTICS=ON`.

NOTE: Can statistics entities be deleted as normal entities? There is a specific method to delete them `eprosima::fastdds::statistics::dds::DomainParticipantImpl::delete_statistics_builtin_entities()` and ` eprosima::fastdds::statistics::dds::DomainParticipantImpl::disable_statistics_datawriter(const std::string &topic_name)`, maybe is not so trivial to remove them as any other entity.

ADDITION: Check in `core/src/fastdds/src/cpp/fastdds/domain/DomainParticipantFactory.cpp` (135) `eprosima::fastdds::dds::DomainParticipantFactory::delete_participant(eprosima::fastdds::dds::DomainParticipant *part)` that statistical entities are being deleted before checking if deletion could be done.

jparisu:
> - [x] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added.

This is not totally correct

src/cpp/fastdds/domain/DomainParticipantImpl.hpp
Mario-DL: Mmm not really since it only affects inner Impl (I think), is that correct @MiguelCompany ?

MiguelCompany: Correct. This only affects internal implementation, so no ABI break.

test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
```diff
@@ -698,6 +698,49 @@ TEST_F(StatisticsDomainParticipantTests, EnableStatisticsDataWriterFailureIncomp
...
+            count_type->getName(), eprosima::fastdds::dds::TOPIC_QOS_DEFAULT);
+
+    DomainParticipant* statistics_participant = DomainParticipant::narrow(participant);
+    ASSERT_NE(statistics_participant, nullptr);
+
+    EXPECT_EQ(ReturnCode_t::RETCODE_OK, statistics_participant->enable_statistics_datawriter(HEARTBEAT_COUNT_TOPIC,
+            STATISTICS_DATAWRITER_QOS));
+
+    // 3. Perform a delete_contained_entities() in the statistics participant
+    EXPECT_EQ(statistics_participant->delete_contained_entities(), ReturnCode_t::RETCODE_OK);
```
jparisu, line 736: I would call this method from `participant` and not from `statistics_participant`, as this would be the most expected way to do it from a user. (It should be the same, but just in case, to avoid virtualization method problems).

test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
```diff
@@ -698,6 +698,49 @@ TEST_F(StatisticsDomainParticipantTests, EnableStatisticsDataWriterFailureIncomp
...
+/**
+
+ * This test checks that a participant is correctly deleted by the factory
+ * after calling delete_contained_entities() method
+ * 1. Create a statistics participant, register type
+ * 2. Create a sample topic
+ * 3. Perform a delete_contained_entities() in the statistics participant
+ * 4. Delete the participant
+ */
+TEST_F(StatisticsDomainParticipantTests, DeleteParticipantAfterDeleteContainedEntitiesFailure)
```
jparisu, line 710: This test seems fine, but testing it without the fix does not give the expected result.
Expected result: it fails with a SegFault as the publisher of statistics have already been destroyed.
Actual result: it fails due to timeout as `delete_participant` get locked getting mutex `eprosima::fastdds::dds::DomainParticipantImpl::mtx_gs_`

test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
```diff
@@ -698,6 +698,49 @@ TEST_F(StatisticsDomainParticipantTests, EnableStatisticsDataWriterFailureIncomp
...
+    ASSERT_NE(participant, nullptr);
+
+    // Register type
+    participant->register_type(count_type);
+
+    // 2. Create a sample topic
+    participant->create_topic(HEARTBEAT_COUNT_TOPIC,
+            count_type->getName(), eprosima::fastdds::dds::TOPIC_QOS_DEFAULT);
+
+    DomainParticipant* statistics_participant = DomainParticipant::narrow(participant);
```
jparisu, line 729: NIT: I would not encapsulate all tests in `eprosima::fastdds::statistics::dds` as there are classes equally called in dds:: and statistics::dds:: , and could lead to error.

jparisu (changes requested):

test/unittest/statistics/dds/StatisticsDomainParticipantTests.cpp
Mario-DL: You are right, if you launch the test as it is, it gets blocked on that mutex. 
The thing is that the program in under undefined behaviour at this point (in fact, it is since the previous condition is skipped and enter the statement) because builtin_publisher_ is a dangling pointer.
```
if (nullptr != builtin_publisher_)
{... /*undefined behaviour*/...}
```
Try to launch the test with valgrind, the output will be the expected. We will discuss the best way to tackle this test.

rsanchez15:

jparisu (approved):
LGTM

rsanchez15:

JLBuenoLopez-eProsima (approved):
LGTM

Issue #3056, opened 2022-11-02
Failing to create DataReader leads to Segmentation Fault


fujitatomoya:
### Is there an already existing issue for this?

- [X] I have searched the existing issues

### Expected behavior

Successfully create the DataReader and will not receive Segmentation Fault signal.

### Current behavior

Failed to create the DataReader and ends up with Segmentation Fault.

### Steps to reproduce

W.I.P

### Fast DDS version/commit

- https://github.com/eProsima/Fast-DDS/tree/2.6.1 ***(We cannot reproduce the issue)***
- https://github.com/eProsima/Fast-DDS/tree/2.6.2
- https://github.com/eProsima/Fast-DDS/tree/2.8.0 + https://github.com/eProsima/Fast-DDS/pull/2976

### Platform/Architecture

Ubuntu Focal 20.04 arm64

### Transport layer

Default configuration, UDPv4 & SHM

### Relevant log output

```shell
2022-10-31T01:45:08.000047480Z [RTPS_EDP Error] Adding already existent reader 0.0.11.7 in topic elements/current -> Function operator()
2022-10-31T01:45:08.000070203Z [RTPS_EDP Error] Adding already existent reader 0.0.11.7 in topic elements/current -> Function operator()
2022-10-31T01:45:08.000074282Z [RTPS_EDP Warning] Failed register ReaderProxyData in EDP -> Function addLocalReader
2022-10-31T01:45:08.000077628Z [RTPS_EDP Warning] Failed register ReaderProxyData in EDP -> Function addLocalReader
2022-10-31T01:45:08.000080788Z [DATA_READER Error] Could not register reader on discovery protocols -> Function enable
2022-10-31T01:45:08.000083944Z [DATA_READER Error] Could not register reader on discovery protocols -> Function enable
2022-10-31T01:45:08.118559752Z [RTPS_LIVELINESS Warning] Reader not removed from WLP, unknown reader -> Function remove_local_reader
2022-10-31T01:45:08.118570303Z [RTPS_LIVELINESS Warning] Reader not removed from WLP, unknown reader -> Function remove_local_reader
```

jingTian-z:
The commit you provided is tag:v2.8.0, but the crash log shows that the library you are using is libfastrtps.so.2.6.0. Is it because of version mismatch?

fujitatomoya:
@jingTian-z good eye, let me check it, will get back to you.

fujitatomoya:
[...]

PR #3057, opened 2022-11-03
Check if not null before calling delete.

fujitatomoya:
general code improvement with null check before delete.

Signed-off-by: Tomoya.Fujita <Tomoya.Fujita@sony.com>

## Description

If these objects are not create as expected when it fails to create entity successfully, this will lead to segmentation fault.

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/fastdds/publisher/DataWriterImpl.cpp
+++ b/src/cpp/fastdds/publisher/DataWriterImpl.cpp
@@ -414,8 +414,14 @@ ReturnCode_t DataWriterImpl::check_delete_preconditions()
 
 DataWriterImpl::~DataWriterImpl()
 {
-    delete lifespan_timer_;
-    delete deadline_timer_;
+    if (lifespan_timer_) {
+        delete lifespan_timer_;
+        lifespan_timer_ = nullptr;
+    }
+    if (deadline_timer_) {
+        delete deadline_timer_;
+        deadline_timer_ = nullptr;
+    }
 
     if (writer_ != nullptr)
     {
--- a/src/cpp/fastdds/subscriber/DataReaderImpl.cpp
+++ b/src/cpp/fastdds/subscriber/DataReaderImpl.cpp
@@ -289,8 +289,14 @@ void DataReaderImpl::disable()
 
 void DataReaderImpl::stop()
 {
-    delete lifespan_timer_;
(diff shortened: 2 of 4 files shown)


fujitatomoya:
found this minor fix during https://github.com/eProsima/Fast-DDS/issues/3056

fujitatomoya:
friendly ping, we detected this core crash with this stack trace.

EduPonz:
Hi @fujitatomoya ,

I think we are OK with the change, but we are wondering why is it necessary. According to the [delete expression article in cppreference](https://en.cppreference.com/w/cpp/language/delete), it should be OK to call delete on `nullptr`:

* For the first (non-array) form, expression must be a pointer to an object type or a class type [contextually implicitly convertible](https://en.cppreference.com/w/cpp/language/implicit_conversion) to such pointer, and its value must be either null or pointer to a non-array object created by a [new-expression](https://en.cppreference.com/w/cpp/language/new), or a pointer to a base subobject of a non-array object created by a [new-expression](https://en.cppreference.com/w/cpp/language/new).

I guess that this could become a problem when redefining the behaviour of `delete` with some other implementation. Can that be what you are doing?

fujitatomoya:
> it should be OK to call delete on nullptr:

this is no problem, but i am suggesting to store the nullptr after delete here. calling `delete` on already deleted object will go undefined behavior and mostly crash. it does not make sense to store nullptr here? i am not sure if this sequence is considered. we just detected that stack trace that show delete the deleted object, so this is just a suggested fix.

if you do not want to `if` statement to check nullptr, that makes sense to me. and i can do that.

thanks,

MiguelCompany:
@fujitatomoya You are right. But then the root cause is that `DataWriterImpl` or `DataReaderImpl` are being destructed twice... We'd then need to find why that is happening, and add a fix for it along with the already proposed changes.

fujitatomoya:
> But then the root cause is that DataWriterImpl or DataReaderImpl are being destructed twice...

precisely correct !

> We'd then need to find why that is happening, and add a fix for it along with the already proposed changes.

yeah, that is no problem. just checking, no pressure.

fujitatomoya:
@MiguelCompany @EduPonz it is all up to you guys if taking this or not. if not, please go ahead to close this PR 👍

Issue #2969, opened 2022-09-22
Sequences and Arrays of Nested Structs Cause Segfault on Pub-Sub for Dynamic Types (with minimal examples)


methylDragon:
## ⚠️ This is a blocker for demoing a [ROS 2 REP](https://github.com/ros-infrastructure/rep/pull/358) reference implementation including FastDDS for ROSCon 2022 ⚠️
We just need a fix in-source (no need for a release yet) to do the demo. The trouble is, I think the issue is kinda inside the guts of FastDDS...

### Is there an already existing issue for this?

- [X] I have searched the existing issues

They exist for earlier versions of FastDDS (2.3.X), it seems the issue still persists in the current version.

### Expected behavior

1. Create a dynamic type consisting of a sequence/array of nested structs
2. Pass it to a `TypeSupport`
3. Register the `TypeSupport`

### Current behavior

1. Create a dynamic type consisting of a sequence/array of nested structs
2. Pass it to a `TypeSupport`
3. Register the `TypeSupport` <--- This segfaults

This issue occurs for empty sequences, sequences, and arrays.

### Steps to reproduce

I've prepared a minimal example:
https://github.com/methylDragon/fastdds-pr-minimal-examples

Once built, run:
```
cd build

./src/nested-sequences-bug/empty_sequence_of_nested_structs
./src/nested-sequences-bug/sequence_of_nested_structs
./src/nested-sequences-bug/array_of_nested_structs
```

The relevant code can be found [here](https://github.com/methylDragon/fastdds-pr-minimal-examples/tree/main/src/nested-sequences-bug/src).

The types can be constructed, dynamic data can be created, and dynamic data is printable, it's just the registration of the type support that breaks for some reason.

### Fast DDS version/commit

FastDDS 2.8.x: https://github.com/eProsima/Fast-DDS/commit/78473a0de3d878a71565ee5d6c8e1e01c34beb70

The examples workspace is using that specific SHA in the submodule, for convenience.

### Platform/Architecture

Other. Please specify in Additional context section.

### Transport layer

Default configuration, UDPv4 & SHM

### Additional context

This bug is hampering the progress of the dynamic types and type introspection reference implementation for ROS 2 as proposed in [REP 2011](https://github.com/ros-infrastructure/rep/pull/358), to be presented in ROSCon 2022. Fixing it will expedite the progress for completing the demo before the talk.

[...]

PR #3086, opened 2022-11-11
Fix sequence of nested structs.

clalancette:
It turns out that with the current XTypes code, attempting to create a sequence of nested structs ends up in a segfault. See https://github.com/eProsima/Fast-DDS/issues/2969 for the details.

After tracing through, it looks like what is happening is that there is some confusion on what should be registered as a "complete" type vs a "minimal" type, according to XTypes 1.3.  In particular, what is happening is that the nested structure is being created as a complete type, but then when we go to look it up, we look it up as both a complete type and a minimal type.  The former succeeds, and the latter fails, leading to the crash.

This attempted fix just skips trying to look it up as a minimal type at all.  This is enough to fix the crash.  This is probably not the correct fix, but it is unclear whether:

1.  We should be "building" the type as both a complete and minimal one at build() time, or
2.  We should only look it up as either a complete or minimal

I'm looking for feedback on which way to go here.

## Contributor Checklist
- [ ] Changes are ABI compatible. 

## Reviewer Checklist

### DIFF
--- a/src/cpp/dynamic-types/DynamicTypeBuilderFactory.cpp
+++ b/src/cpp/dynamic-types/DynamicTypeBuilderFactory.cpp
@@ -1188,18 +1188,24 @@ void DynamicTypeBuilderFactory::build_sequence_type_code(
         object.complete().sequence_type().element().common().element_flags().IS_KEY(false);
         object.complete().sequence_type().element().common().element_flags().IS_DEFAULT(false);
 
-        //TypeIdentifier ident;
-        //build_type_identifier(descriptor->get_base_type()->descriptor_, ident);
         TypeObject obj;
         build_type_object(descriptor->get_element_type(), obj, complete);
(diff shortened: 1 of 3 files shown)


JLBuenoLopez-eProsima:
Thanks for your contribution, @clalancette!

Right now efforts to fix and improve the XTypes API are being made by @MiguelBarro. @MiguelBarro, can you confirm that this issue is going to be fixed with the Xtypes rework?

JLBuenoLopez-eProsima:
@clalancette this issue is fixed in the XTypes v1.3 rework that is being done in [feature/xtypes1.3](https://github.com/eProsima/Fast-DDS/tree/feature/xtypes1.3) that will be released in the coming months. I will proceed and close this PR as this is fixed within that PR and tested using the IDLs contained in the [DDS types test repository](https://github.com/eProsima/dds-types-test). This specific case is tested [here](https://github.com/eProsima/dds-types-test/blob/main/IDL/sequences.idl#L125).

PR #3096, opened 2022-11-17
Fix bug in Topic creation with differfent Type Name [16242]

jparisu:
## Description

When registering a Type and when creating a Topic, a type name is requested.
However, this type name is not used in Topic creation because it uses internally the type name of the type support, that can actually not be the same as the type name that was registered.

Simple regression test has been uploaded in commit a393cf71424e57a076ccab8931703b1afdac022a that fails before the fix, and passes afterwards.

## Contributor Checklist
- NA New feature has been added to the `versions.md` file (if applicable).
- NA New feature has been documented/Current behavior is correctly described in the documentation. 

## Reviewer Checklist

### DIFF
--- a/test/unittest/dds/publisher/PublisherTests.cpp
+++ b/test/unittest/dds/publisher/PublisherTests.cpp
@@ -493,32 +493,6 @@ TEST(PublisherTests, CreateDataWriter)
     ASSERT_EQ(DomainParticipantFactory::get_instance()->delete_participant(participant), ReturnCode_t::RETCODE_OK);
 }
 
-TEST(PublisherTests, CreateDataWriterWithDifferentTypeName)
-{
-    DomainParticipant* participant =
-            DomainParticipantFactory::get_instance()->create_participant(0, PARTICIPANT_QOS_DEFAULT);
-    ASSERT_NE(participant, nullptr);
-
-    Publisher* publisher = participant->create_publisher(PUBLISHER_QOS_DEFAULT);
-    ASSERT_NE(publisher, nullptr);
-
-    std::string type_name = "other_different_type_name_because_of_reasons_eg_mangling";
-    TypeSupport type(new TopicDataTypeMock());
-    type.register_type(participant, type_name);
-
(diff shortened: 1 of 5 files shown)


test/unittest/dds/publisher/PublisherTests.cpp
```diff
@@ -493,6 +493,31 @@ TEST(PublisherTests, CreateDataWriter)
...
+    ASSERT_NE(topic, nullptr);
+
+    DataWriter* datawriter = publisher->create_datawriter(topic, DATAWRITER_QOS_DEFAULT);
+    ASSERT_NE(datawriter, nullptr);
+
+    ASSERT_EQ(publisher->delete_datawriter(datawriter), ReturnCode_t::RETCODE_OK);
+    ASSERT_EQ(participant->delete_publisher(publisher), ReturnCode_t::RETCODE_OK);
+    ASSERT_EQ(participant->delete_topic(topic), ReturnCode_t::RETCODE_OK);
+    ASSERT_EQ(DomainParticipantFactory::get_instance()->delete_participant(participant), ReturnCode_t::RETCODE_OK);
+}
```
MiguelCompany, line 519: The failure this PR is trying to fix is that, after creating a topic, calling `topic->get_type_name()` returns the incorrect value. So I think the correct test would be something like

```
    Topic* topic = participant->create_topic("footopic", type_name, TOPIC_QOS_DEFAULT);
    ASSERT_NE(topic, nullptr);
    ASSERT_EQ(type_name, topic->get_type_name());
```

I would just extend [this test](https://github.com/eProsima/Fast-DDS/blob/f0888e2d90185c7991488ece6de2aa87e6f7ceca/test/unittest/dds/participant/ParticipantTests.cpp#L1976) with some negative cases.

MiguelCompany (changes requested):
Changes LGTM, except for the test, which I think belongs in ParticipantTests

MiguelCompany (approved):
LGTM with green CI

MiguelCompany:
@richiprosima Please test mac

MiguelCompany:
@richiprosima Please test mac

PR #3162, opened 2022-12-19
[16500] PDP endpoints refactor

MiguelCompany:
## Description

We need to make the PDP classes agnostic of their endpoints, as we will create more than a single writer and reader on the secure case.

This PR models that with a PDPEndpoints base class.

## Contributor Checklist
    - Just checking that the current tests pass should be enough

## Reviewer Checklist


src/cpp/rtps/builtin/discovery/participant/PDPServer.cpp
```diff
@@ -402,6 +424,7 @@ void PDPServer::assignRemoteEndpoints(
 {
     EPROSIMA_LOG_INFO(RTPS_PDP_SERVER, "Assigning remote endpoint for RTPSParticipant: " << pdata->m_guid.guidPrefix);
 
+    auto endpoints = static_cast<fastdds::rtps::DiscoveryServerPDPEndpoints*>(builtin_endpoints_.get());
```
Mario-DL, line 427: Except for a couple of `dynamic_cast<>` operations which are needed upon stateful writer/readers creation (a `RTPSReader/Writer*` is returned), perhaps some of the `static_cast<DiscoveryServerPDPEndpoints> `calls could be saved up in this file, since it is costy and I find it redundant (I dont see now any external operation that can potentially change the type except for those above) neither the `builtin_endpoints_` member is exposed to be externally modified.

Mario-DL (commented):
Nice PDPEndpoints abstraction reliance. Improves code readability and prepares it for easy extension

src/cpp/rtps/builtin/discovery/participant/PDPServer.cpp
MiguelCompany: > It can be that .get() from unique_ptr returns the raw pointer to the base class, so the cast is always needed

It is exactly that

Mario-DL (approved):
LGTM

PR #3175, opened 2022-12-23
[16603] Secure PDP exchange

MiguelCompany:
## Contributor Checklist
- **N/A** Fast DDS test suite has been run locally. 

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/builtin/discovery/endpoint/EDPClient.cpp
+++ b/src/cpp/rtps/builtin/discovery/endpoint/EDPClient.cpp
@@ -46,7 +46,8 @@ bool EDPClient::processLocalReaderProxyData(
 
     auto* writer = &subscriptions_writer_;
 
-#if HAVE_SECURITY
+// TODO(Miguel C): Re-enable this when EDPServer creates the secure EDP endpoints
+#if 0 // HAVE_SECURITY
     if (local_reader->getAttributes().security_attributes().is_discovery_protected)
     {
         writer = &subscriptions_secure_writer_;
@@ -79,7 +80,8 @@ bool EDPClient::processLocalWriterProxyData(
 
     auto* writer = &publications_writer_;
 
-#if HAVE_SECURITY
+    // TODO(Miguel C): Re-enable this when EDPServer creates the secure EDP endpoints
+#if 0 //HAVE_SECURITY
     if (local_writer->getAttributes().security_attributes().is_discovery_protected)
     {
         writer = &publications_secure_writer_;
@@ -110,7 +112,8 @@ bool EDPClient::removeLocalWriter(
 
     auto* writer = &publications_writer_;
 
-#if HAVE_SECURITY
+    // TODO(Miguel C): Re-enable this when EDPServer creates the secure EDP endpoints
+#if 0 // HAVE_SECURITY
     if (W->getAttributes().security_attributes().is_discovery_protected)
     {
         writer = &publications_secure_writer_;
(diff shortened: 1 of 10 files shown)


MiguelCompany:

src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
```diff
@@ -472,11 +472,73 @@ void PDPClient::assignRemoteEndpoints(
...
+                match_pdp_writer_nts_(svr, pdata.m_guid.guidPrefix);
+                break;
+            }
+        }
+    }
+#endif // HAVE_SECURITY
+
+    perform_builtin_endpoints_matching(pdata);
+}
+
```
Mario-DL, line 506: I dont know if this should be within an `else {}` statement. If HAVE_SECURITY is true and `should_protect_discovery()` returns true, perhaps it could be called twice, one here and again in assignRemoteEndpoints().

```suggestion
        }
    }
    else
#endif // HAVE_SECURITY
    {
        perform_builtin_endpoints_matching(pdata);
    }
}

```

Mario-DL (commented):
Nice manual matching of `reliable_pdp_endpoints() `upon notifyAboveEndpoints() if compiled WITH_SECURITY and secure participants are used !
1) Ensure security keys secret exchange is ok in SecurityManager before matching.
2) specialize `pair_..._after_security()` for PDPClient and PDPServer and call them from SecurityManager if previous step was ok.
3) call the corresponding `matched_w/r_add()` for each builtin_endpoint.

src/cpp/rtps/builtin/discovery/endpoint/EDPClient.cpp
```diff
@@ -46,7 +46,8 @@ bool EDPClient::processLocalReaderProxyData(
 
     auto* writer = &subscriptions_writer_;
 
-#if HAVE_SECURITY
+// TODO(Miguel C): Re-enable this when EDPServer creates the secure EDP endpoints
```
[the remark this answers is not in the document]
MiguelCompany: I don't think we should please the linter on a commit that is meant to be reversed

src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
MiguelCompany: Nice catch. But I think the correct solution would be to remove the call to `perform_builtin_endpoints_matching` from here.

src/cpp/rtps/builtin/discovery/participant/PDPServer.cpp
```diff
@@ -571,68 +571,143 @@ void PDPServer::initializeParticipantProxyData(
...
+    {
+        perform_builtin_endpoints_matching(*pdata);
+    }
 }
 
 void PDPServer::notifyAboveRemoteEndpoints(
         const ParticipantProxyData& pdata)
+{
+#if HAVE_SECURITY
+    match_reliable_pdp_endpoints(pdata);
```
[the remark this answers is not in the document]
MiguelCompany: Methods `match_pdp_reader_nts_` and `match_pdp_writer_nts_` are meant for server to server matching. Method `match_reliable_pdp_endpoints` is meant for any participant with which the server should communicate. Different purpose, different methods.

I agree that it would be nice for the code of PDPClient and PDPServer to be as similar as possible, but I don't think it is possible in this case.

src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
Mario-DL: Okey. Although having a second glance at it perhaps it is not a bad idea to leave it as it is, so we ensure we call it. 
The `perform_builtin_endpoints_matching()` call within `assignRemoteEndpoints()` is not always performed in every callback invokation because the `if(mp_RTPSParticipant->security_manager().discovered_participant(*pdata))` can return false (despite participant already exists) due to the fact that AuthInfoUniquePtr in the discovered_participants_ collection can be busy for that participant (other thread may have taken the ownership as the pointer is being moved from the collection in a lot of methods).

BTW, I have the mangled guid PR ready, will upload later after local testing.

PR #3208, opened 2023-01-12
Update ubuntu version in ASAN tests [17250]

jparisu:
This PR only affects GitHub Action ASAN tests

## Contributor Checklist
- [N/A] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added. 
- [N/A] Any new/modified methods have been properly documented using Doxygen. 
- [N/A] Fast DDS test suite has been run locally. 
- [N/A] Changes are ABI compatible. 
- [N/A] Changes are API compatible. 
- [N/A] Documentation builds and tests pass locally. 
- [N/A] New feature has been added to the `versions.md` file (if applicable).
- [N/A] New feature has been documented/Current behavior is correctly described in the documentation. 
- [N/A] Applicable backports have been included in the description.

## Reviewer Checklist


.github/workflows/sanitizer-tests.yaml
```diff
@@ -101,10 +108,17 @@ jobs:
           git clone https://github.com/google/googletest.git --branch release-1.10.0
           cd ..
 
+      - name: Build Gtest (without -Wall)
```
jparisu, line 111: Reuse Fetch repositories already existing action and add argument to add or not discovery server

.github/workflows/sanitizer-tests.yaml
```diff
@@ -36,10 +36,17 @@ jobs:
       - name: Fetch Fast DDS dependencies
         uses: ./src/Fast-DDS/.github/actions/fetch-fastdds-repos
 
+      - name: Build Gtest (without -Wall)
```
jparisu, line 39: Remove GTest step

.github/workflows/utils/log_parser.py
```diff
@@ -229,4 +230,4 @@ def main():
 
 
 if __name__ == '__main__':
-    main()
+    exit(1 if main() else 0)
```
jparisu, line 233: could return n_errors instead of 1

rsanchez15: Usually every return code means something. That's why I changed it so it can be `1`, there are errors, or `0`, everything okay.

.github/workflows/sanitizer-tests.yaml
jparisu: This has not been done

rsanchez15:
Merge after #3209

.github/workflows/sanitizer-tests.yaml
```diff
@@ -32,14 +32,14 @@ jobs:
       - name: Install apt packages
         uses: ./src/Fast-DDS/.github/actions/install-apt-packages
 
-      - name: Install GTest
-        uses: ./src/Fast-DDS/.github/actions/install-gtest-linux
```
EduPonz, line 36: It seems we are not using this action anymore, may as well delete it

.github/workflows/sanitizer-tests.yaml
```diff
@@ -90,23 +90,18 @@ jobs:
...
-          git clone https://github.com/eProsima/foonathan_memory_vendor.git
-          git clone https://github.com/eProsima/Fast-CDR.git
-          git clone https://github.com/eProsima/Discovery-Server.git --branch ${{ github.event.inputs.discovery_server_branch || env.DEFAULT_DISCOVERY_SERVER_BRANCH }}
-          git clone https://github.com/google/googletest.git --branch release-1.10.0
-          cd ..
+        with:
+          target_directory: src
+          googletest: true
+          discovery_server: true
+          discovery_server_branch: ${{ github.event.inputs.discovery_server_branch || env.DEFAULT_DISCOVERY_SERVER_BRANCH }}
```
EduPonz, line 104: I see we use the specified version in case the workflow is manually dispatched, but if it's not, then I don't think we should always use master of Discovery Server. I think we should:

1. Use a branch with the same name as the PR target if exists; else
2. Use a branch with the same name as the PR base if exists; else
3. Use master

.github/workflows/utils/log_parser.py
```diff
@@ -130,14 +130,14 @@ def _common_line_splitter(
...
-    return line[(start_index + len(text_to_split_start)):].strip()
+    return line[(start_index + len(text_to_split_start)):].split(None, 1)[0]
 
 
 def asan_line_splitter(
         line: str):
     return _common_line_splitter(
         line=line,
-        text_to_split_start='==ERROR: ')
+        text_to_split_start='==ERROR: AddressSanitizer: ')
```
EduPonz, line 140: This leaves out some errors kinds like `LeakSanitizer` (see [here](https://github.com/eProsima/Fast-DDS/actions/runs/4199782572))

EduPonz (changes requested):

.github/workflows/utils/log_parser.py
rsanchez15: Addressed in b601c28c1

rsanchez15: Addressed in b601c28

.github/workflows/sanitizer-tests.yaml
rsanchez15: Addressed in 1563fc7774d965e0f6f8591e8f156c9554448e7e

.github/workflows/sanitizer-tests.yaml
```diff
@@ -101,7 +101,7 @@ jobs:
           target_directory: src
           googletest: true
           discovery_server: true
-          discovery_server_branch: ${{ github.event.inputs.discovery_server_branch || env.DEFAULT_DISCOVERY_SERVER_BRANCH }}
+          discovery_server_branch: ${{ github.base_ref || github.head_ref || github.event.inputs.discovery_server_branch || env.DEFAULT_DISCOVERY_SERVER_BRANCH }}
```
EduPonz, line 104: I think we should prioritize source branch (head) over target branch (base). Furthermore, as far as I can tell, when manually run, the specified branch does not have priority over neither head or base. I don't know if we can detect that a workflow has been manually run and the branch as been explicitly set. Else, we could make the option mandatory in that case, thus avoiding the default value, and always prioritize the option.

```suggestion
          discovery_server_branch: ${{ github.head_ref || github.base_ref || github.event.inputs.discovery_server_branch || env.DEFAULT_DISCOVERY_SERVER_BRANCH }}
```

EduPonz (changes requested):

.github/workflows/sanitizer-tests.yaml
rsanchez15: `github.base_ref` and `github.head_ref` only exists if the workflow is running from a PR. The same happens with `github.event.inputs.discovery_server_branch` but for workflow dispatch. Knowing that, we can guarantee the the workflow is running from a PR, a workflow dispatch or another event (push on main branch). In any case, I've just seen that according to this, the `github.base_ref` will never be checked so we need to find a way to pass both to the action.

rsanchez15: Addressed in 58c7943

EduPonz:
Your changes LGTM, but I will not approve the PR until we actually fix the leak errors. We'll assign some resources to it next week.

EduPonz:
We've agreed to merge this PR so that we do not include more leaks accidentally.

EduPonz (approved):

PR #3294, opened 2023-02-13
[17099] Fix chain of trust issues

### LINKED ISSUE #3239: [16903] Chain of trust issues with a single CA certificate

### Expected behavior

I'll quote the same succinctly put issue filled over for OpenDDS here:

https://github.com/OpenDDS/OpenDDS/pull/3992#issue-1548228907

> ## Problem
> Assume a usage of DDS Security where the same CA is used for both permissions and identity. The certificates issued to particpants allow them to sign documents. Assume the participant generates a permissions file and then signs it. Chain verification causes verification attempts to succeed since the signing certificate, i.e., the participant's certificate, can be chained back to the permission CA's certificate (which is also the identity CA).
> 
> This problem was identified in [ros2/sros2#282](https://github.com/ros2/sros2/issues/282).
> 
> ## Solution
> Implement the suggestion in [ros2/sros2#282](https://github.com/ros2/sros2/issues/282). Specifically, use PKCS7_NOINTERN to not accept any signatures in the signed document. This, in turn, requires the use of the `certs` parameter to `PKCS7_verify`. PKCS7_NOVERIFY is used since the permissions CA certificate will not be chain verified.

### Current behavior

https://github.com/ros2/sros2/issues/282#issue-1494316312

### Steps to reproduce

https://github.com/ros2/sros2/issues/282#issuecomment-1348663745

### Fast DDS version/commit

Example call site for reference:

https://github.com/eProsima/Fast-DDS/blob/5076ebc0c5d030cac6225b94e18ef5b17c996ef3/src/cpp/security/accesscontrol/Permissions.cpp#L518

### Platform/Architecture

Other. Please specify in Additional context section.

### Transport layer

### Additional context

Fast-DDS v2.6.2 ships with ROS 2 Humble

### XML configuration file

_No response_

### Relevant log output

_No response_

### Network traffic capture

_No response_

MiguelCompany:
## Description

This should fix #3239 using the proposed solution.
A regression test has been added following the misbehavior described.

@Mergifyio backport 2.9.x 2.8.x 2.6.x 2.1.x

## Contributor Checklist

## Reviewer Checklist


MiguelCompany:
@richiprosima Please test this.

src/cpp/security/accesscontrol/Permissions.cpp
```diff
@@ -346,6 +346,46 @@ static bool rfc2253_string_compare(
...
+    BIO* out = nullptr;
+
+    assert(nullptr != store);
+    assert(nullptr != in);
+
+    BIO* indata = nullptr;
+    PKCS7* p7 = SMIME_read_PKCS7(in, &indata);
+    if (nullptr == p7)
+    {
+        exception = _SecurityException_("Cannot read as PKCS7 the file ");
```
EduPonz, line 364: This log becomes less complete as it is missing the actual file name in the message, but since another more verbose error pops right after when returning `nullptr`, I'd just rephrase this a bit so it does not read as a hanging sentence

src/cpp/security/accesscontrol/Permissions.cpp
```diff
@@ -346,6 +346,46 @@ static bool rfc2253_string_compare(
...
+        return nullptr;
```
EduPonz, line 365: Don't we need to free `indata` before returning here??

EduPonz (changes requested):

src/cpp/security/accesscontrol/Permissions.cpp
MiguelCompany: I don't think so, since it is an output argument. See [the docs](https://www.openssl.org/docs/man1.0.2/man3/SMIME_read_PKCS7.html) for more info.

EduPonz (approved):

MiguelCompany:
https://github.com/Mergifyio backport 2.9.x 2.8.x 2.6.x 2.1.x

PR #3309, opened 2023-02-16
[17004] Fix performance tests - Overflow in received samples

MRicoIE2CS:
## Description

Samples sent and throughput value numbers suffer overflow.

## How to reproduce:

Run colcon test for any of the following tests:
- interprocess_reliable_shm
- interprocess_reliable_udp

For payloads;demand:
- 16;100
- 8388608;100

And recoveries:
- 0;10;20;30;40;50;60;70;80;90;100;

## Fix performed

- Fix overflow by counting received sampler directly instead of doing the following calculation, which is how it was done: counting lost samples comparing the sequence number of each received sample and the sequence number of the previous one, and then calculating received samples from last sequence number minus the count of lost samples.
- Then, lost samples calculation is also done now from the calculation of sent samples minus the received samples, instead of the mentioned calculation in receiver.

## Contributor Checklist
- [N/A] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- [N/A] Any new/modified methods have been properly documented using Doxygen. 
- [N/A] New feature has been added to the `versions.md` file (if applicable).
- [N/A] New feature has been documented/Current behavior is correctly described in the documentation. 

- [N/A] Applicable backports have been included in the description.

## Reviewer Checklist

### DIFF
--- a/test/performance/throughput/ThroughputSubscriber.cpp
+++ b/test/performance/throughput/ThroughputSubscriber.cpp
@@ -127,7 +127,7 @@ void ThroughputSubscriber::DataReaderListener::on_data_available(
             lost_samples_ += last_seq_num - last_seq_num_ - size;
         }
         last_seq_num_ = last_seq_num;
-        //received_samples_ += 1;
+        received_samples_ += 1;
 
         // release the reader loan
         if (ReturnCode_t::RETCODE_OK != reader->return_loan(data_seq, infos))
--- a/test/performance/throughput/ThroughputSubscriber.cpp
+++ b/test/performance/throughput/ThroughputSubscriber.cpp
@@ -780,11 +780,7 @@ void ThroughputSubscriber::run()
             std::cout << "Last Received Sample: " << command_sample.m_lastrecsample << std::endl;
             std::cout << "Lost Samples: " << command_sample.m_lostsamples << std::endl;
             std::cout << "Received Samples: " << command_sample.m_receivedsamples << std::endl;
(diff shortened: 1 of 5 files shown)


test/performance/throughput/ThroughputSubscriber.cpp
```diff
@@ -126,6 +127,7 @@ void ThroughputSubscriber::DataReaderListener::on_data_available(
             lost_samples_ += last_seq_num - last_seq_num_ - size;
         }
         last_seq_num_ = last_seq_num;
+        received_samples_ += 1;
```
jsantiago-eProsima, line 130:
```suggestion
        received_samples_ += size;
```

jsantiago-eProsima (changes requested):
Partial review pending thorough testing of the branch

test/performance/throughput/ThroughputSubscriber.cpp
MRicoIE2CS: Fixed. 
I applied another change, because there is a possibility that a received element is not valid, and that counts as a loss, so `size` variable should not be used here.

jsantiago-eProsima (approved):
LGTM

jsantiago-eProsima:

jsantiago-eProsima:
@richiprosima please test linux

jsantiago-eProsima:
@richiprosima please test mac

PR #3389, opened 2023-03-17
[17671] Add on_incompatible_type to RTPS listeners API

MiguelCompany:
## Description

As part of the implementation of on_inconsistent_topic, we need to add a callback to the RTPS listeners to notify of an incompatible type.

Marked as no-test since new functionality is not added.

## Contributor Checklist
- **NO** Changes are ABI compatible. 
- **N/A** New feature has been documented/Current behavior is correctly described in the documentation.  Not applicable since we don't document RTPS layer other than in the API reference
    
## Reviewer Checklist


include/fastdds/rtps/reader/ReaderListener.h
```diff
@@ -198,6 +198,18 @@ class RTPS_DllAPI ReaderListener
         should_notify_individual_changes = true;
     }
 
+    /**
+     * This method is called when a new Writer is discovered, with a Topic that
+     * matches the name of a local reader, but with an incompatible type
+     *
+     * @param writer Pointer to the RTPSReader.
+     */
+    virtual void on_incompatible_type(
```
Mario-DL, line 207: Perhaps I miss some info from the incompatible writer. I don't know if it could be useful to add some extra parameter, for example the `writer's_guid`.

Same suggestion in the case of the writer.

Mario-DL (changes requested):
Just a NIT and a suggestion

include/fastdds/rtps/reader/ReaderListener.h
MiguelCompany: We are not doing it for `on_requested_incompatible_qos`, which I consider to be similar to this new method. What do you think @EduPonz @JLBuenoLopez-eProsima ?

JLBuenoLopez-eProsima: I do not think this is really necessary. The `on_requested_incompatible_qos` DDS callback is defined in the specification and it does not include the remote endpoint information. I understand that the reasoning behind this issue is that you can only make changes locally, and knowing which remote endpoint has an inconsistent QoS or an incompatible type is not meaningful (other than with debugging purposes). But these callbacks are intended to be used in the user application logic, not for debugging (other tools are available for that purposes, like sniffing the network traffic that would provide the user with the Writer GUID).

EduPonz: Since we're adding this API for the implementation of DDS' `on_inconsistent_topic`, which gives access to the `Topic` and the `InconsistentTopicStatus`, that is constituted by a simple `total_count` & `total_count_change`, we definitely do not need it. I could however see it returned, but I'm inclined to keep the API consistent with the other similar callbacks in `ReaderListener` and just leave it as it is

Mario-DL (approved):
LGTM

Mario-DL:

MiguelCompany:
@richiprosima Please test mac

PR #3745, opened 2023-07-24
[19256] Remove mutex from TimedEventImpl

cvilas:
Eliminates the overhead of mutex locking in a method in TimeEventImpl that is frequently called. 

## Description

While investigating scalability of discovery service on a large distributed system, hotspot analysis using Intel vtune showed that most time is spent with mutexes. Further analysis identifies a particular culprit whose removal increases the performance by 2x.

The following is from a machine with Haswell CPU running a SERVER participant. Evidently, 50% of the compute is spent locking and unlocking a mutex. The CPU consumption measured by top is of the order of 37%.

hotspot

The performance analysis with vtune allows the construction of a so-called flamegraph that makes it very easy to understand the caller-callee relationship. The graph makes it very clear that the vast majority of the mutex locks and unlocks can be attributed to the function `eProsima::fastrtps::rtps::event_compare`.

flame_graph

Looking at the source code we can see that the mutex is locked and unlocked during a sort operation:

```
// src/cpp/rtps/resources/TimedEventImpl.h
std::chrono::steady_clock::time_point next_trigger_time()
{   
    std::unique_lock<std::mutex> lock(mutex_);
    return next_trigger_time_;
}

// src/cpp/rtps/resources/ResourceEvent.cpp
static bool event_compare(
        TimedEventImpl* lhs,
        TimedEventImpl* rhs)
{
    return lhs->next_trigger_time() < rhs->next_trigger_time();
}

void ResourceEvent::sort_timers()
{
    std::sort(active_timers_.begin(), active_timers_.end(), event_compare);
}
```

The mutex protects a couple of data members that can be wrapped in `std::atomic` instead. After eliminating the mutex, the flame graph becomes

flame_graph_after

At least in our use case, we are able to see an improvement in how well the discovery service scales without the lock.

## Contributor Checklist
    
## Reviewer Checklist

@Mergifyio backport 2.11.x 2.10.x 2.6.x

### DIFF
--- a/src/cpp/rtps/resources/TimedEventImpl.cpp
+++ b/src/cpp/rtps/resources/TimedEventImpl.cpp
@@ -22,9 +22,6 @@
 
 #include <fastrtps/utils/TimeConversion.h>
 
-#include <atomic>
-#include <functional>
-
 namespace eprosima {
 namespace fastrtps {
 namespace rtps {
@@ -73,12 +70,10 @@ bool TimedEventImpl::update(
 
     if (set_time)
     {
-        std::lock_guard<std::mutex> lock(mutex_);
-        next_trigger_time_ = current_time + interval_microsec_;
+        next_trigger_time_ = current_time + interval_microsec_.load();
     }
     else if (expected == StateCode::INACTIVE)
     {
-        std::lock_guard<std::mutex> lock(mutex_);
         next_trigger_time_ = cancel_time;
     }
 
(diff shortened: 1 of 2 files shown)


MiguelCompany:
@richiprosima Please test mac

cvilas:
> Thank you @cvilas for such a well-crafted description.

Thanks @MiguelCompany. Credit goes to https://github.com/iamholger for identifying the issue and proposing a potential fix. I am just the messenger, as he is away from work at the moment.

MiguelCompany:
@Mergifyio backport 2.11.x 2.10.x 2.6.x

MiguelCompany:
Thanks to @iamholger, then!

PR #4220, opened 2024-01-09
[19696][20409] Fix data race on PDP

Tempate:
## Description

This PR intends to solve a race condition on PDP discovery phase, in which a data UP is received and processed while its corresponding previously received data P is still being processed.

As a result, the participant proxy created and registered can be cleared before calling assignRemoteEndpoints, as the PDP mutex is released just before (on purpose to avoid interlocking problems).

This issue has been reported by TSAN in this particular scenario, when the data P is received in the UDP multicast thread and then the data UP in the unicast one. However, this problem should also affect other scenarios such as:

- Same but with SHM transport.
- Same but receiving data P in unicast thread and data UP in multicast one.
- Losing remote participant's liveliness (events thread) while still processing its data P.

The proposed solution is to copy the received participant proxy data before releasing the PDP mutex, and then passing this copy forward to the methods requiring this information.

@Mergifyio backport 2.12.x 2.10.x 2.6.x

## Contributor Checklist
    
## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
+++ b/src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
@@ -489,7 +489,7 @@ void PDPClient::assignRemoteEndpoints(
                 if (data_matches_with_prefix(svr.guidPrefix, *pdata))
                 {
                     std::unique_lock<std::recursive_mutex> lock(*getMutex());
-                    svr.proxy = pdata;
+                    svr.proxy = get_participant_proxy_data(pdata->m_guid.guidPrefix);
                 }
             }
         }
--- a/src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
+++ b/src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
@@ -472,7 +472,6 @@ bool PDPClient::create_ds_pdp_reliable_endpoints(
     return true;
 }
 
-// the ParticipantProxyData* pdata must be the one kept in PDP database
 void PDPClient::assignRemoteEndpoints(
         ParticipantProxyData* pdata)
 {
@@ -488,8 +487,7 @@ void PDPClient::assignRemoteEndpoints(
             {
                 if (data_matches_with_prefix(svr.guidPrefix, *pdata))
                 {
-                    std::unique_lock<std::recursive_mutex> lock(*getMutex());
-                    svr.proxy = get_participant_proxy_data(pdata->m_guid.guidPrefix);
+                    svr.is_connected = true;
                 }
             }
         }
@@ -517,11 +515,11 @@ void PDPClient::notifyAboveRemoteEndpoints(
         {
             if (data_matches_with_prefix(svr.guidPrefix, pdata))
             {
-                if (nullptr == svr.proxy)
+                if (!svr.is_connected && nullptr != get_participant_proxy_data(svr.guidPrefix))
                 {
-                    //! try to retrieve the participant proxy data from an unmangled prefix in case
-                    //! we could not fill svr.proxy in assignRemoteEndpoints()
-                    svr.proxy = get_participant_proxy_data(svr.guidPrefix);
+                    //! mark proxy as connected from an unmangled prefix in case
+                    //! it could not be done in assignRemoteEndpoints()
+                    svr.is_connected = true;
                 }
 
                 match_pdp_reader_nts_(svr, pdata.m_guid.guidPrefix);
@@ -596,7 +594,7 @@ void PDPClient::removeRemoteEndpoints(
             if (svr.guidPrefix == pdata->m_guid.guidPrefix)
             {
                 std::unique_lock<std::recursive_mutex> lock(*getMutex());
-                svr.proxy = nullptr; // reasign when we receive again server DATA(p)
+                svr.is_connected = false;
                 is_server = true;
                 mp_sync->restart_timer(); // enable announcement and sync mechanism till this server reappears
             }
@@ -768,11 +766,11 @@ void PDPClient::announceParticipantState(
                     for (auto& svr : mp_builtin->m_DiscoveryServers)
                     {
                         // if we are matched to a server report demise
-                        if (svr.proxy != nullptr)
+                        if (svr.is_connected)
                         {
                             //locators.push_back(svr.metatrafficMulticastLocatorList);
                             locators.push_back(svr.metatrafficUnicastLocatorList);
-                            remote_readers.emplace_back(svr.proxy->m_guid.guidPrefix,
+                            remote_readers.emplace_back(svr.guidPrefix,
                                     endpoints->reader.reader_->getGuid().entityId);
                         }
                     }
@@ -805,7 +803,7 @@ void PDPClient::announceParticipantState(
                     {
                         // non-pinging announcements like lease duration ones must be
                         // broadcast to all servers
-                        if (svr.proxy == nullptr || !_serverPing)
+                        if (!svr.is_connected || !_serverPing)
                         {
                             locators.push_back(svr.metatrafficMulticastLocatorList);
                             locators.push_back(svr.metatrafficUnicastLocatorList);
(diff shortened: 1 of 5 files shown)


Tempate:

Tempate:

MiguelCompany (changes requested):
1. There is a test that has failed with segmentation fault both in Mac and Windows
    * [Mac](http://jenkins.eprosima.com:8080/job/fastdds_mac/5011/testReport/junit/projectroot.test/blackbox/DDSDiscovery_AddDiscoveryServerToList/)
    * [Windows](http://jenkins.eprosima.com:8080/job/fastdds_windows/label=windows-secure,toolset=v142/5463/testReport/projectroot.test/blackbox/DDSDiscovery_AddDiscoveryServerToList/)
2. I don't like the solution of copying the whole `ParticipantProxyData`, it might be better to change the API of the methods receiving it so they only use the necessary information, and just copy that information

Tempate:

src/cpp/rtps/builtin/discovery/participant/PDPClient.cpp
```diff
@@ -517,11 +515,11 @@ void PDPClient::notifyAboveRemoteEndpoints(
...
             {
-                if (nullptr == svr.proxy)
+                if (!svr.is_connected && nullptr != get_participant_proxy_data(svr.guidPrefix))
                 {
-                    //! try to retrieve the participant proxy data from an unmangled prefix in case
-                    //! we could not fill svr.proxy in assignRemoteEndpoints()
-                    svr.proxy = get_participant_proxy_data(svr.guidPrefix);
+                    //! mark proxy as connected from an unmangled prefix in case
+                    //! it could not be done in assignRemoteEndpoints()
+                    svr.is_connected = true;
```
juanlofer-eprosima, line 522: This block might be deleted, however I took the most conservative approach.

juanlofer-eprosima:

JesusPoderoso:
@juanlofer-eprosima please address linter issue. Is the linux discovery failed test related with the PR?

Mario-DL (approved):
Good job in this PR as it effectively solves the data race it describes, I also like the simplification on the `PDPClient` in order to use the `is_connected`, avoiding to store the entire `ParticipantProxyData`. I am approving though leaving a reflection.

Trying to give a second thought on @MiguelCompany suggested, I think this work may admit an even fine-grained solution to avoid copying the entire `ParcipantProxyData` in stack. But it is also true that we are highly tied to the entire `pdata` within `PDP::assign_remote_endpoints()`  when we are notifying the user with the `DiscoveredParticipantInfo`. The rest of the methods, `matched_reader/writer_add()` and `assign_low_level_endpoints()` do not make an intensive use of the `ParticipantProxyData` and it would be enough to pass them only some of the fields: some `locator lists`, the `guid` and the `m_availableBuiltinEndpoints` in essence.

Finally, I am attaching a possible patch of a different approach instead of stack-copying, which explores the option of synchronize the `PDP::remove_remote_participant()`  and `PDP::assign_remote_endpoints()` **only**  in the case that we are trying to delete a  `pdata` that is is currently being processed. The fix is fast and dirty just to make an idea of the approach and I would still need to think if it would solve all the situations that we are solving by making that stack copy.

[pdp_patch_another_approach.txt](https://github.com/eProsima/Fast-DDS/files/14372684/pdp_patch_another_approach.txt)

Approving !

Mario-DL:
I included the backports in the PR description

EduPonz:
@Mergifyio backport 2.12.x 2.10.x 2.6.x

juanlofer-eprosima:
> Good job in this PR as it effectively solves the data race it describes, I also like the simplification on the `PDPClient` in order to use the `is_connected`, avoiding to store the entire `ParticipantProxyData`. I am approving though leaving a reflection.
> 
> Trying to give a second thought on @MiguelCompany suggested, I think this work may admit an even fine-grained solution to avoid copying the entire `ParcipantProxyData` in stack. But it is also true that we are highly tied to the entire `pdata` within `PDP::assign_remote_endpoints()` specially when we are notifying the user with the `DiscoveredParticipantInfo`. The rest of the methods, `matched_reader/writer_add()` and `assign_low_level_endpoints()` do not make an intensive use of the `ParticipantProxyData` and it would be enough to pass them only some of the fields: some `locator lists`, the `guid` and the `m_availableBuiltinEndpoints` in essence.
> 
> Finally, I am attaching a possible patch of a different approach instead of stack-copying, which explores the option of synchronizing the `PDP::remove_remote_participant()` and `PDP::assign_remote_endpoints()` **only** in the case that we are trying to delete a `pdata` that is is currently being processed. The fix is fast and dirty just to make an idea of the approach and I would still need to think if it would solve all the situations that we are solving by making that stack copy.
> 
> [pdp_patch_another_approach.txt](https://github.com/eProsima/Fast-DDS/files/14372684/pdp_patch_another_approach.txt)
> 
> Approving !

I also agree the solution @MiguelCompany suggested was better, but we decided to proceed with the copy approach given the effort required, and also considering that the root of the problem (releasing the PDP mutex before completing an entire discovery operation, which might lead to inconsistent states apart from data races as the one we are dealing with here) would still remain unsolved. We also discussed that a possible solution to tackle the root of the problem would be to derive all discovery actions to a dedicated thread.

Regarding your proposal, I believe it is a bit risky to wait as some lower level resources might be blocked by that thread, leading to potential deadlocks. And two additional comments:

- I believe that (in remove_remote_participant) locking the PDP mutex should be done after calling wait (to avoid intraprocess deadlock).
- I also believe that a PDP instance could process more than one data P before processing a data uP, so I guess the complexity of this solution would need to grow.

juanlofer-eprosima:
> > > Good job in this PR as it effectively solves the data race it describes, I also like the simplification on the `PDPClient` in order to use the `is_connected`, avoiding to store the entire `ParticipantProxyData`. I am approving though leaving a reflection.
> > > Trying to give a second thought on @MiguelCompany suggested, I think this work may admit an even fine-grained solution to avoid copying the entire `ParcipantProxyData` in stack. But it is also true that we are highly tied to the entire `pdata` within `PDP::assign_remote_endpoints()` specially when we are notifying the user with the `DiscoveredParticipantInfo`. The rest of the methods, `matched_reader/writer_add()` and `assign_low_level_endpoints()` do not make an intensive use of the `ParticipantProxyData` and it would be enough to pass them only some of the fields: some `locator lists`, the `guid` and the `m_availableBuiltinEndpoints` in essence.
> > > Finally, I am attaching a possible patch of a different approach instead of stack-copying, which explores the option of synchronizing the `PDP::remove_remote_participant()` and `PDP::assign_remote_endpoints()` **only** in the case that we are trying to delete a `pdata` that is is currently being processed. The fix is fast and dirty just to make an idea of the approach and I would still need to think if it would solve all the situations that we are solving by making that stack copy.
> > > [pdp_patch_another_approach.txt](https://github.com/eProsima/Fast-DDS/files/14372684/pdp_patch_another_approach.txt)
> > > Approving !
> > 
> > 
> > I also agree the solution @MiguelCompany suggested was better, but we decided to proceed with the copy approach given the effort required, and also considering that the root of the problem (releasing the PDP mutex before completing an entire discovery operation, which might lead to inconsistent states apart from data races as the one we are dealing with here) would still remain unsolved. We also discussed that a possible solution to tackle the root of the problem would be to derive all discovery actions to a dedicated thread.
> > Regarding your proposal, I believe it is a bit risky to wait as some lower level resources might be blocked by that thread, leading to potential deadlocks. And two additional comments:
> > 
> > * I believe that (in remove_remote_participant) locking the PDP mutex should be done after calling wait (to avoid intraprocess deadlock).
> > * I also believe that a PDP instance could process more than one data P before processing a data uP, so I guess the complexity of this solution would need to grow.
> 
> Yeah, agree that it would be risky as a consequence of relying on a cond var.
> 
> A further `lock-free` approach I can think of would be that the thread executing `remove_remote_participant()` with the participant proxy that is being currently processed in `assign_remote_endpoints `, could leave a flag to true in PDPListener meaning that _“I am not removing the participant proxy now, as you are using it, but I notify you that when you finish the `assign_remote_endpoints() ` you will have to delete it on behalf of me”_ So, in that case, the same discovery thread could call `remove_remote_participant() ` if the flag is up, after executing the `assign_remote_endpoints() ` and then reset it. What do you think ?

That's an interesting approach we could further explore. The only caveat I see at the moment is that `remove_remote_participant` can also be called from places other than a `PDPListener`, like for example from the events thread when liveliness is lost.

PR #4335, opened 2024-01-30
[20311] Add missing TypeLookup listeners

irenebm:
## Description

This pull request addresses an issue where Fast DDS fails to discover certain types when receiving a significant number of dynamic types (approximately 25).

This PR implements the `onWriterChangeReceivedByAll` listener for both `TypeLookupReplyListener` and `TypeLookupRequestListener`. This enables the deletion of changes received by all readers matched with the writer from the history. It prevents the `CacheChange` from becoming saturated, thus enabling the system to accommodate further changes effectively.

@Mergifyio backport 2.12.x 2.10.x 2.6.x

## Contributor Checklist
- _N/A_; Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- N/A New feature has been added to the `versions.md` file (if applicable).
- N/A New feature has been documented/Current behavior is correctly described in the documentation. 
    
## Reviewer Checklist

### DIFF
--- a/include/fastdds/dds/builtin/typelookup/TypeLookupReplyListener.hpp
+++ b/include/fastdds/dds/builtin/typelookup/TypeLookupReplyListener.hpp
@@ -72,7 +72,7 @@ public:
      */
     void onNewCacheChangeAdded(
             fastrtps::rtps::RTPSReader* reader,
-            const fastrtps::rtps::CacheChange_t* const  change) override;
+            const fastrtps::rtps::CacheChange_t* const change) override;
 
 private:
 
@@ -117,5 +117,5 @@ private:
 } /* namespace dds */
 } /* namespace fastdds */
 } /* namespace eprosima */
-#endif
+#endif // ifndef DOXYGEN_SHOULD_SKIP_THIS_PUBLIC
 #endif /* TYPELOOKUP_REPLY_LISTENER_HPP_*/
--- a/include/fastdds/dds/builtin/typelookup/TypeLookupReplyListener.hpp
+++ b/include/fastdds/dds/builtin/typelookup/TypeLookupReplyListener.hpp
@@ -21,6 +21,8 @@
 #define TYPELOOKUP_REPLY_LISTENER_HPP_
 #ifndef DOXYGEN_SHOULD_SKIP_THIS_PUBLIC
 #include <fastrtps/rtps/reader/ReaderListener.h>
+#include <fastrtps/rtps/writer/WriterListener.h>
+
 
 namespace eprosima {
 namespace fastrtps {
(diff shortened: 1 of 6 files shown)


include/fastdds/dds/builtin/typelookup/TypeLookupReplyListener.hpp
```diff
@@ -81,9 +83,39 @@ class TypeLookupReplyListener : public fastrtps::rtps::ReaderListener
     fastrtps::types::TypeObjectFactory* factory_;
 };
 
+class TypeLookupReplyWListener : public fastrtps::rtps::WriterListener
```
EduPonz, line 86: Instead of creating a new pair of classes, I'd make the already existing ones to also implement the `WriterListener` interface. That way we'd have less friends, less members, less news/deletes, etc. so easier maintainability. Applies to the request listener as well

src/cpp/fastdds/builtin/typelookup/TypeLookupManager.cpp
```diff
@@ -567,7 +578,13 @@ bool TypeLookupManager::send_request(
         SerializedPayload_t payload;
         payload.max_size = change->serializedPayload.max_size - 4;
         payload.data = change->serializedPayload.data + 4;
-        if (valid && request_type_.serialize(&req, &payload, DataRepresentationId_t::XCDR2_DATA_REPRESENTATION))
+
+        bool serialize_ret = request_type_.serialize(&req, &payload, DataRepresentationId_t::XCDR_DATA_REPRESENTATION);
```
EduPonz, line 582:
```suggestion
        bool serialize_ret = request_type_.serialize(&req, &payload, DataRepresentationId_t::XCDR2_DATA_REPRESENTATION);
```

src/cpp/fastdds/builtin/typelookup/TypeLookupManager.cpp
```diff
@@ -610,7 +627,13 @@ bool TypeLookupManager::send_reply(
         SerializedPayload_t payload;
         payload.max_size = change->serializedPayload.max_size - 4;
         payload.data = change->serializedPayload.data + 4;
-        if (valid && reply_type_.serialize(&rep, &payload, DataRepresentationId_t::XCDR2_DATA_REPRESENTATION))
+
+        bool serialize_ret = reply_type_.serialize(&rep, &payload, DataRepresentationId_t::XCDR_DATA_REPRESENTATION);
```
EduPonz, line 631:
```suggestion
        bool serialize_ret = reply_type_.serialize(&rep, &payload, DataRepresentationId_t::XCDR2_DATA_REPRESENTATION);
```

EduPonz (changes requested):

JesusPoderoso:

src/cpp/fastdds/builtin/typelookup/TypeLookupManager.cpp
```diff
@@ -348,14 +348,15 @@ bool TypeLookupManager::create_endpoints()
     // Built-in request writer
     if (builtin_protocols_->m_att.typelookup_config.use_client)
     {
+        request_listener_ = new TypeLookupRequestListener(this);
```
EduPonz, line 351: On the creation of the request reader there is another statement just like this one. Since this is the first time, the `new` should be done here, but on the reader creation it should be changed to:
1. Checking if it's `nullptr` before calling `new` (else we'll have leaks)
2. Only call `delete` if the listener was created for the reader (meaning if it had been created for the writter then it should not be deleted there)

I think I'd refactor this method to:
1. Be single return
2. Have a `ret` boolean initialized to `true` and used as entry condition to each step, if something fails it should be changed to `false`.
3. At the very end check `ret` and only call delete if `false`

EduPonz (changes requested):

EduPonz:

EduPonz (approved):

PR #4534, opened 2024-03-08
[20543] Create Participant with default profile (use environment XML configuration)

JesusPoderoso:
## Description

DomainParticipant XML profile configuration allows setting the [domain id](https://fast-dds.docs.eprosima.com/en/latest/fastdds/xml_configuration/domainparticipant.html#domainparticipant-configuration) out of the rtps.
That domain ID is considered in the domain participant creation only if [the domain participant profile name is defined](https://github.com/eProsima/Fast-DDS/blob/5c87d9d02173f190f7b69cb4dbbb6041d9fcf177/src/cpp/fastdds/domain/DomainParticipantFactory.cpp#L242).

This PR stores (if exists) the Domain ID value while loading the default XML profile, and exposes new methods to simplify participant creation without requiring neither the Domain ID (uses the one set in the XML, or `0` by default), neither the QoS (uses the qos set in the XML, or PARTICIPANT_DEFAULT_QOS by default)

## Contributor Checklist
- [x] Commit messages follow the project guidelines. 
- [x] The code follows the style guidelines of this project. 
- [x] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- [x] Any new/modified methods have been properly documented using Doxygen. 
- [x] Changes are ABI compatible. 
- [x] Changes are API compatible. 
- [x] New feature has been added to the `versions.md` file (if applicable).
- [x] New feature has been documented/Current behavior is correctly described in the documentation. 
    Related documentation PR: eProsima/Fast-DDS-docs#725
- **N/A** Applicable backports have been included in the description.

## Reviewer Checklist
- [x] The PR has a milestone assigned.
- [x] The title and description correctly express the PR's purpose.
- [x] Check contributor checklist is correct.
- [x] Check CI results: changes do not issue any warning.
- [x] Check CI results: failing tests are unrelated with the changes.

PR #4547, opened 2024-03-11
[20543] Examples refactor: Hello World

JesusPoderoso:
## Description

This PR should be merge right after:
- #4534 

This PR is the first of a suite of PR which would make a refactor in the repository examples.
It is intended to apply to most of the examples, by making them homogeneous, more understandable, and more specific to the case they were meant to be.

In this hello world example, the key changes are:
- remove --env option and set that environment behavior as default
- add subscriber waitsets class and its usage
- provide XML profiles examples that target several scenarios (other tests as SampleConfig_Controller / Events / Multimedia)

The following changes apply to this and the remain examples:
- remove the DDS prefix from the binaries
- snake case in binary names and directory names
- launch only one endpoint per binary (always running them in different shells)
- unify args names and utility
- write proper README.md with the purpose of the example and expected inputs, outputs, behavior… (do not include options in readme → enhance user to run --help )
- implement listener callback in the same class (inheritance) 
- include initialization ( init() ) methods in the constructors
- improve parsing options to be a black box to the user

## Contributor Checklist
    
## Reviewer Checklist


examples/cpp/hello_world/cli_options.hpp
```diff
@@ -0,0 +1,75 @@
...
+
+void print_help()
+{
+    std::cout << "Usage: hello_world [entity] [options]\n"
+              << "Entities:                       Mandatory field\n"
+              << "  publisher\n"
+              << "  subscriber\n"
+              << "Options:                        Optional fields:\n"
+              << "  -w, --waitset                 Use waitset read condition (subscriber only)\n"
+              << "  -h, --help                    Print this help message\n";
```
EduPonz, line 23: IMO, using one cout and one endl per line looks cleaner:
```suggestion
    std::cout << "Usage: hello_world <entity> [options]"              << std::endl;
    std::cout << ""                                                   << std::endl;
    std::cout << "Entities:"                                          << std::endl;
    std::cout << "  publisher             Run a publisher entity"     << std::endl;
    std::cout << "  subscriber            Run a subscriber entity"    << std::endl;
    std::cout << ""                                                   << std::endl;
    std::cout << "Common options:"                                    << std::endl;
    std::cout << "    -h, --help          Print this help message"    << std::endl;
    std::cout << ""                                                   << std::endl;
    std::cout << "Subscriber options:"                                << std::endl;
    std::cout << "    -w, --waitset       Use waitset read condition" << std::endl;
```

examples/cpp/hello_world/cli_options.hpp
```diff
@@ -0,0 +1,75 @@
...
+    config.entity = "";
+    config.sub_config.use_waitset = false;
+
+    for (int i = 1; i < argc; ++i)
+    {
+        std::string arg = argv[i];
+        if (arg == "-h" || arg == "--help")
+        {
+            print_help();
+            exit(EXIT_SUCCESS);
```
EduPonz, line 40: I'd pass the exit code to the `help` function and exit from there instead

test/examples/hello_world.compose.yml
```diff
@@ -0,0 +1,29 @@
+# FASTDDS_TODO_BEFORE(3, 0, "This compose file should be used for the future configuration example");
+version: "3"
+
+services:
+  publisher:
+    image: @DOCKER_IMAGE_NAME@
+    volumes:
+      - @PROJECT_BINARY_DIR@:@PROJECT_BINARY_DIR@
+      - @fastcdr_LIB_DIR@:@fastcdr_LIB_DIR@
+      - @PROJECT_BINARY_DIR@/examples/cpp/hello_world/HelloWorld_controller_profile.xml:@PROJECT_BINARY_DIR@/examples/cpp/hello_world/HelloWorld_controller_profile.xml
```
EduPonz, line 10: This is not needed, as the `PROJECT_BINARY_DIR` is already shared

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+    std::thread pub_thread([&]
+            {
+                while (!is_stopped())
+                {
+                    if (publish())
+                    {
+                        std::cout << "Message: '" << hello_.message() << "' with index: '" << hello_.index()
+                                  << "' SENT" << std::endl;
+                    }
+                    std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
EduPonz, line 135: I'd make the period a class member

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+                }
+            });
+    std::cout << "Publisher running. Please press Ctrl+C to stop the Publisher at any time." << std::endl;
+    signal(SIGINT, [](int signum)
+            {
+                std::cout << "SIGINT received, stopping Publisher execution." << std::endl;
+                static_cast<void>(signum); HelloWorldPublisher::stop();
+            });
```
EduPonz, line 143: We should register handles for other signals IMO
```suggestion
    signal(SIGINT, [](int /*signum*/)
            {
                std::cout << "SIGINT received, stopping Publisher execution." << std::endl;
                HelloWorldPublisher::stop();
            });
```

examples/cpp/hello_world/HelloWorldPublisher.h
```diff
@@ -0,0 +1,77 @@
...
+            const PublicationMatchedStatus& info) override;
+
+    //! Publish a sample
+    bool publish();
+
+    //! Run publisher
+    void run();
+
+    //! Return the current state of execution
+    static bool is_stopped();
```
EduPonz, line 51: This method can be private I think

examples/cpp/hello_world/HelloWorldPublisher.h
```diff
@@ -0,0 +1,77 @@
...
+
+    Publisher* publisher_;
+
+    Topic* topic_;
+
+    DataWriter* writer_;
+
+    TypeSupport type_;
+
+    static std::atomic<bool> stop_;
```
EduPonz, line 70: Why is this static?

examples/cpp/hello_world/HelloWorldPublisher.h
EduPonz, line 51: Why is this static?

examples/cpp/hello_world/HelloWorldPublisher.h
```diff
@@ -0,0 +1,77 @@
...
+    bool publish();
+
+    //! Run publisher
+    void run();
+
+    //! Return the current state of execution
+    static bool is_stopped();
+
+    //! Trigger the end of execution
+    static void stop();
```
EduPonz, line 54: Why is this static?

examples/cpp/hello_world/HelloWorldPublisher.h
```diff
@@ -0,0 +1,77 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @file HelloWorldPublisher.h
+ *
+ */
+
+#ifndef HELLO_WORLD_PUBLISHER_H_
```
EduPonz, line 20:
```suggestion
#ifndef _FASTDDS_HELLO_WORLD_PUBLISHER_HPP_
```

examples/cpp/hello_world/HelloWorldPublisher.h
EduPonz, line 1: Make the headers .hpp

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+        throw std::runtime_error("Participant initialization failed");
+    }
+
+    // Register the type
+    type_.register_type(participant_);
+
+    // Create the publisher
+    PublisherQos pub_qos = PUBLISHER_QOS_DEFAULT;
+    participant_->get_default_publisher_qos(pub_qos);
+    publisher_ = participant_->create_publisher(pub_qos, nullptr);
```
EduPonz, line 61:
```suggestion
    publisher_ = participant_->create_publisher(pub_qos);
```

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+    if (publisher_ == nullptr)
+    {
+        throw std::runtime_error("Publisher initialization failed");
+    }
+
+    // Create the topic
+    TopicQos topic_qos = TOPIC_QOS_DEFAULT;
+    participant_->get_default_topic_qos(topic_qos);
+    topic_ = participant_->create_topic("Hello_world_topic", "HelloWorld", topic_qos);
```
EduPonz, line 70: Get the type name from the type support. Same for subs

EduPonz, line 70:
```suggestion
    topic_ = participant_->create_topic("hello_world_topic", "HelloWorld", topic_qos);
```

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+        publisher_->delete_datawriter(writer_);
+    }
+    if (publisher_ != nullptr)
+    {
+        participant_->delete_publisher(publisher_);
+    }
+    if (topic_ != nullptr)
+    {
+        participant_->delete_topic(topic_);
+    }
```
EduPonz, line 99: Use participant's `delete_contained_entities`. Same elsewhere

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+    DomainParticipantFactory::get_instance()->delete_participant(participant_);
+}
+
+void HelloWorldPublisher::on_publication_matched(
+        DataWriter*,
```
EduPonz, line 104: Just aesthetics
```suggestion
        DataWriter* /*writer*/,
```

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+            {
+                std::cout << "SIGINT received, stopping Publisher execution." << std::endl;
+                static_cast<void>(signum); HelloWorldPublisher::stop();
+            });
+    pub_thread.join();
+}
+
+bool HelloWorldPublisher::publish()
+{
+    if (matched_ > 0)
```
EduPonz, line 149: I'm not very much convinced with the matching busy waiting

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {
+        hello_.index(hello_.index() + 1);
+        writer_->write(&hello_);
+        return true;
```
EduPonz, line 153: `write` returns either a code or a bool, use it to return here

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+    return false;
```
EduPonz, line 155: Try to have a single return per function

examples/cpp/hello_world/HelloWorldPublisher.h
```diff
@@ -0,0 +1,77 @@
+// Copyright 2019 Proyectos y Sistemas de Mantenimiento SL (eProsima).
```
EduPonz, line 1: This is everywhere
```suggestion
// Copyright 2024 Proyectos y Sistemas de Mantenimiento SL (eProsima).
```

examples/cpp/hello_world/HelloWorldSubscriber.cpp
```diff
@@ -152,33 +121,43 @@ void HelloWorldSubscriber::SubListener::on_subscription_matched(
     }
 }
 
-void HelloWorldSubscriber::SubListener::on_data_available(
+void HelloWorldSubscriber::on_data_available(
         DataReader* reader)
 {
     SampleInfo info;
-    if (reader->take_next_sample(&hello_, &info) == ReturnCode_t::RETCODE_OK)
+    if (reader->take_next_sample(&hello_, &info) == ReturnCode_t::RETCODE_OK && !is_stopped())
```
EduPonz, line 128:
```suggestion
    if ((!is_stopped()) && (ReturnCode_t::RETCODE_OK == reader->take_next_sample(&hello_, &info)))
```

examples/cpp/hello_world/HelloWorldSubscriber.cpp
```diff
@@ -152,33 +121,43 @@ void HelloWorldSubscriber::SubListener::on_subscription_matched(
...
     {
         if (info.instance_state == ALIVE_INSTANCE_STATE)
```
EduPonz, line 130: IMO, is's a good practice to check `valid_data` as well`

examples/cpp/hello_world/HelloWorldSubscriber.cpp
```diff
@@ -152,33 +121,43 @@ void HelloWorldSubscriber::SubListener::on_subscription_matched(
...
 void HelloWorldSubscriber::run()
 {
-    std::cout << "Subscriber running. Please press enter to stop the Subscriber" << std::endl;
-    std::cin.ignore();
+    std::cout << "Subscriber running. Please press Ctrl+C to stop the Subscriber at any time." << std::endl;
+    signal(SIGINT, [](int signum)
+            {
+                std::cout << "SIGINT received, stopping Subscriber execution." << std::endl;
+                static_cast<void>(signum); HelloWorldSubscriber::stop();
+            });
```
EduPonz, line 146: Same suggestions as in the pub

examples/cpp/hello_world/HelloWorldSubscriber.h
```diff
@@ -0,0 +1,80 @@
...
+
+    Topic* topic_;
+
+    DataReader* reader_;
+
+    TypeSupport type_;
+
+    static std::atomic<bool> stop_;
+
+    static std::mutex terminate_cv_mtx_;
```
EduPonz, line 75:
```suggestion
    mutable std::mutex terminate_cv_mtx_;
```

examples/cpp/hello_world/HelloWorldSubscriber.cpp
```diff
@@ -19,99 +19,70 @@
...
-    {
-        subscriber_->get_default_datareader_qos(rqos);
+        throw std::runtime_error("Topic initialization failed");
     }
 
-    reader_ = subscriber_->create_datareader(topic_, rqos, &listener_);
-
+    // Create the reader
+    DataReaderQos reader_qos = DATAREADER_QOS_DEFAULT;
+    reader_qos.reliability().kind = RELIABLE_RELIABILITY_QOS;
```
EduPonz, line 79: I guess we should document this

examples/cpp/hello_world/HelloWorldPublisher.cpp
```diff
@@ -0,0 +1,166 @@
...
+    participant_->get_default_topic_qos(topic_qos);
+    topic_ = participant_->create_topic("Hello_world_topic", "HelloWorld", topic_qos);
+    if (topic_ == nullptr)
+    {
+        throw std::runtime_error("Topic initialization failed");
+    }
+
+    // Create the data writer
+    DataWriterQos writer_qos = DATAWRITER_QOS_DEFAULT;
+    publisher_->get_default_datawriter_qos(writer_qos);
```
EduPonz, line 78: Shouldn't the writer always be transient local and reliable?, probably we should fix the depth as well? Same for sub

examples/cpp/hello_world/HelloWorldSubscriberWaitset.cpp
```diff
@@ -0,0 +1,189 @@
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            });
+    std::cout << "Waitset Subscriber running. Please press Ctrl+C to stop the Waitset Subscriber at any time."
+              << std::endl;
+    signal(SIGINT, [](int signum)
```
EduPonz, line 172:
```suggestion

    std::cout << "Waitset Subscriber running. Please press Ctrl+C to stop the Waitset Subscriber at any time."
              << std::endl;

    signal(SIGINT, [](int signum)
```

examples/cpp/hello_world/HelloWorldSubscriberWaitset.cpp
```diff
@@ -0,0 +1,189 @@
...
+            {
+                std::cout << "SIGINT received, stopping Waitset Subscriber execution." << std::endl;
+                static_cast<void>(signum); HelloWorldSubscriberWaitset::stop();
+            });
```
EduPonz, line 176: Same as before

examples/cpp/hello_world/HelloWorldSubscriberWaitset.cpp
```diff
@@ -0,0 +1,189 @@
...
+    if (topic_ == nullptr)
+    {
+        throw std::runtime_error("Topic initialization failed");
+    }
+
+    // Create the reader
+    DataReaderQos reader_qos = DATAREADER_QOS_DEFAULT;
+    reader_qos.reliability().kind = RELIABLE_RELIABILITY_QOS;
+    subscriber_->get_default_datareader_qos(reader_qos);
+    reader_ = subscriber_->create_datareader(topic_, reader_qos);
```
EduPonz, line 84: Should we be explicit about the status mask used? Same on the others

examples/cpp/hello_world/HelloWorldSubscriberWaitset.h
```diff
@@ -0,0 +1,73 @@
...
+
+    HelloWorldSubscriberWaitset();
+
+    virtual ~HelloWorldSubscriberWaitset();
+
+    //! Run subscriber
+    void run();
+
+    //! Return the current state of execution
+    static bool is_stopped();
```
EduPonz, line 47: Again, private; no statics if possible

examples/cpp/hello_world/HelloWorld_controller_profile.xml
EduPonz, line 1: Please rename the file to snake case

examples/cpp/hello_world/HelloWorld_controller_profile.xml
```diff
@@ -0,0 +1,53 @@
...
+        <topic>
+            <historyQos>
+                <kind>KEEP_LAST</kind>
+                <depth>100</depth>
+            </historyQos>
+            <resourceLimitsQos>
+                <max_samples>100</max_samples>
+                <max_instances>1</max_instances>
+                <max_samples_per_instance>100</max_samples_per_instance>
+            </resourceLimitsQos>
```
EduPonz, line 28: I don't think this is what I'd do for a controller

examples/cpp/hello_world/HelloWorld_event_profile.xml
EduPonz, line 1: Rename file

examples/cpp/hello_world/HelloWorld_event_profile.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<profiles xmlns="http://www.eprosima.com" >
+    <participant profile_name="hello_world_events_participant_profile" is_default_profile="true">
+        <domainId>0</domainId>
+        <rtps>
+            <name>hello_world_events_participant</name>
+            <builtinTransports>UDPv4</builtinTransports>
+        </rtps>
+    </participant>
+    <data_writer profile_name="hello_world_events_dataWriter_profile" is_default_profile="true">
```
EduPonz, line 10: Apply everywhere
```suggestion
    <data_writer profile_name="hello_world_events_datawriter_profile" is_default_profile="true">
```

examples/cpp/hello_world/HelloWorld_event_profile.xml
```diff
@@ -0,0 +1,58 @@
...
+                <max_samples>10</max_samples>
+                <max_instances>1</max_instances>
+                <max_samples_per_instance>10</max_samples_per_instance>
+            </resourceLimitsQos>
+        </topic>
+        <times>
+            <heartbeatPeriod>
+                <nanosec>10000000</nanosec>
+            </heartbeatPeriod>
+        </times>
```
EduPonz, line 34:
```suggestion
```

examples/cpp/hello_world/HelloWorld_main.cpp
```diff
@@ -0,0 +1,91 @@
...
+            {
+                EPROSIMA_LOG_ERROR(SUBSCRIBER_WAITSET, e.what());
+                ret = EXIT_FAILURE;
+            }
+        }
+    }
+    // example should never reach this point
+    else
+    {
+        std::cerr << "Error: unknown entity " << config.entity << "\n";
```
EduPonz, line 84: Why not using the log here?

In fact, we could configure the log to use an stdouterr consumer

examples/cpp/hello_world/HelloWorld_multimedia_profile.xml
EduPonz, line 1: Rename file. We need to discuss what to do with this profiles internally

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,198 @@
...
+user@machine:example_path$ export FASTDDS_DEFAULT_PROFILES_FILE=example_profile.xml
+```
+
+#### Windows
+
+```powershell
+example_path> set FASTDDS_DEFAULT_PROFILES_FILE=example_profile.xml
+```
+
+Other alternative is naming the XML profiles file as ``DEFAULT_FASTDDS_PROFILES.xml`` and make sure the file is besides the ``hello_world`` executable.
```
EduPonz, line 175: I would not point to that here
```suggestion
```

examples/cpp/hello_world/cli_options.hpp
```diff
@@ -0,0 +1,75 @@
+#include <iostream>
```
EduPonz, line 1: License is missing

examples/cpp/hello_world/cli_options.hpp
EduPonz, line 40: Remember to include `cstdlib`
```suggestion
            std::exit(EXIT_SUCCESS);
```

test/examples/hello_world.compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    environment:
+      # TODO(eduponz): LD_LIBRARY_PATH is not the correct variable for Windows
+      LD_LIBRARY_PATH: @PROJECT_BINARY_DIR@/src/cpp:@fastcdr_LIB_DIR@
+      EXAMPLE_DIR: @PROJECT_BINARY_DIR@/examples/cpp/hello_world
+      FASTDDS_DEFAULT_PROFILES_FILE: @PROJECT_BINARY_DIR@/examples/cpp/hello_world/HelloWorld_controller_profile.xml
+    command: @SHELL_EXECUTABLE@ -c "$${EXAMPLE_DIR}/hello_world@FILE_EXTENSION@ publisher  & $${EXAMPLE_DIR}/hello_world@FILE_EXTENSION@ subscriber --waitset"
```
EduPonz, line 16:
```suggestion
    command: @SHELL_EXECUTABLE@ -c "$${EXAMPLE_DIR}/hello_world@FILE_EXTENSION@ publisher & $${EXAMPLE_DIR}/hello_world@FILE_EXTENSION@ subscriber --waitset"
```

test/examples/hello_world.compose.yml
```diff
@@ -0,0 +1,29 @@
...
+
+  subscriber:
```
EduPonz, line 18:
```suggestion
  waitset-subscriber:
```

test/examples/hello_world.compose.yml
```diff
@@ -0,0 +1,29 @@
+# FASTDDS_TODO_BEFORE(3, 0, "This compose file should be used for the future configuration example");
+version: "3"
+
+services:
+  publisher:
```
EduPonz, line 5:
```suggestion
  publisher-subscriber:
```

test/examples/test_examples.py
```diff
@@ -37,3 +39,35 @@ def test_basic_configuration():
         print('TIMEOUT')
 
     assert(ret)
+
+def test_hello_world():
+    """."""
+    ret = False
+    out = ''
+    with Popen('@DOCKER_EXECUTABLE@ compose -f hello_world.compose.yml up',
```
EduPonz, line 47: I think we can use `run` instead, can't we?

test/examples/test_examples.py
```diff
@@ -1,6 +1,8 @@
 """."""
+import os
+import signal
 import subprocess
-
+from subprocess import Popen, PIPE, TimeoutExpired
```
EduPonz, line 5: Why already include the subprocess module
```suggestion
```

examples/cpp/hello_world/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.20)
 
-project(DDSHelloWorldExample VERSION 1 LANGUAGES CXX)
+project(hello_world VERSION 1 LANGUAGES CXX)
```
EduPonz, line 17:
```suggestion
project(fastdds_hello_world_example VERSION 1 LANGUAGES CXX)
```

examples/cpp/hello_world/HelloWorld.idl
```diff
@@ -1,3 +1,4 @@
+@extensibility(APPENDABLE)
```
EduPonz, line 1: Is this what we want?

EduPonz (changes requested):

examples/cpp/hello_world/HelloWorldPublisher.h
JesusPoderoso: Because `stop()` method needs to be static

examples/cpp/hello_world/HelloWorldSubscriber.cpp
JesusPoderoso: As discussed internally, this configuration would be part of the XML files

examples/cpp/hello_world/HelloWorld_controller_profile.xml
JesusPoderoso: Removed

examples/cpp/hello_world/HelloWorld_main.cpp
JesusPoderoso: How?

test/examples/test_examples.py
JesusPoderoso: Performing other approach for this

examples/cpp/hello_world/HelloWorld.idl
JesusPoderoso: Agreed internally

examples/cpp/hello_world/HelloWorldSubscriberWaitset.h
JesusPoderoso: I've got ride of the many static as possible, but as long as calling them from threads, most of them are required.

examples/cpp/hello_world/SubscriberWaitset.hpp
```diff
@@ -0,0 +1,79 @@
...
+    HelloWorldSubscriberWaitset(
+            const CLIParser::subscriber_config& config);
+
+    virtual ~HelloWorldSubscriberWaitset();
+
+    //! Run subscriber
+    void run();
+
+    //! Return the current state of execution
+    static bool is_stopped();
```
EduPonz, line 49: Private

examples/cpp/hello_world/SubscriberWaitset.cpp
```diff
@@ -0,0 +1,213 @@
...
+    {
+        std::cout << "Waitset Subscriber running. Please press Ctrl+C to stop the Waitset Subscriber at any time."
+                  << std::endl;
+    }
+    else
+    {
+        std::cout << "Waitset Subscriber running until " << samples_ <<
+            " samples have been received. Please press Ctrl+C to stop the Waitset Subscriber at any time." << std::endl;
+    }
+    signal(SIGINT, [](int /*signum*/)
```
EduPonz, line 179: Same as for the others

examples/cpp/hello_world/SubscriberWaitset.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+    // Prepare a wait-set
+    wait_set_.attach_condition(reader_->get_statuscondition());
+    wait_set_.attach_condition(terminate_condition_);
+}
+
+HelloWorldSubscriberWaitset::~HelloWorldSubscriberWaitset()
+{
+    // Delete DDS entities contained within the DomainParticipant
+    participant_->delete_contained_entities();
```
EduPonz, line 100: Same

test/examples/test_examples.py
```diff
@@ -25,7 +24,43 @@ def test_basic_configuration():
...
+        if sent != 0 and received != 0 and sent * 2 == received:
+            ret = True
+        else:
+            raise subprocess.CalledProcessError(1, '')
+
+    except subprocess.CalledProcessError:
+        for l in out:
+            print(l)
+    except subprocess.TimeoutExpired:
+        print('TIMEOUT')
```
EduPonz, line 36: I just realized that in case of timeout we are not printing and we probably should. Same goes for the hello_world example test

EduPonz (changes requested):

test/examples/test_examples.py
JesusPoderoso: I think that timeout exception makes out NOT to be populated.
We can try to print `out` but I would guess that nothing would be printed.

examples/cpp/hello_world/CLIParser.hpp
```diff
@@ -28,20 +28,26 @@ class CLIParser
 
     CLIParser() = delete;
 
+    enum entity_kind
```
EduPonz, line 31: Prefer enum class over plain enums. Also, please me explicit about the type. As an enum class, name should be capital camel cased. I'd suggest adding a namespace for the example, something like `eprosima::fastdds::examples::hello_world` so `EntityKind` does not clash with any other thing.

examples/cpp/hello_world/CLIParser.hpp
```diff
@@ -68,34 +75,49 @@ class CLIParser
             char* argv[])
     {
         hello_world_config config;
-        config.entity = "";
-        config.pub_config.samples = 0;
-        config.sub_config.samples = 0;
-        config.sub_config.use_waitset = false;
 
-        for (int i = 1; i < argc; ++i)
+        std::string first_argument = argv[1];
```
EduPonz, line 79: Please check argc before accessing. If not arguments are passed, this would be our of bounds. In fact, if no args are passed, we probably should `print_help(EXIT_FAILURE)`

examples/cpp/hello_world/CLIParser.hpp
```diff
@@ -68,34 +75,49 @@ class CLIParser
...
+
+        if (first_argument == "publisher" )
         {
-            std::string arg = argv[i];
-            if (arg == "-h" || arg == "--help")
+            config.entity = entity_kind::PUBLISHER;
+        }
+        else if ( first_argument == "subscriber")
```
EduPonz, line 85:
```suggestion
        else if (first_argument == "subscriber")
```

examples/cpp/hello_world/CLIParser.hpp
```diff
@@ -0,0 +1,172 @@
...
+                        }
+                        else if (config.entity == entity_kind::SUBSCRIBER)
+                        {
+                            config.sub_config.samples = samples;
+                        }
+                        else
+                        {
+                            EPROSIMA_LOG_ERROR(CLI_PARSER, "entity not specified for --sample argument");
+                            print_help(EXIT_FAILURE);
+
```
EduPonz, line 128:
```suggestion
```

examples/cpp/hello_world/CLIParser.hpp
```diff
@@ -0,0 +1,172 @@
...
+            if (first_argument == "-h" || first_argument == "--help")
+            {
+                print_help(EXIT_SUCCESS);
+            }
+            else
+            {
+                EPROSIMA_LOG_ERROR(CLI_PARSER, "parsing entity argument");
+                print_help(EXIT_FAILURE);
+            }
+        }
```
EduPonz, line 100:
```suggestion
```suggestion
        else if (first_argument == "-h" || first_argument == "--help")
        {
            print_help(EXIT_SUCCESS);
        }
        else
        {
            EPROSIMA_LOG_ERROR(CLI_PARSER, "parsing entity argument");
            print_help(EXIT_FAILURE);
        }
```
```

File: examples/cpp/hello_world/CLIParser.hpp [hunk 50]
```diff
@@ -0,0 +1,172 @@
...
+            {
+                print_help(EXIT_SUCCESS);
+            }
+            else if (arg == "-s" || arg == "--samples")
+            {
+                if (i + 1 < argc)
+                {
+                    try
+                    {
+                        uint16_t samples = static_cast<uint16_t>(std::stoi(argv[++i]));
```
EduPonz, line 115: What if my input is `-1`? Check [here](https://godbolt.org/z/GhcPq36Yb)

examples/cpp/hello_world/CMakeLists.txt
```diff
@@ -34,15 +34,29 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
...
     $<$<AND:$<NOT:$<BOOL:${WIN32}>>,$<STREQUAL:"${CMAKE_BUILD_TYPE}","Debug">>:__DEBUG>
     $<$<BOOL:${INTERNAL_DEBUG}>:__INTERNALDEBUG> # Internal debug activated.
+    $<$<BOOL:${SHM_TRANSPORT_DEFAULT}>:SHM_TRANSPORT_BUILTIN> # Enable SHM as built-in transport
 )
-target_link_libraries(DDSHelloWorldExample fastdds fastcdr fastdds::optionparser)
-install(TARGETS DDSHelloWorldExample
-    RUNTIME DESTINATION examples/cpp/dds/HelloWorldExample/${BIN_INSTALL_DIR})
+target_link_libraries(hello_world fastdds fastcdr)
+install(TARGETS hello_world
+    RUNTIME DESTINATION examples/cpp/hello_world/${BIN_INSTALL_DIR})
```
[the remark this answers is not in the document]
EduPonz: Let's decide offline

examples/cpp/hello_world/CMakeLists.txt
```diff
@@ -34,15 +34,29 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
...
+
+# Copy the XML files over to the build directory
+file(GLOB_RECURSE XML_FILES ${CMAKE_CURRENT_SOURCE_DIR}/*.xml)
```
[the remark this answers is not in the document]
EduPonz: Is this addressed?

EduPonz (changes requested):
We're definitely getting there! I like the idea of the classes not having any static members.

Now, I have a "crazy" idea to simplify the main and to also group some of the subscribers' code. Right now, all three entities have similar public API:
1. A ctor
2. A `run` member function with no args
3. A `stop` member function with no args

We can do some hierarchy of classes:

```mermaid
classDiagram
class Entity {
    +run() void
    +stop() void
   +create_entity(const hello_world_config& config)$ std::shared_ptr~Entity~
}
Entity <|-- Publisher
Entity <|-- AbstractSubscriber : Optional grouping of subscriber code
AbstractSubscriber <|-- Subscriber
AbstractSubscriber <|-- WaitSubscriber

Entity ..> Publisher : creates
Entity ..> Subscriber : creates
Entity ..> WaitSubscriber : creates
```

We would have a base class for the entities which can also act as a factory of the specializations via an static method. This way you don't need all those pointers, just the one, nor do you need any switches. You'd just create the entity passing the config and then spawn the thread to run it's run method. Since the entity is a `shared_ptr`, you wouldn't need to delete it either. You can see a PoC [here](https://godbolt.org/z/WrKMnqbhx).

Mario-DL:
> We're definitely getting there! I like the idea of the classes not having any static members.
> 
> Now, I have a "crazy" idea to simplify the main and to also group some of the subscribers' code. Right now, all three entities have similar public API:
> 
> 1. A ctor
> 2. A `run` member function with no args
> 3. A `stop` member function with no args
> 
> We can do some hierarchy of classes:
> 
> ```mermaid
> classDiagram
> class Entity {
>     +run() void
>     +stop() void
>    +create_entity(const hello_world_config& config)$ std::shared_ptr~Entity~
> }
> Entity <|-- Publisher
> Entity <|-- AbstractSubscriber : Optional grouping of subscriber code
> AbstractSubscriber <|-- Subscriber
> AbstractSubscriber <|-- WaitSubscriber
> 
> Entity ..> Publisher : creates
> Entity ..> Subscriber : creates
> Entity ..> WaitSubscriber : creates
> ```
> We would have a base class for the entities which can also act as a factory of the specializations via an static method. This way you don't need all those pointers, just the one, nor do you need any switches. You'd just create the entity passing the config and then spawn the thread to run it's run method. Since the entity is a `shared_ptr`, you wouldn't need to delete it either. You can see a PoC [here](https://godbolt.org/z/WrKMnqbhx).

(NIT) Apart from the internally agreed things I would call the creation method (or global free function) `make_app()` instead of `create` :smiley:

examples/cpp/hello_world/CLIParser.hpp
JesusPoderoso: I guess that we should be more explicit with unsigned options. Updating the usage and input parsing / check.

examples/cpp/hello_world/CMakeLists.txt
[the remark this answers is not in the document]
JesusPoderoso: It has been finally addressed in [c139dce](https://github.com/eProsima/Fast-DDS/pull/4547/commits/c139dce8ce425a26fc998dbecc9bb5fdc5f5e89e).

Current path: 
colcon_ws
├ install
│      ├ fastdds
│      │      ├ share
│      │      │      ├ fastdds
│      │      │      │      ├ examples
│      │      │      │      │      ├ cpp
│      │      │      │      │      │      ├ hello_world
│      │      │    ...     ...    ...      ...      
│      │      └ bin ...
│      ├ fastcdr
...     ...

examples/cpp/hello_world/HelloWorld_main.cpp
```diff
@@ -0,0 +1,204 @@
...
+                break;
+            case CLIParser::entity_kind::SUBSCRIBER:
+            default:
+                std::cout << entity_name << " running until " << samples << " samples have been received. Please press "
+                          << "Ctrl+C to stop the " << entity_name << " at any time." << std::endl;
+                break;
+        }
+    }
+
+    signal_handler = [&](std::string signal)
```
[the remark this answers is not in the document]
JesusPoderoso: I am keeping the `std:function` because handler registration fails:
`error: cannot convert ‘main(int, char**)::<lambda(int)>’ to ‘__sighandler_t’ {aka ‘void (*)(int)’}`

I've improved anyways its usage in [fbb9c0e](https://github.com/eProsima/Fast-DDS/pull/4547/commits/fbb9c0e5219cb41c9fdd6dc1df030640540adfb6)

examples/cpp/hello_world/main.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+int main(
+        int argc,
+        char** argv)
+{
+    auto ret = EXIT_SUCCESS;
+    const std::string topic_name = "hello_world_topic";
+    CLIParser::hello_world_config config = CLIParser::parse_cli_options(argc, argv);
+    uint16_t samples = config.entity == CLIParser::EntityKind::PUBLISHER ? config.pub_config.samples :
+            config.entity == CLIParser::EntityKind::SUBSCRIBER ? config.sub_config.samples : 0;
```
EduPonz, line 49: I think two ternary operators make the code difficult to follow. I think an `if-else if-else` would be more readable

examples/cpp/hello_world/CMakeLists.txt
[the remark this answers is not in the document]
EduPonz: Not completely. Although when compiling Fast DDS with `INSTALL_EXAMPLES` the XML do get installed, they won't if I'm compiling this as a stand alone.

examples/cpp/hello_world/main.cpp
```diff
@@ -0,0 +1,103 @@
...
+            case CLIParser::EntityKind::PUBLISHER:
+                std::cout << app_name << " running " << samples << " samples. Please press Ctrl+C to stop the "
+                          << app_name << " at any time." << std::endl;
+                break;
+            case CLIParser::EntityKind::SUBSCRIBER:
+            default:
+                std::cout << app_name << " running until " << samples << " samples have been received. Please press "
+                          << "Ctrl+C to stop the " << app_name << " at any time." << std::endl;
+                break;
+        }
```
EduPonz, line 83:
```suggestion
        std::cout << app_name << " running for " << samples << " samples. Please press Ctrl+C to stop the " << app_name << " at any time." << std::endl;
```

examples/cpp/hello_world/CLIParser.hpp
EduPonz: Not applied

EduPonz (changes requested):
This is turning out nicely, great job! 🤖

EduPonz (changes requested):

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+## Introduction
+
+This example is part of the suite of examples designed by eProsima that aims to illustrate the features and possible configurations of DDS deployments through *eProsima Fast DDS*.
+
+In this case, the *hello world* example describes the simplest deployment of a Fast DDS publisher and subscriber.
+
+## Example in deep
+
+Each entity of the example (publisher and subscriber) creates different nested DDS entities: domain participant, publisher & dataWriter, and subscriber & dataReader, respectively.
```
EduPonz, line 13:
```suggestion
Each entity of the example (publisher and subscriber) creates different nested DDS entities: domain participant, publisher, and data writer; and subscriber, and data reader, respectively.
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+In both cases, the three DDS entities (domain participant, publisher/subscriber and dataWriter/dataReader) load their default configuration from the environment.
```
EduPonz, line 14:
```suggestion
In both cases, the three DDS entities (domain participant, publisher/subscriber and data writer/data reader) load their default configuration from the environment.
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+If the environment does not specify the expected configuration, they take the default configuration per entity.
+For further information regarding the configuration environment, please refer to the *[XML profile playground](#xml-profile-playground)* subsection in the *[advanced configuration](#advanced-configuration)* section.
+
+The *hello world* example, together with the remain examples, would include a listening callback on the subscriber side. The subscriber will manage the new available data in the same thread as the main subscriber application.
+For simplicity, the ``HelloWorldSubscriber`` implements it own callback.
+
+Furthermore, this particular example includes a wait-set implementation. In contrast to implementing a listening callback, the wait-set is a mechanism where a dedicated thread waits until a status condition occurs. In that moment, that status condition triggering event would be evaluated to determine witch actions should be taken against it.
```
EduPonz, line 21: This needs to be rephrased. I would not refer to the rest of the examples here, just to this one. I'd say that it can be run with two different subscription paradigms, and then explain the two

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+For this example, both listening callback and wait-set implementation would run the same code and generate the same output for both triggering events: subscription matching and new data available.
```
EduPonz, line 23: Although similar, they do not run the same code

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
+# Hello world example
+
+The *eProsima Fast DDS hello world* example is a simple application intended to demonstrate a basic DDS deployment.
+
+## Introduction
```
EduPonz, line 5:
```suggestion
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+This example is part of the suite of examples designed by eProsima that aims to illustrate the features and possible configurations of DDS deployments through *eProsima Fast DDS*.
+
+In this case, the *hello world* example describes the simplest deployment of a Fast DDS publisher and subscriber.
+
```
EduPonz, line 10: I'm missing an index here

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+## Example in deep
```
EduPonz, line 11: I don't think this is the correct formulation

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+The *hello world* example, together with the remain examples, would include a listening callback on the subscriber side. The subscriber will manage the new available data in the same thread as the main subscriber application.
+For simplicity, the ``HelloWorldSubscriber`` implements it own callback.
+
+Furthermore, this particular example includes a wait-set implementation. In contrast to implementing a listening callback, the wait-set is a mechanism where a dedicated thread waits until a status condition occurs. In that moment, that status condition triggering event would be evaluated to determine witch actions should be taken against it.
+
+For this example, both listening callback and wait-set implementation would run the same code and generate the same output for both triggering events: subscription matching and new data available.
+
+## Run the example
+
+To launch this test open two different consoles. One of them will run the publisher side of the example, and the other would run the subscriber side.
```
EduPonz, line 27:
```suggestion
To launch this example, two different terminals are required.
One of them will run the publisher side of the example, and the other will run the subscriber side.
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+user@machine:example_path$ ./hello_world publisher
+Publisher running. Please press Ctrl+C to stop the Publisher at any time.
+```
+
+#### Windows
+
+```powershell
+example_path> hello_world.exe publisher
+Publisher running. Please press Ctrl+C to stop the Publisher at any time.
+```
```
EduPonz, line 43: I'd make this a bullet point list, same for sub

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+```powershell
+example_path> hello_world.exe subscriber
+Subscriber running. Please press Ctrl+C to stop the Subscriber at any time.
+```
+
+## Expected output
+
+It does not matter which entity is launched first, because the publisher has a condition to wait until the first match to start sending hello world samples.
+The expected output on both cases is a first displayed message acknowledging they have matched, followed by the amount of samples sent or received until user press Ctrl+C.
```
EduPonz, line 64:
```suggestion
Regardless of which application is run first, since the published will not start sending that until a subscriber is discovered, the expected output both for publishers and subscriber is a first displayed message acknowledging the match, followed by the amount of samples sent or received until Ctrl+C is pressed
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+```shell
+Subscriber running. Please press Ctrl+C to stop the Subscriber at any time.
+Subscriber matched.
+Message: 'Hello world' with index: '1' RECEIVED
+Message: 'Hello world' with index: '2' RECEIVED
+Message: 'Hello world' with index: '3' RECEIVED
+...
+```
+
+When pressed Ctrl+C to stop one of the applications, the other will detect the unmatched status, and will display an informative message.
```
EduPonz, line 88:
```suggestion
When Ctrl+C is pressed to stop one of the applications, the other one will show the unmatched status, displaying an informative message.
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Hello world subscriber
+
+```shell
+...
+Message: 'Hello world' with index: '9' RECEIVED
+Message: 'Hello world' with index: '10' RECEIVED
+Message: 'Hello world' with index: '11' RECEIVED
+user@machine:example_path$ |
+```
```
EduPonz, line 112:
```suggestion
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+Subscriber running. Please press Ctrl+C to stop the Subscriber at any time.
+Subscriber matched.
+Message: 'Hello world' with index: '1' RECEIVED
+Message: 'Hello world' with index: '2' RECEIVED
+Message: 'Hello world' with index: '3' RECEIVED
+...
+```
+
+When pressed Ctrl+C to stop one of the applications, the other will detect the unmatched status, and will display an informative message.
+This output example represents stopping the subscriber application.
```
EduPonz, line 89:
```suggestion
The following is a possible output of the publisher application when stopping the subscriber one.
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Hello world publisher
```
EduPonz, line 91:
```suggestion
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+```shell
+...
+Message: 'Hello world' with index: '9' RECEIVED
+Message: 'Hello world' with index: '10' RECEIVED
+Message: 'Hello world' with index: '11' RECEIVED
+user@machine:example_path$ |
+```
+
+Stopping the subscriber application does not make the publisher stop sending data. But in the opposite case, stopping the publisher application will make subscriber stop receiving data.
```
EduPonz, line 114: As far as I understand, if there are no matched subs, the publisher does stop sending data

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Hello world subscriber
+
+```shell
+...
+Message: 'Hello world' with index: '9' RECEIVED
+Message: 'Hello world' with index: '10' RECEIVED
+Message: 'Hello world' with index: '11' RECEIVED
+Subscriber unmatched.
+```
```
EduPonz, line 134:
```suggestion
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+```shell
+user@machine:example_path$ ./hello_world subscriber --waitset
+```
+
+#### Windows
+
+```powershell
+example_path> hello_world.exe subscriber --waitset
+```
```
EduPonz, line 152: Bullet list

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+```shell
+...
+Message: 'Hello world' with index: '9' RECEIVED
+Message: 'Hello world' with index: '10' RECEIVED
+Message: 'Hello world' with index: '11' RECEIVED
+Subscriber unmatched.
+```
+
+## Advanced configuration
```
EduPonz, line 136: I don't think we need this title. You'd need to adjust the level of the rest of titles
```suggestion
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+#### Windows
+
+```powershell
+example_path> hello_world.exe subscriber --waitset
+```
+
+The expected output is exactly the same as the described in the *[previous](#expected-output)* section.
+
+
+All the example available flags can be queried running the executable with the ``-h`` or ``--help`` flag.
```
EduPonz, line 157: This could have come way earilier

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+### XML profile playground
+
+The *eProsima Fast DDS* entities can be configured through an XML profile from the environment by adding a reference the XML profiles file setting the environment variable ``FASTDDS_DEFAULT_PROFILES_FILE`` to its path.
```
EduPonz, line 161: Please rephrase this

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+```shell
+user@machine:example_path$ export FASTDDS_DEFAULT_PROFILES_FILE=hello_world_profile.xml
+```
+
+#### Windows
+
+```powershell
+example_path> set FASTDDS_DEFAULT_PROFILES_FILE=hello_world_profile.xml
+```
```
EduPonz, line 173: Bullet list

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+The example contains a XML profiles files with certain QoS:
```
EduPonz, line 175:
```suggestion
The example provides with an XML profiles files with certain QoS:
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+- Reliable reliability: avoid sample loss.
+- Transient local durability: enable late-join-participants to receive previous samples if connection was lost.
```
EduPonz, line 178:
```suggestion
- Transient local durability: enable late-join subscriber applications to receive previous samples.
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+- Keep-last history with high depth: ensure certain amount of previous samples for late-joiners.
+
+Applying different configurations to the entities would change the sample management behavior, among other configurations, but in any case these configurations would affect in the behavior of the *hello_world* example.
+The expected output would be exactly the same as launching it with no environment configuration.
+
+Try your own XML profile to see how your configuration affects the *hello world* example communication.
```
EduPonz, line 184:
```suggestion
```

examples/cpp/hello_world/README.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+The example contains a XML profiles files with certain QoS:
+
+- Reliable reliability: avoid sample loss.
+- Transient local durability: enable late-join-participants to receive previous samples if connection was lost.
+- Keep-last history with high depth: ensure certain amount of previous samples for late-joiners.
+
+Applying different configurations to the entities would change the sample management behavior, among other configurations, but in any case these configurations would affect in the behavior of the *hello_world* example.
+The expected output would be exactly the same as launching it with no environment configuration.
```
EduPonz, line 182: This reads weird

examples/cpp/hello_world/PublisherApp.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+    TypeSupport type_;
+
+    int16_t matched_;
+
+    uint16_t samples_;
+
+    std::mutex mutex_;
+
+    std::condition_variable matched_cv_;
```
EduPonz, line 87:
```suggestion
    std::condition_variable cv_;
```

EduPonz (changes requested):

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
+In both cases, the three DDS entities (domain participant, publisher/subscriber and data writer/data reader) load their default configuration from the environment.
 If the environment does not specify the expected configuration, they take the default configuration per entity.
-For further information regarding the configuration environment, please refer to the *[XML profile playground](#xml-profile-playground)* subsection in the *[advanced configuration](#advanced-configuration)* section.
+For further information regarding the configuration environment, please refer to the *[XML profile playground](#xml-profile-playground)* section.
 
-The *hello world* example, together with the remain examples, would include a listening callback on the subscriber side. The subscriber will manage the new available data in the same thread as the main subscriber application.
-For simplicity, the ``HelloWorldSubscriber`` implements it own callback.
+This particular example includes two different subscription paradigms: listening callbacks and wait-sets:
```
EduPonz, line 22:
```suggestion
This particular example includes two different subscription paradigms; i.e. listening callbacks and wait-sets:
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
 
-Furthermore, this particular example includes a wait-set implementation. In contrast to implementing a listening callback, the wait-set is a mechanism where a dedicated thread waits until a status condition occurs. In that moment, that status condition triggering event would be evaluated to determine witch actions should be taken against it.
+* The listening callback mechanism consists on declaring a listener class and attach it to the data reader.
```
EduPonz, line 24:
```suggestion
* The listening callback mechanism consists on declaring a listener class and attaching it to the data reader.
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
-## Introduction
-
 This example is part of the suite of examples designed by eProsima that aims to illustrate the features and possible configurations of DDS deployments through *eProsima Fast DDS*.
 
 In this case, the *hello world* example describes the simplest deployment of a Fast DDS publisher and subscriber.
 
-## Example in deep
+* [Description of the example](#description-of-the-example)
+* [Run the example](#run-the-example)
+* [Expected output](#expected-output)
```
EduPonz, line 11:
```suggestion
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
 
-The *hello world* example, together with the remain examples, would include a listening callback on the subscriber side. The subscriber will manage the new available data in the same thread as the main subscriber application.
-For simplicity, the ``HelloWorldSubscriber`` implements it own callback.
+This particular example includes two different subscription paradigms: listening callbacks and wait-sets:
 
-Furthermore, this particular example includes a wait-set implementation. In contrast to implementing a listening callback, the wait-set is a mechanism where a dedicated thread waits until a status condition occurs. In that moment, that status condition triggering event would be evaluated to determine witch actions should be taken against it.
+* The listening callback mechanism consists on declaring a listener class and attach it to the data reader.
+When the data reader is triggered by an event, it runs the listener's method associated to that event, as a callback.
+  For simplicity, in this example the subscriber class inherits the listener class, and it implements it own callback.
```
EduPonz, line 26:
```suggestion
  For simplicity, in this example, the subscriber class inherits from the listener class, overriding the corresponding callback.
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
+For further information regarding the configuration environment, please refer to the *[XML profile playground](#xml-profile-playground)* section.
 
-The *hello world* example, together with the remain examples, would include a listening callback on the subscriber side. The subscriber will manage the new available data in the same thread as the main subscriber application.
-For simplicity, the ``HelloWorldSubscriber`` implements it own callback.
+This particular example includes two different subscription paradigms: listening callbacks and wait-sets:
 
-Furthermore, this particular example includes a wait-set implementation. In contrast to implementing a listening callback, the wait-set is a mechanism where a dedicated thread waits until a status condition occurs. In that moment, that status condition triggering event would be evaluated to determine witch actions should be taken against it.
+* The listening callback mechanism consists on declaring a listener class and attach it to the data reader.
+When the data reader is triggered by an event, it runs the listener's method associated to that event, as a callback.
```
EduPonz, line 25:
```suggestion
  When the data reader is triggered by an event, it runs the listener's method associated to that event, as a callback.
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
+  For simplicity, in this example the subscriber class inherits the listener class, and it implements it own callback.
 
-For this example, both listening callback and wait-set implementation would run the same code and generate the same output for both triggering events: subscription matching and new data available.
+* The wait-set is a mechanism where a dedicated thread waits until a status condition occurs.
+In that moment, that status condition triggering event would be evaluated to determine witch actions should be taken against it.
```
EduPonz, line 29:
```suggestion
  In that moment, that status condition triggering event would be evaluated to determine witch actions should be taken against it.
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
+
+For this example, both listening callback and wait-set implementation would run similar code and generate equivalent output for both triggering events: subscription matching and new data available.
 
 ## Run the example
 
-To launch this test open two different consoles. One of them will run the publisher side of the example, and the other would run the subscriber side.
+To launch this example, two different terminals are required.
+One of them will run the publisher example application, and the other would run the subscriber application.
```
EduPonz, line 36:
```suggestion
One of them will run the publisher example application, and the other will run the subscriber application.
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
+    Subscriber running. Please press Ctrl+C to stop the Subscriber at any time.
+    ```
+
+All the example available flags can be queried running the executable with the ``-h`` or ``--help`` flag.
 
 ## Expected output
 
-It does not matter which entity is launched first, because the publisher has a condition to wait until the first match to start sending hello world samples.
-The expected output on both cases is a first displayed message acknowledging they have matched, followed by the amount of samples sent or received until user press Ctrl+C.
+Regardless of which application is run first, since the published will not start sending that until a subscriber is discovered, the expected output both for publishers and subscriber is a first displayed message acknowledging the match, followed by the amount of samples sent or received until Ctrl+C is pressed.
```
EduPonz, line 74:
```suggestion
Regardless of which application is run first, since the publisher will not start sending data until a subscriber is discovered, the expected output both for publishers and subscribers is a first displayed message acknowledging the match, followed by the amount of samples sent or received until Ctrl+C is pressed.
```

EduPonz, line 74:
```suggestion
Regardless of which application is run first, since the publisher will not start sending data until a subscriber is discovered, the expected output both for publishers and subscribers is a first displayed message acknowledging the match, followed by the amount of samples sent or received until Ctrl+C is pressed.
```

examples/cpp/hello_world/README.md
```diff
@@ -85,100 +95,58 @@ Message: 'Hello world' with index: '3' RECEIVED
...
 
+## XML profile playground
 
-All the example available flags can be queried running the executable with the ``-h`` or ``--help`` flag.
-
-### XML profile playground
-
-The *eProsima Fast DDS* entities can be configured through an XML profile from the environment by adding a reference the XML profiles file setting the environment variable ``FASTDDS_DEFAULT_PROFILES_FILE`` to its path.
+The *eProsima Fast DDS* entities can be configured through an XML profile from the environment.
+It is performed by setting the environment variable ``FASTDDS_DEFAULT_PROFILES_FILE`` with the path to the XML profiles file:
```
EduPonz, line 131:
```suggestion
This is accomplished by setting the environment variable ``FASTDDS_DEFAULT_PROFILES_FILE`` to path to the XML profiles file:
```

examples/cpp/hello_world/README.md
```diff
@@ -85,100 +95,58 @@ Message: 'Hello world' with index: '3' RECEIVED
...
-- Transient local durability: enable late-join-participants to receive previous samples if connection was lost.
+- Transient local durability: enable late-join subscriber applications to receive previous samples.
 - Keep-last history with high depth: ensure certain amount of previous samples for late-joiners.
 
-Applying different configurations to the entities would change the sample management behavior, among other configurations, but in any case these configurations would affect in the behavior of the *hello_world* example.
-The expected output would be exactly the same as launching it with no environment configuration.
-
-Try your own XML profile to see how your configuration affects the *hello world* example communication.
+Applying different configurations to the entities will change to a greater or lesser extent how the application behaves in relation to sample management.
+In any case these settings will affect the behavior of the sample itself: the output of the sample will be the same regardless of the configuration (or lack of configuration) provided by the environment variable.
```
EduPonz, line 152:
```suggestion
Even when these settings affect the behavior of the sample management, the applications' output will be the similar.
```

examples/cpp/hello_world/README.md
```diff
@@ -2,66 +2,76 @@
...
-Subscriber running. Please press Ctrl+C to stop the Subscriber at any time.
-```
+    ```powershell
+    example_path> hello_world.exe subscriber
+    Subscriber running. Please press Ctrl+C to stop the Subscriber at any time.
+    ```
+
+All the example available flags can be queried running the executable with the ``-h`` or ``--help`` flag.
 
 ## Expected output
```
EduPonz, line 72:
```suggestion
### Expected output
```

EduPonz (changes requested):

EduPonz (approved):

EduPonz (approved):

EduPonz (changes requested):

EduPonz (approved):

EduPonz (approved):

JesusPoderoso:

JesusPoderoso:

JesusPoderoso:

JesusPoderoso:

JesusPoderoso:
@richiprosima please test_3 this

JesusPoderoso:

JesusPoderoso:
@richiprosima please test_3 this

JesusPoderoso:
I would like to apply some @Mario-DL's suggestions in the configuration example in this example too (as including doxygen comments or re-ordering a couple of things).
Marking this PR with the label `to-do` to apply them ASAP

JesusPoderoso:
@richiprosima please test_3 this

JesusPoderoso:
CI issues are unrelated to the PR.
Ready to merge! @EduPonz

test/examples/CMakeLists.txt
```diff
@@ -56,6 +58,13 @@ else()
     message(FATAL_ERROR "Unsupported platform")
 endif()
 
+# Configure TinyXML2 library path if installed in user library path
+if(NOT (TINYXML2_FROM_SOURCE OR TINYXML2_FROM_THIRDPARTY))
+    get_filename_component(TINYXML2_LIB_DIR ${TINYXML2_LIBRARY} DIRECTORY)
+    set(TINYXML2_LIB_DIR_COMPOSE_VOLUME "- ${TINYXML2_LIB_DIR}:/usr/local/shared/fastdds:ro")
+    set(TINYXML2_LIB_DIR_COMPOSE_LD_LIBRARY_PATH ":/usr/local/shared/fastdds")
```
EduPonz, line 65:
```suggestion
    set(TINYXML2_LIB_DIR_COMPOSE_VOLUME "- ${TINYXML2_LIB_DIR}:${CMAKE_INSTALL_PREFIX}/${DATA_INSTALL_DIR}/fastdds:ro")
    set(TINYXML2_LIB_DIR_COMPOSE_LD_LIBRARY_PATH ":${CMAKE_INSTALL_PREFIX}/${DATA_INSTALL_DIR}/fastdds")
```

EduPonz (changes requested):

EduPonz (approved):

PR #4554, opened 2024-03-13
[20608][20610] Address XMLProfiles fuzzer regressions

Mario-DL:
## Description

This PR addresses the regressions described in [20608](https://eprosima.easyredmine.com/issues/20608) and [20610](https://eprosima.easyredmine.com/issues/20610). One has to  do with catching an out of bounds exception and the other with fixing the regex pattern.

## Contributor Checklist
- [X] Changes are ABI compatible. 
    
## Reviewer Checklist

### DIFF
--- a/src/cpp/utils/UnitsParser.cpp
+++ b/src/cpp/utils/UnitsParser.cpp
@@ -64,7 +64,17 @@ uint32_t parse_value_and_units(
         {"GIB", 1024 * 1024 * 1024},
     };
 
-    uint64_t num = std::stoull(value);
+    uint64_t num = 0;
+    try
+    {
+        num = std::stoull(value);
+    }
+    catch (std::out_of_range&)
+    {
+        throw std::invalid_argument("Failed to parse value from string." \
+                      " The number is too large to be converted to bytes (Max: (2^32)-1 Bytes).");
+    }
+
     to_uppercase(units);
 
     std::regex pattern(R"(([kibmgKIBMG]{1,3})|(\d+))");
--- a/src/cpp/utils/UnitsParser.cpp
+++ b/src/cpp/utils/UnitsParser.cpp
@@ -77,7 +77,7 @@ uint32_t parse_value_and_units(
 
     to_uppercase(units);
 
-    std::regex pattern(R"(([kibmgKIBMG]{1,3})|(\d+))");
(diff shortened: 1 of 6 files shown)


src/cpp/utils/UnitsParser.cpp
```diff
@@ -77,7 +77,7 @@ uint32_t parse_value_and_units(
 
     to_uppercase(units);
 
-    std::regex pattern(R"(([kibmgKIBMG]{1,3})|(\d+))");
+    std::regex pattern(R"(([bB]{1})|([kibmgKIBMG]{2,3})|(\d+))");
```
MiguelCompany, line 80: 1. We already converted to uppercase, so we don't need the lowercase letters.
2. The 'at()` below is doing a search in the map (i.e. performing string comparisons) so, in a way, it is doing the job of the regex again.

This means we can remove the regex altogether, and just capture `out_of_range` around the `at()` operation, throwing the `invalid_argument` we have in line 83

An additional update we could make is adding the empty string to the map, with the same value we have for `"B"`, and remove the conditionals in the places where `parse_value_and_units` is being called

MiguelCompany (changes requested):

src/cpp/utils/UnitsParser.cpp
Mario-DL: I like this suggestion

src/cpp/utils/UnitsParser.cpp
```diff
@@ -64,16 +65,29 @@ uint32_t parse_value_and_units(
...
+    {
+        throw std::invalid_argument("Failed to parse value from string." \
+                      " The number is too large to be converted to bytes (Max: (2^32)-1 Bytes).");
+    }
+
     to_uppercase(units);
 
-    std::regex pattern(R"(([kibmgKIBMG]{1,3})|(\d+))");
-    if (!std::regex_match(units, pattern))
+    unsigned int magnitude = 0;
```
MiguelCompany, line 81:
```suggestion
    uint32_t magnitude = 0;
```

MiguelCompany (changes requested):

MiguelCompany (approved):

Mario-DL:

PR #4583, opened 2024-03-19
[20641] Explicitly pass vendor ID to readFromCdrMessage

EduPonz:
## Description

With this PR, the incoming vendor ID is explicitly passed to the `readFromCdrMessage` function. By doing so, we avoid using the default value (Fast DDS), which can cause deserialization errors in case the incoming data uses a custom PID within the parameter list that we also happen to use for a different purpose. The PR adds a regression test, since CycloneDDS also uses PID 0x8007, but with a different content. Since the default vendor ID was being used, the deserialization of CycloneDDS Data(p) was failing.

## Contributor Checklist

## Reviewer Checklist


EduPonz:

test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -2131,3 +2141,69 @@ TEST(Discovery, MultipleXMLProfileLoad)
...
+
+    /* Disable builtin transport and add custom one */
+    DomainParticipantQos participant_qos = PARTICIPANT_QOS_DEFAULT;
+    participant_qos.transport().use_builtin_transports = false;
+    participant_qos.transport().user_transports.clear();
+    participant_qos.transport().user_transports.push_back(transport);
+
+    /* Create participant with custom transport and listener */
+    DiscoveryListener listener;
+    uint32_t domain_id = static_cast<uint32_t>(GET_PID()) % 230;
```
JesusPoderoso, line 2194: Shouldn't be the same Domain Id as the one in the datagram?

EduPonz: Yes and no, Right now, Fast DDS does not use the incoming `PID_DOMAIN_ID` for participant discovery. IMO the way to go here would be to craft a Data(p) with only the minimum set of PIDs (domain ID is not included in that set) and the offending ones. That way we'd verify exactly what we want. I decided to randomize the domain here to avoid discovering participants from elsewhere.

For 3.0, we will honor the `PID_DOMAIND_ID`, and since this is a hotfix, I've decided to do what you suggest in a follow up then.

JesusPoderoso (dismissed):
LGTM

EduPonz:

JesusPoderoso (approved):

PR #4685, opened 2024-04-15
[20732] Improve xml loading method in fuzz_XMLProfiles

elianalf:
## Description

This PR improves the XML loading process in fuzz_XMLProfiles.cxx. `LLVMFuzzerTestOneInput()` accepts a XML array as argument, so this PR replaces `loadXMLFile()`, which requires copying the XML array to a file, with `load_XML_profiles_string()`, which directly takes an array.

## Contributor Checklist

- N/A Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- N/A Any new/modified methods have been properly documented using Doxygen. 
- N/A Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
- N/A New feature has been added to the `versions.md` file (if applicable).
- N/A New feature has been documented/Current behavior is correctly described in the documentation. 
    
- N/A Applicable backports have been included in the description.

## Reviewer Checklist


fuzz/C++/fuzz_XMLProfiles/fuzz_XMLProfiles.cxx
```diff
@@ -1,5 +1,5 @@
 #include <fastrtps/Domain.h>
```
MiguelCompany, line 1: I think we can remove this line

fuzz/C++/fuzz_XMLProfiles/fuzz_XMLProfiles.cxx
```diff
@@ -23,18 +23,7 @@ extern "C" int LLVMFuzzerTestOneInput(
...
-    {
-        return EXIT_FAILURE;
-    }
-
-    // TODO change this to a func. taking buf + len (or C string)
-    // to avoid using `buf_to_file`
-    xmlparser::XMLProfileManager::loadXMLFile(filename);
-
-    if (delete_file(filename) != 0)
+    if (ReturnCode_t::RETCODE_OK != fastdds::dds::DomainParticipantFactory::get_instance()->load_XML_profiles_string(reinterpret_cast<const char*>(data), size))
```
MiguelCompany, line 26: We can remove `buf_to_file` and `delete_file` from fuzz_utils

MiguelCompany (changes requested):

fuzz/C++/fuzz_XMLProfiles/fuzz_XMLProfiles.cxx
```diff
@@ -23,18 +22,7 @@ extern "C" int LLVMFuzzerTestOneInput(
...
     {
         return EXIT_FAILURE;
     }
```
MiguelCompany, line 28: I don't think we need to return with failure if the parser correctly identified an error
```suggestion
    fastdds::dds::DomainParticipantFactory::get_instance()->load_XML_profiles_string(reinterpret_cast<const char*>(data), size)
```

MiguelCompany (changes requested):

fuzz/C++/fuzz_XMLProfiles/fuzz_XMLProfiles.cxx
elianalf: The logic is the same as before. I think it makes sense that `LLVMFuzzerTestOneInput()` return an error if it is not able to load or parse the XML. Do you think would it be better if we return ERROR only if `==RETCODE_ERROR`? In all the other cases we would still fail in parsing the XML, but it wouldn't return as error.

MiguelCompany: The behavior has changed. Previously, the return of `xmlparser::XMLProfileManager::loadXMLFile()` was not being checked. The purpose of fuzzers is fail if and only if an uncontrolled situation is reached (i.e. unhandled exception, access violations, etc.). They will be constantly trying with wrong inputs to check if the library crashes. If `load_XML_profiles_string` returns then the fuzzer succeeded, no matter which code it is returning

elianalf: It actually returned error if there was an error in creating the file or in deleting it. Ok, if you think it's outside the scope of the method informing about parsing failure, I will always return true

MiguelCompany (approved):

PR #4835, opened 2024-05-23
[20815] Only apply content filter to ALIVE changes

EduPonz:
## Description

This PR fixes a bug that caused the content filter to also be applied to `unregister` and `disposed` samples. Since in those messages the only fields populated (if any) are the ones annotated with `@key`, the `unregister` and `dispose` samples did not pass the filter (in general) and thus were being discarded. This caused several issues:

1. When writing to more than 10 instances with default Qos (keep last 1, max_instances 10), only the first 10 samples were been received.
2. In best effort, when setting the history depth and max instances appropriately, all samples were received, but calls to `unregister` or `dispose` followed by a `write` were triggering `sample_lost` events, as the received sequence numbers were not consecutive (because of the filtering out of the `unregister`/`dispose`).

This PR fixes these issues by only querying for sample relevance when the `CacheChange` kind is `ALIVE`.

@EduPonz backport 2.14.x 2.13.x 2.10.x 2.6.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/reader/reader_utils.hpp
+++ b/src/cpp/rtps/reader/reader_utils.hpp
@@ -35,14 +35,14 @@ using GUID = fastrtps::rtps::GUID_t;
  * @brief Check if a change is relevant for a reader.
  *
  * @param change The CacheChange_t to be evaluated.
- * @param reader_guid Remote reader GUID_t.
+ * @param reader_guid Reader's GUID_t.
  * @param filter The IReaderDataFilter to be used.
  *
  * @return true if relevant, false otherwise.
  */
 bool change_is_relevant_for_filter(
         const CacheChange& change,
-        const GUID& guid,
+        const GUID& reader_guid,
         const IReaderDataFilter* filter);
 
 } // namespace rtps
--- a/src/cpp/rtps/reader/reader_utils.hpp
+++ b/src/cpp/rtps/reader/reader_utils.hpp
@@ -1,0 +1,53 @@
+// Copyright 2024 Proyectos y Sistemas de Mantenimiento SL (eProsima).
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * @file reader_utils.hpp
+ */
+
+#ifndef _FASTDDS_RTPS_READER_READERUTILS_H_
+#define _FASTDDS_RTPS_READER_READERUTILS_H_
+
+#include <fastdds/rtps/common/CacheChange.h>
+#include <fastdds/rtps/common/Guid.h>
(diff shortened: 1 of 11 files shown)


src/cpp/rtps/reader/reader_utils.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace rtps {
+
+using CacheChange = fastrtps::rtps::CacheChange_t;
+using GUID = fastrtps::rtps::GUID_t;
+
+/**
+ * @brief Check if a change is relevant for a reader.
+ *
+ * @param change The CacheChange_t to be evaluated.
+ * @param reader_guid Remote reader GUID_t.
```
Mario-DL, line 38: Either this suggestion or changing the argument to `reader_guid`
```suggestion
 * @param guid Remote reader GUID_t.
```

test/blackbox/common/DDSBlackboxTestsContentFilter.cpp
```diff
@@ -610,6 +610,188 @@ TEST(DDSContentFilter, CorrectlyHandleAliasOtherHeader)
...
+ * Regression test for https://eprosima.easyredmine.com/issues/20815
+ * Check that the content filter is only applied to alive changes.
+ * The test creates a reliable writer and a reader with a content filter that only accepts messages with a specific
+ * string. After discovery, the writer sends 10 samples which pass the filer in 10 different instances, with the
+ * particularity that after each write, the instance is unregistered.
+ * The DATA(u) generated would not pass the filter if it was applied. To check that the filter is only applied to
+ * ALIVE changes (not unregister or disposed), the test checks that the reader receives 10 valid samples (one per
+ * sample sent) and 10 invalid samples (one per unregister). Furthermore, it also checks that no samples are lost.writer
+ */
+TEST(DDSContentFilter, OnlyFilterAliveChanges)
```
Mario-DL, line 623: I think it would be worth to try extending the already existing `PubSubReader` to accept `content_filter` stuff (also for future usage) and then inheriting from it to implementing test's logic  i.e `receive_one()` method could be overridden and the samples logic can be placed there. The `PubSubWriter` can be also used in the test. (e.g discovery, writer `matched_`). What do you think ? IMO it would give more simplicity instead of creating everything custom to this specific test (which I do not argue that it may be necessary if extending the `PubSubReader` and `PubSubWriter` gets too tangled). 

NIT, would it be more correct using `take_next_instance` given that the type is keyed ? Although I understand that the use-case scenario may be using `take_next_sample` and this test tries to mimic the same behavior.

Mario-DL (changes requested):
This PR brings an interesting `bugix`. Leaving a couple of suggestions

test/blackbox/common/DDSBlackboxTestsContentFilter.cpp
```diff
@@ -612,6 +618,91 @@ TEST(DDSContentFilter, CorrectlyHandleAliasOtherHeader)
...
+    reader.wait_discovery();
+
+    /* Send 10 samples, each on a different instance, unregistering instances after writing */
+    const size_t num_samples = 10;
+    reader.startReception(num_samples);
+
+    for (size_t i = 0; i < num_samples; ++i)
+    {
+        KeyedHelloWorld data;
+        data.key(i);
```
Mario-DL, line 689: I think we need to cast this to `uint16_t` to avoid windows compilation warning as error. Other than that, nice  test refactor, it gained readability, simplicity and prepared the `PubSub` classes for future CFT tests.

EduPonz:
Backport to 2.14.x:

* #4876 

The rest of the backport will be ordered from there.

PR #4920, opened 2024-06-10
Using socket to get IP instead of getifaddrs 

advancenXX:
I am using fastdds v2.7.0 on Android 12.
create_participant is invoked whenever my target network interface enabled, and I check it through some java framework api like,
        Enumeration<NetworkInterface> nifs = NetworkInterface.getNetworkInterfaces();
        while (nifs.hasMoreElements()) {
            NetworkInterface nif = nifs.nextElement();
        
            Enumeration<InetAddress> addresses = nif.getInetAddresses();
            while (addresses.hasMoreElements()) {
                InetAddress addr = addresses.nextElement();
        
                if (addr instanceof Inet4Address) {
                    String host = addr.getHostAddress();
                    // host equals("1.2.3.4")
                }
            }
        }

I use NetworkInterface to ensure my participant can find the target ip before it started, but I still find it has no ip sometimes according to tcpdump.
I searched from internet and found two ways to get ip,
 - getifaddrs, getnameinfo. which may miss if ip is assigned but no any actual socket IO;
 - socket. Android use this one, from InterfaceController::getCfg;

I am wondering if it can be used to get ip by socket instead of getifaddrs, as I have tested it on my own environment, especially for Android 12.

### DIFF
--- a/src/cpp/utils/IPFinder.cpp
+++ b/src/cpp/utils/IPFinder.cpp
@@ -157,70 +157,46 @@ bool IPFinder::getIPs(
         std::vector<info_IP>* vec_name,
         bool return_loopback)
 {
-    struct ifaddrs* ifaddr, * ifa;
-    int family, s;
-    char host[NI_MAXHOST];
+    int sockfd = 0;
+    unsigned int num = 0;
+    std::string ipv4addr;
+    struct ifreq buf[INET_ADDRSTRLEN];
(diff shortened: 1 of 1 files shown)


advancenXX:
hi @EduPonz , could you please review this PR, currently it is a draft patch to fix my problem only, but I can complete it if you think it is useful for Android 12.
Thanks

PR #5018, opened 2024-07-02
[21314] Improve resilience against clock adjustments

ma30002000:
## Description

When investigating a system undergoing larger system clock adjustments, I noticed that in some places, std::chrono::system_clock and std::chrono::high_resolution_clock are used for handling timeouts and status check intervals.
However, std::chrono::system_clock  is definitely not steady, and std::chrono::high_resolution_clock is not steady quite often (see [cppreference](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock)).

When undergoing clock adjustments (manually or due to clock server synchronization), timeouts and status checks might no longer be triggered when relying on timestamps based on std::chrono::system_clock.

@Mergifyio backport 2.14.x 2.10.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/transport/shared_mem/SharedMemGlobal.hpp
+++ b/src/cpp/rtps/transport/shared_mem/SharedMemGlobal.hpp
@@ -99,7 +99,7 @@ public:
 
     struct PortNode
     {
-        alignas(8) std::atomic<std::chrono::steady_clock::rep> last_listeners_status_check_time_ms;
+        alignas(8) std::atomic<std::chrono::high_resolution_clock::rep> last_listeners_status_check_time_ms;
         alignas(8) std::atomic<uint32_t> ref_counter;
 
         SharedMemSegment::Offset buffer;
@@ -324,17 +324,17 @@ public:
 
                 port_node->last_listeners_status_check_time_ms.exchange(
                     std::chrono::duration_cast<std::chrono::milliseconds>(
-                        std::chrono::steady_clock::now().time_since_epoch()).count());
+                        std::chrono::high_resolution_clock::now().time_since_epoch()).count());
 
                 return listeners_found == port_node->num_listeners;
             }
 
             void run()
             {
-                auto now = std::chrono::steady_clock::now();
+                auto now = std::chrono::high_resolution_clock::now();
 
                 auto timeout_elapsed = [](
-                    std::chrono::steady_clock::time_point& now,
+                    std::chrono::high_resolution_clock::time_point& now,
                     const PortContext& port_context)
                         {
                             return std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()).count()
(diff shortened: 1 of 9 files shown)


ma30002000:
Is there anything I need to to in order to get CI results?

JesusPoderoso:
Hi @ma30002000, thanks for your contribution.
We will look at it in the following days and order CI.
Thanks for your patience!

src/cpp/rtps/transport/shared_mem/SharedMemGlobal.hpp
```diff
@@ -99,7 +99,7 @@ class SharedMemGlobal
 
     struct PortNode
     {
-        alignas(8) std::atomic<std::chrono::high_resolution_clock::rep> last_listeners_status_check_time_ms;
+        alignas(8) std::atomic<std::chrono::steady_clock::rep> last_listeners_status_check_time_ms;
```
MiguelCompany, line 102: This is changing the behaviour on a structure that is used in the shared memory transport.

This would mean that current versions of the code would misinterpret the meaning of this field.

I suggest putting the changes on this file on a separate PR.

MiguelCompany (changes requested):
@ma30002000 

First of all, thank you for your patience waiting for this to be reviewed.

I took the liberty of rebasing this PR, and added a commit to fix one build issue.

Except for the changes in `SharedMemGlobal.hpp`, the changes look good to me.

src/cpp/rtps/transport/shared_mem/SharedMemGlobal.hpp
MiguelCompany: I went ahead and reverted the changes in commit f94d111b6b2af4af3427a233edc0e815fe888f39

MiguelCompany:
Manual CIs:
- [Ubuntu](https://github.com/eProsima/Fast-DDS/actions/runs/10627921613)
- [Windows](https://github.com/eProsima/Fast-DDS/actions/runs/10627928798)
- [MacOS](https://github.com/eProsima/Fast-DDS/actions/runs/10627925904)
- [Sanitizers](https://github.com/eProsima/Fast-DDS/actions/runs/10627939328)

src/cpp/utils/time_t_helpers.hpp
```diff
@@ -40,8 +40,12 @@ static void current_time_since_unix_epoch(
 {
     using namespace std::chrono;
 
+    static const auto init_time_since_epoch = system_clock::now().time_since_epoch();
+    static const auto init_steady_time = steady_clock::now();
+
     // Get time since epoch
-    auto t_since_epoch = system_clock::now().time_since_epoch();
+    auto t_elapsed = steady_clock::now() - init_steady_time;
+    auto t_since_epoch = init_time_since_epoch + t_elapsed;
```
Mario-DL, line 48: I'm not sure I understand this, we may have intended to write
`static const auto init_steady_time = steady_clock::now().time_since_epoch();`.
But even in that case, I dont fully understand how we are protecting against adjustments in the `system_clock` by adding up the elapsed time since the last reboot (the most common return across platforms for the steady_clock's `time_since_epoch()`).

Other than that, LGTM

MiguelCompany: We protect against adjustments in the `system_clock` by doing a single call to `system_clock::now()` the first time `current_time_since_unix_epoch` is called. We then return that initial reference plus the amount of `steady_clock` time since the first time the method was called.

The idea is to make `t_since_epoch` be the ``

Mario-DL: I overlooked the `static`. Thanks !

Mario-DL (approved):
LGTM

MiguelCompany:
@Mergifyio backport 2.14.x 2.10.x

ma30002000:
Hi, sorry for the late response, could you please provide some more details on why you have not included the change of std::chrono::high_resolution_clock to std::chrono::steady_clock? Using a (on some systems) non-steady clock for timeout checks will have undesired effects. I also do not see a drawback of using steady_clock instead.

MiguelCompany:
@ma30002000 Because the changes affected the PortNode structure, which is part of the Shared Memory port segment. This means the changes would break interoperability with previous versions, which we should not do on a patch release.

ma30002000:
@MiguelCompany I understand - how and when can we address the issue in the future then? Unfortunately this change did not make it into 3.x...

PR #5034, opened 2024-07-07
[21293] Fix destruction data-race on participant removal in intra-process

Mario-DL:
## Description

This PR addresses a race issue happening in stressed intraprocess scenarios when `EDP's` writer intends to use the remote local reader pointer of an already removed participant. This happens because the participant hasn't received the other's one disposal yet (as it goes through transport). 

Some [ci flaky tests](https://github.com/eProsima/Fast-DDS/actions/runs/10829350591/job/30046734211) have already been identified to be related with this issue.

The proposed solution introduces a new state in the Readers `LocalReaderViewStatus` in which the reader will notify that it is inactive as soon as it is destroyed and *noone* is using it.
On the other side, the remote local writers using pointers to it, now holds a `LocalReaderPointer` which wraps the raw reader's pointer plus the view. An internal counter now accounts for the number of references.

Thanks @MiguelCompany for helping with the final's solution design.

*Note: the test may be launched with `--restest-until-fail 20` or so, in order to reproduce the issue. For a more frequent failure, review can launch the `colcon test` with the `taskset -c 0,1` prefix to make the test to stress more and make it fail more frequently.* 

@Mergifyio backport 3.1.x 3.0.x 2.14.x 2.10.x

## Contributor Checklist

- [X] Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
    
- [X] Applicable backports have been included in the description.

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
@@ -1350,7 +1350,7 @@ bool RTPSParticipantImpl::createReader(
     return create_reader(ReaderOut, param, entityId, isBuiltin, enable, callback);
 }
 
-BaseReader* RTPSParticipantImpl::find_local_reader(
+std::shared_ptr<LocalReaderPointer> RTPSParticipantImpl::find_local_reader(
         const GUID_t& reader_guid)
 {
     shared_lock<shared_mutex> _(endpoints_list_mutex);
@@ -1359,11 +1359,11 @@ BaseReader* RTPSParticipantImpl::find_local_reader(
     {
         if (reader->getGuid() == reader_guid)
         {
-            return reader;
+            return reader->get_local_pointer();
         }
     }
 
-    return nullptr;
+    return std::shared_ptr<LocalReaderPointer>();
 }
 
 BaseWriter* RTPSParticipantImpl::find_local_writer(
@@ -1960,6 +1960,7 @@ bool RTPSParticipantImpl::deleteUserEndpoint(
 
     bool found = false, found_in_users = false;
     Endpoint* p_endpoint = nullptr;
+    BaseReader* reader = nullptr;
 
     if (endpoint.entityId.is_writer())
     {
@@ -1994,6 +1995,7 @@ bool RTPSParticipantImpl::deleteUserEndpoint(
         {
             if ((*rit)->getGuid().entityId == endpoint.entityId) //Found it
             {
+                reader = *rit;
                 m_userReaderList.erase(rit);
                 found_in_users = true;
                 break;
@@ -2004,6 +2006,7 @@ bool RTPSParticipantImpl::deleteUserEndpoint(
         {
             if ((*rit)->getGuid().entityId == endpoint.entityId) //Found it
             {
+                reader = *rit;
                 p_endpoint = *rit;
                 m_allReaderList.erase(rit);
                 found = true;
@@ -2062,6 +2065,10 @@ bool RTPSParticipantImpl::deleteUserEndpoint(
 #endif // if HAVE_SECURITY
     }
 
+    if (reader)
+    {
+        reader->local_actions_on_reader_removed();
+    }
     delete(p_endpoint);
     return true;
 }
@@ -2148,6 +2155,11 @@ void RTPSParticipantImpl::deleteAllUserEndpoints()
             (m_security_manager.*unregister_endpoint[kind])(endpoint->getGuid());
         }
 #endif // if HAVE_SECURITY
+
+        if (kind == READER)
+        {
+            static_cast<BaseReader*>(endpoint)->local_actions_on_reader_removed();
+        }
 
         // remove the endpoints
         delete(endpoint);
@@ -2874,8 +2886,11 @@ bool RTPSParticipantImpl::register_in_reader(
     }
     else if (!fastdds::statistics::is_statistics_builtin(reader_guid.entityId))
     {
-        BaseReader* reader = find_local_reader(reader_guid);
-        res = reader->add_statistics_listener(listener);
+        LocalReaderPointer::Instance local_reader(find_local_reader(reader_guid));
+        if (local_reader)
+        {
+            res = local_reader->add_statistics_listener(listener);
+        }
     }
 
     return res;
--- a/src/cpp/rtps/reader/BaseReader.cpp
+++ b/src/cpp/rtps/reader/BaseReader.cpp
@@ -107,6 +107,11 @@ BaseReader::BaseReader(
     setup_datasharing(att);
 }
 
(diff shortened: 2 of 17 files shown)


src/cpp/rtps/reader/BaseReader.cpp
```diff
@@ -272,6 +279,11 @@ void BaseReader::allow_unknown_writers()
     accept_messages_from_unkown_writers_ = true;
 }
 
+LocalReaderPointer BaseReader::get_local_pointer()
+{
+    return LocalReaderPointer(this, local_view_);
```
Mario-DL, line 284: A comment here: I haven't taken the reader's mutex before the `return` since tsan reported one potential deadlock in the `LatencyTest` between that one and the rtps endpoint collection mutex. I think it is fair (and quick) enough copying the shared pointer without protection.

src/cpp/rtps/writer/StatefulWriter.cpp
```diff
@@ -437,12 +437,11 @@ bool StatefulWriter::intraprocess_heartbeat(
...
     bool returned_value = false;
+    LocalReaderPointer::Instance local_reader = reader_proxy->local_reader();
 
-    std::lock_guard<RecursiveTimedMutex> guardW(mp_mutex);
-    RTPSReader* reader = RTPSDomainImpl::find_local_reader(reader_proxy->guid());
-
-    if (reader)
+    if (local_reader)
     {
+        std::lock_guard<RecursiveTimedMutex> guardW(mp_mutex);
```
MiguelCompany, line 444: Even though this guard is not new, so no deadlock would be introduced by the changes here, I would unlock it before calling `process_heartbeat_msg` on the reader. This would require copying `heartbeat_count_` into a local variable before unlocking the mutex.

src/cpp/rtps/writer/StatefulWriter.cpp
```diff
@@ -405,14 +405,14 @@ bool StatefulWriter::intraprocess_delivery(
         CacheChange_t* change,
         ReaderProxy* reader_proxy)
 {
-    BaseReader* reader = reader_proxy->local_reader();
-    if (reader)
+    LocalReaderPointer::Instance local_reader = reader_proxy->local_reader();
```
MiguelCompany, line 408: Include `LocalReaderPointer.hpp`

src/cpp/rtps/writer/StatelessWriter.cpp
```diff
@@ -947,9 +947,13 @@ bool StatelessWriter::get_connections(
...
-                    connection.guid(fastdds::statistics::to_statistics_type(reader.local_reader()->getGuid()));
-                    connection.mode(fastdds::statistics::ConnectionMode::INTRAPROCESS);
-                    connection_list.push_back(connection);
+                    LocalReaderPointer::Instance local_reader = reader.local_reader();
+                    if (local_reader)
+                    {
+                        connection.guid(fastdds::statistics::to_statistics_type(local_reader->getGuid()));
+                        connection.mode(fastdds::statistics::ConnectionMode::INTRAPROCESS);
+                        connection_list.push_back(connection);
+                    }
```
MiguelCompany, line 956: We could avoid getting the pointer to the local reader by using `ReaderLocator::remote_guid()` to retrieve the reader's GUID

src/cpp/utils/RefCountedPointer.hpp
```diff
@@ -0,0 +1,178 @@
...
+
+    class Instance;
+
+    explicit RefCountedPointer(
+            T* ptr)
+        : ptr_(ptr)
+        , is_active_(true)
+        , instances_(0)
+    {
+    }
```
MiguelCompany, line 65: Add doxydoc here to explain that the pointer should not be `nullptr`, and add an `assert(nullptr != ptr);`

src/cpp/utils/RefCountedPointer.hpp
```diff
@@ -0,0 +1,178 @@
...
+    RefCountedPointer(
+            RefCountedPointer&&) = delete;
+    RefCountedPointer& operator =(
+            RefCountedPointer&&) = delete;
+
+    class Instance
+    {
+    public:
+
+        explicit Instance(
```
MiguelCompany, line 83: Doxydoc

src/cpp/utils/RefCountedPointer.hpp
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    T* const ptr_;
+
+    std::atomic<bool> is_active_;
+
+    mutable std::mutex mutex_;
+    std::condition_variable cv_;
+    size_t instances_;
```
MiguelCompany, line 172: Doxydoc methods and members

MiguelCompany (changes requested):

src/cpp/utils/RefCountedPointer.hpp
Mario-DL: Shouldn't we let `ptr` be `nullptr` for the cases in which `find_local_reader()` returns an `std::shared_ptr<LocalReaderPointer>(nullptr)` ?

src/cpp/utils/RefCountedPointer.hpp
```diff
@@ -163,12 +189,26 @@ class RefCountedPointer
...
     }
 
+    /**
+     * Pointer to the managed object.
+     */
     T* const ptr_;
 
+    /**
+     * Indicates whether the pointee is still alive
+     * and the accessing the pointer is valid.
```
MiguelCompany, line 199:
```suggestion
     * and accessing the pointer is valid.
```

MiguelCompany (changes requested):
Apart from my nit below, this needs a rebase

MiguelCompany (approved):
LGTM with green CI 🏅

MiguelCompany:
@Mergifyio backport 3.1.x 3.0.x 2.14.x

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -2149,6 +2156,11 @@ void RTPSParticipantImpl::deleteAllUserEndpoints()
         }
 #endif // if HAVE_SECURITY
 
+        if (kind == READER)
+        {
+            static_cast<BaseReader*>(endpoint)->local_actions_on_reader_removed();
```
richiware, line 2161: Should not be `dynamic_cast`?

richiware (changes requested):

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
Mario-DL: The collections always store `BaseReader*` so it is well known to static_cast at this point

PR #5115, opened 2024-07-30
[21435] Solve SecurityManager memory issue

juanlofer-eprosima:
## Description

A memory issue is reported by ASAN when running tests using security such as BlackboxTests_DDS_PIM.Security.BuiltinAuthenticationPlugin_second_participant_creation_loop.
 
The issue is that the events thread might call `SecurityManager::resend_handshake_message_token`, which internally calls `WriterHistory::remove_change_and_reuse`, the latter not being (atomically) thread safe. If at the same time a handshake is processed (`SecurityManager::on_process_handshake`) from another thread, this may resolve in a change being added to the aforementioned writer history. As a result a change is added to `History::m_changes` vector, which would invalidate the iterator dealt with in `WriterHistory::remove_change_and_reuse` without protection.

Two potential solutions are proposed:

- Take the history/endpoint mutex before calling `WriterHistory::remove_change_and_reuse` from the security manager, as it's done when calling `PDPServer::remove_change_from_history_nts`
- Protect `WriterHistory::remove_change_and_reuse` fully, converting it in an atomic thread-safe method. Note that since the functions internally called from within this method already take the mutex, no (new) deadlocks should be introduced.

@Mergifyio backport 2.14.x 2.10.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/history/WriterHistory.cpp
+++ b/src/cpp/rtps/history/WriterHistory.cpp
@@ -332,6 +332,8 @@ CacheChange_t* WriterHistory::remove_change_and_reuse(
         return nullptr;
     }
 
+    std::lock_guard<RecursiveTimedMutex> guard(*mp_mutex);
+
     // Create a temporary reference change associated to the sequence number
     CacheChange_t ch;
     ch.sequenceNumber = sequence_number;


Mario-DL (commented):
The change makes sense to me.
I like more the approach of making the method thread safe, since `find_change` and `remove_change` already take the mutex as commented in the description.

However, if the first solution proposal is finally chosen, I think it would make sense to rename the method to `remove_change_and_reuse_nts`

Mario-DL (approved):
LGTM

Mario-DL:
We would need to backport this to `2.14` I think

MiguelCompany:
@Mergifyio backport 2.14.x 2.10.x

PR #5116, opened 2024-07-30
[19036] Link SHM locator kind with Fast DDS major version

EduPonz:
## Description

This PR links the definition of the SHM locator kind to the Fast DDS major version. That way, we avoid keeping SHM locators from participants of different majors, as the port structure will be different (the SHM port ABI version will change from major to major).

This PR fixes compatibility issues with 2.x versions when all the participants have an SHM transport without requiring further modifications on any 2.x version.

## Contributor Checklist

- [x] Commit messages follow the project guidelines. 
- [x] The code follows the style guidelines of this project. 
- [x] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
    - A static assert is added to ensure the value
- _N/A_: Any new/modified methods have been properly documented using Doxygen. 
- _N/A_: Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
- _N/A_: Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- ❌ Changes are API compatible. 
- [x] New feature has been added to the `versions.md` file (if applicable).
- _N/A_: New feature has been documented/Current behavior is correctly described in the documentation. 
    
- _N/A_: Applicable backports have been included in the description.

## Reviewer Checklist

- [x] The PR has a milestone assigned.
- [x] The title and description correctly express the PR's purpose.
- [x] Check contributor checklist is correct.
- _N/A_: If this is a critical bug fix, backports to the critical-only supported branches have been requested.
- [x] Check CI results: changes do not issue any warning.
- [x] Check CI results: failing tests are unrelated with the changes.

PR #1801, opened 2021-03-03
Use process and user id on SHM locators

MiguelCompany:

PR #5118, opened 2024-07-30
[19036] Discard local SHM locators that cannot be openned

EduPonz:
## Description

This PR adds a `is_locator_reachable` API to `NetworkFactory` so that unreachable locators can be discarded while deserializing the `DATA(p)`, `DATA(r)`, or `DATA(w)` messages. To do so, the `NetworkFactory` iterates over the transports looking for one that can reach the given locator. It does so using a new `TransportInterface::is_locator_reachable` API.

This PR partially solves a long standing communication issue when the participants are run in the same host but under different  users. This used to cause the Fast DDS participants to keep SHM locators from the remote participant that could not be opened (due to permissions); and since Fast DDS limits the communication to SHM as soon as the remote participant has a reachable SHM locator, communication was not possible.

After this PR:
* ✅ Communication with between two different users with similar permissions works in both directions
* ✅ On a super-user user, subscribing to a topic from a normal user works.
* ❌ On a normal user, subscribing to a topic from a user with more permissions (such as a super-user) still does not work.

**Depends on:**

* #5116

## Contributor Checklist

- ❌ Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- ❌ Changes are API compatible. 
    
## Reviewer Checklist

- _N/A_: If this is a critical bug fix, backports to the critical-only supported branches have been requested.

### DIFF
--- a/src/cpp/rtps/builtin/data/ParticipantProxyData.cpp
+++ b/src/cpp/rtps/builtin/data/ParticipantProxyData.cpp
@@ -32,7 +32,7 @@
 #include <rtps/builtin/data/ReaderProxyData.hpp>
 #include <rtps/builtin/data/WriterProxyData.hpp>
 #include <rtps/builtin/discovery/participant/PDPSimple.h>
-#include <rtps/network/NetworkFactory.h>
+#include <rtps/network/NetworkFactory.hpp>
 #include <rtps/resources/TimedEvent.h>
 #include <rtps/transport/shared_mem/SHMLocator.hpp>
 #include <utils/TimeConversion.hpp>
--- a/src/cpp/rtps/builtin/data/ParticipantProxyData.cpp
+++ b/src/cpp/rtps/builtin/data/ParticipantProxyData.cpp
@@ -446,13 +446,12 @@ bool ParticipantProxyData::writeToCDRMessage(
 bool ParticipantProxyData::readFromCDRMessage(
         CDRMessage_t* msg,
         bool use_encapsulation,
-        const NetworkFactory& network,
-        bool is_shm_transport_available,
+        NetworkFactory& network,
         bool should_filter_locators,
         fastdds::rtps::VendorId_t source_vendor_id)
 {
     auto param_process =
-            [this, &network, &is_shm_transport_available, &should_filter_locators, source_vendor_id](
+            [this, &network, &should_filter_locators, source_vendor_id](
         CDRMessage_t* msg, const ParameterId_t& pid, uint16_t plength)
             {
                 static_cast<void>(source_vendor_id);
@@ -503,7 +502,6 @@ bool ParticipantProxyData::readFromCDRMessage(
 
                         m_VendorId[0] = p.vendorId[0];
                         m_VendorId[1] = p.vendorId[1];
-                        is_shm_transport_available &= (m_VendorId == c_VendorId_eProsima);
                         break;
                     }
                     case fastdds::dds::PID_PRODUCT_VERSION:
@@ -613,7 +611,7 @@ bool ParticipantProxyData::readFromCDRMessage(
                                         m_guid.is_from_this_host()))
                             {
                                 ProxyDataFilters::filter_locators(
-                                    is_shm_transport_available,
+                                    network,
                                     metatraffic_locators,
                                     temp_locator,
                                     false);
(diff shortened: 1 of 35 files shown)


MiguelCompany (changes requested):

src/cpp/rtps/transport/shared_mem/SharedMemGlobal.hpp
```diff
@@ -53,10 +54,20 @@ class SharedMemGlobal
...
-        if (domain_name.length() > MAX_DOMAIN_NAME_LENGTH)
+        std::string username;
+        if(fastdds::dds::RETCODE_OK == SystemInfo::get_username(username))
+        {
+            domain_name_ = domain_name_ + "_" + username;
+        }
+        else
+        {
+            throw std::runtime_error("error: couldn't get username to generate port_segment_name");
+        }
```
MiguelCompany, line 65: I don't think we need to fail in this case, just fallback to the old behavior.

```suggestion
```

src/cpp/rtps/transport/shared_mem/SharedMemGlobal.hpp
```diff
@@ -53,10 +54,20 @@ class SharedMemGlobal
...
+
+        if (domain_name_.length() > MAX_DOMAIN_NAME_LENGTH)
         {
             throw std::runtime_error(
-                      domain_name +
+                      domain_name_ +
```
MiguelCompany, line 70: This would most probably fail, since adding the user name will exceed the imposed limit.

Looking at the code, I think we could remove this check if we pass the domain name to the constructor of `SharedMemGlobal::Port`

We will use the domain name passed in, instead of using `node_->domain_name`

src/cpp/rtps/transport/shared_mem/SHMLocator.hpp
```diff
@@ -50,8 +51,17 @@ class SHMLocator
...
+
+        if ((fastdds::dds::RETCODE_OK == SystemInfo::get_username(user_name)) && (user_name.length() < 16))
+        {
+            std::memcpy(&locator.address[1], user_name.c_str(), user_name.length());
+        }
+        else
+        {
+            locator.get_address()[1] = octet(host_id);
+            locator.get_address()[2] = octet(host_id >> 8);
+        }
```
MiguelCompany, line 64: This new approach can make a locator from the same user on a different host to be accepted.

What I would do is having the following format for the locator:

<kind> | <host_id> | <effective_user_id>

Where the first two fields are what we currently have, and the following one would be the result of `geteuid()` on POSIX systems, and something equivalent on Win32

MiguelCompany: Additional note: we followed and discarded a more convoluted approach on #1801

Mario-DL:
We internally agreed to address the convenient changes to overcome the last requirement on a follow up

src/cpp/rtps/builtin/data/ParticipantProxyData.cpp
```diff
@@ -446,16 +446,15 @@ bool ParticipantProxyData::writeToCDRMessage(
...
         bool should_filter_locators,
         fastdds::rtps::VendorId_t source_vendor_id)
 {
     auto param_process =
-            [this, &network, &is_shm_transport_available, &should_filter_locators, source_vendor_id](
+            [this, &network, &should_filter_locators, source_vendor_id](
         CDRMessage_t* msg, const ParameterId_t& pid, uint16_t plength)
             {
-                static_cast<void>(source_vendor_id);
+                m_VendorId = source_vendor_id;
```
juanlofer-eprosima, line 457: Double check this change would not introduce unwanted implications. Related to internal task #21423

Mario-DL: I think it makes sense, we set the `m_VendorId` to whatever is in the `change->vendor_id` and then, if the `PID_VENDORID` is specified, we set it to that value.

juanlofer-eprosima: I was not sure this was super correct since I though the method is supposed to read the content of a data p/w/r , setting its attributes given the fields parsed. Anyways, I did not find any behavior change given the current logic but I wanted someone else to verify just in case. And this will probably be addressed in the future when tackling the related task I made reference to.

juanlofer-eprosima (approved):
LGTM

PR #5154, opened 2024-08-13
repair service found that CDR serial transmission data has a maximum

huangjialu0216:
Repair service found that CDR serial transmission data has a maximum of 65535 bytes


MiguelCompany (changes requested):
@huangjialu0216 Thanks for your contribution.

It would be great if you could add a regression test in [BlackboxTestsDiscovery.cpp](https://github.com/eProsima/Fast-DDS/blob/master/test/blackbox/common/BlackboxTestsDiscovery.cpp)

PR #5381, opened 2024-11-04
[22006] Transform locators using new host_id PID - ABI compatible

cferreiragonz:
## Description
This PR adds a new PID containing a UUID of the machine/host. It is used to check if the metadata of other participants comes from the same host in a more reliable approach than the GUID comparison, which uses IP addresses interfaces.

This feature solves a bug in which a TCP client connecting to `localhost` is unable to identify the locator of a TCP discovery server with a custom GUID listening on `any`. This causes the client to create two different channels for the same connection, where the latest can never be reached.  The reason is that the server's locator is not considered from the same host due to its custom GUID. Hence, the server transforms the client's locator into `localhost` but the client does not do the same with the server's locator.

Another PR needs to be created for 3.2.x with the new PID properly added to `ParameterTypes.hpp`. That PR will not be backported.

@Mergifyio backport 3.0.x 2.14.x 2.10.x

## Contributor Checklist

- [X] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally. Related test: https://github.com/eProsima/Discovery-Server/pull/106
- [X] Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
    
- [X] Applicable backports have been included in the description.

## Reviewer Checklist


cferreiragonz:
This PR will be closed and replaced by the backport from PR:

- https://github.com/eProsima/Fast-DDS/pull/5382

PR #5460, opened 2024-12-04
[22321] Benchmark example updated

### LINKED ISSUE #5398: no benchmark example? and incorrect doc instructions

### Expected behavior

Expect that the docker instructions in this page are accurate:

https://fast-dds.docs.eprosima.com/en/v3.1.0/docker/fastdds_suite/fast_dds_suite.html#benchmark-example

### Current behavior

This is the examples directory in v3.1.0 without the "suite" (ubuntu-fastdds:v3.1.0 image name):

```sh
root@login07:/usr/local/eprosima/fastdds/share/fastdds/examples/cpp# ls -la
total 52
drwxr-xr-x 15 root root  336 Oct 15 11:20 .
drwxr-xr-x  3 root root   25 Oct 15 11:20 ..
drwxr-xr-x  3 root root 4096 Oct 15 11:20 configuration
drwxr-xr-x  3 root root 4096 Oct 15 11:20 content_filter
drwxr-xr-x  3 root root 4096 Oct 15 11:20 custom_payload_pool
drwxr-xr-x  3 root root 4096 Oct 15 11:20 delivery_mechanisms
drwxr-xr-x  3 root root 4096 Oct 15 11:20 discovery_server
drwxr-xr-x  3 root root 4096 Oct 15 11:20 flow_control
drwxr-xr-x  3 root root 4096 Oct 15 11:20 hello_world
drwxr-xr-x  4 root root 4096 Oct 15 11:20 request_reply
drwxr-xr-x  3 root root 4096 Oct 15 11:20 rtps
drwxr-xr-x  4 root root 4096 Oct 15 11:20 security
drwxr-xr-x  3 root root 4096 Oct 15 11:20 static_edp_discovery
drwxr-xr-x  3 root root 4096 Oct 15 11:20 topic_instances
drwxr-xr-x  3 root root 4096 Oct 15 11:20 xtypes
```

This is with the suite:

```sh
root@login07:~# goToExamples
root@login07:/usr/local/eprosima/fastdds/share/fastdds/examples/cpp# ls
configuration   custom_payload_pool  discovery_server  hello_world    rtps      static_edp_discovery  xtypes
content_filter  delivery_mechanisms  flow_control      request_reply  security  topic_instances
```

This is what happens with the 3.0.1 suite (ubuntu-fastdds-suite:v3.0.1):

root@login07:~# goToExamples
bash: cd: /usr/local/eprosima/share/fastdds/examples/cpp: No such file or directory

### Steps to reproduce

Download image, run	

### Fast DDS version/commit

v3.1.0, v3.0.1

### Platform/Architecture

Ubuntu Focal 20.04 amd64

### Transport layer

### Additional context

Hello world example works, but just can't find benchmark.

### XML configuration file

_No response_

### Relevant log output

_No response_

### Network traffic capture

_No response_

Javgilavi:
## Description

I remake the benchmark example to new example style maintaining the original functionality.

@Mergifyio backport 3.1.x 3.0.x 

Fixes #5398 

## Contributor Checklist

## Reviewer Checklist


rsanchez15 (changes requested):
Pending example tests

examples/cpp/benchmark/PublisherApp.cpp
```diff
@@ -0,0 +1,545 @@
...
+                }
+            }
+            break;
+
+        default:
+            throw std::runtime_error("Type invalid");
+    }
+}
+
+void PublisherApp::run()
```
Mario-DL, line 380: I think we could simplify this method. 
1. Wait for the matches (2) or is_stopped() event (loop is not needed, just the matched_lock condition variable)
2. if we are not stopped: 
  3. Publish the corresponding message (this could be done in another method), 
  4. Do the initial wait
  5. Keep the same loop of lines 412:428.  
  6. Present results
 
What do you think ?

examples/cpp/benchmark/PublisherApp.cpp
```diff
@@ -0,0 +1,545 @@
...
+        std::cout << vSamples[i] << ",";
+    }
+    std::cout << std::endl;
+
+    // Wait in case a response is still nedded
+    std::unique_lock<std::mutex> final_lock(mutex_);
+    cv_.wait(final_lock, [&]()
+            {
+                return is_stopped() || (sent == 0);
+            });
```
Mario-DL, line 464: I would say we dont need this logic at the end, the test-time has already finished and I find no reason for waiting a final response.
In addition, there is a typo in the comment (`nedded`).
If this suggestion is accepted, I think we could we could get rid of the `sent` member

examples/cpp/benchmark/PublisherApp.cpp
```diff
@@ -0,0 +1,545 @@
...
+        if (check)
+        {
+            return;
+        }
+        vSamples.push_back(static_cast<uint16_t>(count) - prevCount);
+        prevCount = static_cast<uint16_t>(count);
+        actualTime = std::chrono::steady_clock::now();
+        elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(actualTime - startTime);
+    }
+    std::cout << "RESULTS after " << elapsed.count() << " milliseconds:" << std::endl;
```
Mario-DL, line 429: As we are collecting some interesting data (i.e the number of messages exchanged for each period of time), it could be a good idea to add another result metric of how many bytes per second it has achieved (a kind of `throughtput` measurement)

examples/cpp/benchmark/PublisherApp.cpp
```diff
@@ -0,0 +1,545 @@
...
+
+    switch (msg_size_)
+    {
+        case CLIParser::MsgSizeKind::NONE:
+            while ((!is_stopped()) && (RETCODE_OK == reader->take_next_sample(&benchmark_, &info)))
+            {
+                if ((info.instance_state == ALIVE_INSTANCE_STATE) && info.valid_data)
+                {
+                    std::cout << "Sample with index: '" <<
+                        benchmark_.index() << "' (0 Bytes) RECEIVED" << std::endl;
```
Mario-DL, line 243: I found it a little weird to see `0 bytes` when the `unsigned long` of the index is 4 bytes. We may instead writer smth similar to 
```suggestion
                        benchmark_.index() << "' (Array size 0) RECEIVED" << std::endl;
```
This suggestion would apply also to the subscriber

examples/cpp/benchmark/PublisherApp.cpp
```diff
@@ -0,0 +1,545 @@
...
+        // Wait for period or stop event
+        std::unique_lock<std::mutex> periodic_lock(mutex_);
+        auto check = cv_.wait_for(periodic_lock, std::chrono::milliseconds(period_ms_), [&]()
+                        {
+                            return is_stopped();
+                        });
+        if (check)
+        {
+            return;
+        }
```
Mario-DL, line 423: The `is_stopped()` is being checked after in the while loop. Avoiding these lines checking the output of the `wait_for` seems cleaner to me

examples/cpp/benchmark/SubscriberApp.cpp
```diff
@@ -0,0 +1,375 @@
...
+                    std::cout << "Sample with index: '" <<
+                        benchmark_.index() << "' (0 Bytes) RECEIVED" << std::endl;
+                    benchmark_.index(benchmark_.index() + 1);
+                    while (matched_ == 0)
+                    {
+                        std::unique_lock<std::mutex> initial_lock(mutex_);
+                        auto check = cv_.wait_for(initial_lock, std::chrono::milliseconds(1), [&]()
+                                        {
+                                            return is_stopped();
+                                        });
```
Mario-DL, line 239: We typically recommend not to perform blockign operations in callbacks. I think it would make more sens moving these lines to the first part of the `run()` method.
Additionally, I think we could change the `matched_` condition to be greater of equal to 2 since we expect both of out endpoints to be matched with the publisher ones.

examples/cpp/benchmark/README.md
```diff
@@ -0,0 +1,287 @@
...
+First Sample with index: '0' (8388608 Bytes) SENT
+Sample with index: '1' (8388608 Bytes) RECEIVED
+Sample with index: '0' (8388608 Bytes) SENT
+Sample with index: '1' (8388608 Bytes) RECEIVED
+Sample with index: '0' (8388608 Bytes) SENT
+Sample with index: '1' (8388608 Bytes) RECEIVED
+Sample with index: '2' (8388608 Bytes) SENT
+Sample with index: '3' (8388608 Bytes) RECEIVED
+Sample with index: '4' (8388608 Bytes) SENT
+Sample with index: '5' (8388608 Bytes) RECEIVED
```
Mario-DL, line 82: Testing locally I found a couple of strange things to me in the publisher side:
* The first index seems to be sent and received twice
* The RECEIVED is only printed for odd sample idexes (1,3,5,...)
Is that behavior correct ?

examples/cpp/benchmark/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+## XML profile playground
+
+The *eProsima Fast DDS* entities can be configured through an XML profile from the environment.
+This is accomplished by setting the environment variable ``FASTDDS_DEFAULT_PROFILES_FILE`` to path to the XML profiles file:
+
+* Ubuntu ( / MacOS )
+
+    ```shell
+    user@machine:example_path$ export FASTDDS_DEFAULT_PROFILES_FILE=hello_world_profile.xml
```
Mario-DL, line 271:
```suggestion
    user@machine:example_path$ export FASTDDS_DEFAULT_PROFILES_FILE=benchmark_profile.xml
```

examples/cpp/benchmark/README.md
```diff
@@ -0,0 +1,287 @@
...
+    ```
+
+* Windows
+
+    ```powershell
+    example_path> set FASTDDS_DEFAULT_PROFILES_FILE=hello_world_profile.xml
```
Mario-DL, line 277:
```suggestion
    example_path> set FASTDDS_DEFAULT_PROFILES_FILE=benchmark_profile.xml
```

test/examples/test_benchmark.py
```diff
@@ -0,0 +1,80 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import subprocess
+import pytest
+import re
+
+config_test_cases = [
+    ('--transport DEFAULT --msg-size NONE', '--transport DEFAULT --msg-size NONE'),           # Builtin transports
```
Mario-DL, line 20:
```suggestion
    ('--transport DEFAULT --msg-size NONE', '--transport DEFAULT --msg-size NONE'), 
```

Mario-DL (changes requested):
Good first iteration overall. The PR brings an interesting example.
In addition to my comments below, we would need to add this PR content `Refactor benchmark example` to the `versions.md` and replace the `N/A` with a tick in the checklist.

examples/cpp/benchmark/SubscriberApp.cpp
Javgilavi: The idea is to make sure the subscriber not to sent back the data until the on_publication_match happen. To solve the case of the subscriber receiving the data but not sending it properly.
I don't understand how this could be moved to `run()`.
Also, I understand the comment of change for matched_ to be greater or equal to 2 is for the publisher and not here, right?

examples/cpp/benchmark/README.md
Javgilavi: The problem of repetting the first index is fix.
The RECEIVED is only printed for odd because they are sent by the subscriber and receive in the publisher, and the evens are sent by the publisher and receive by the subscriber

examples/cpp/benchmark/Benchmark_big.idl
```diff
@@ -0,0 +1,5 @@
+struct BenchMarkBig
```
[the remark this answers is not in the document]
Mario-DL: Please, also move the `.idls`

examples/cpp/benchmark/SubscriberApp.cpp
Mario-DL: The first thing in the `run()` method would be:
```
{
        // Wait for the data endpoints discovery
        std::unique_lock<std::mutex> matched_lock(mutex_);
        cv_.wait(matched_lock, [&]()
                {
                    // at least one has been discovered
                    return ((matched_ >= 2) || is_stopped());
                });
}
```
Then, I would remove all `while (matched_ == 0){}` blocks from the `on_data_available()`. This will make the subscriber to first wait to be matched and simplify the 'on_data_available()`. We should until greater or equal than 2 since both subscriber endpoints must be matched.

examples/cpp/benchmark/README.md
Mario-DL: Yeah, I reviewed the logic and I first thought that the subscriber was not incrementing the index but it is doing it, so it is expected.

Mario-DL: Please, update the commands output since the `index 0, index 1, index 0,...` behavior was already corrected and no longer exists

examples/cpp/benchmark/SubscriberApp.cpp
```diff
@@ -0,0 +1,399 @@
...
+                        if (check)
+                        {
+                            return;
+                        }
+                    }
+                    if (benchmark_.index() >= samples_)
```
Mario-DL, line 246: I think we are missing this here, otherwise when the example is launched without `--samples` the subscriber will receive the first sample and skip.
```suggestion
                    if (samples_ != 0 && benchmark_.index() >= samples_)
```

examples/cpp/benchmark/SubscriberApp.cpp
Mario-DL: Whats more, we could avoid waiting for any `matched_` within `run()`, since the reliability and durability will do the late-joining this for us when specified

Javgilavi: As we discussed in private chat, I just delated the blocking operation on the  `on_data_available()`.

test/examples/benchmark.compose.yml
```diff
@@ -0,0 +1,46 @@
...
+      - @PROJECT_BINARY_DIR_COMPOSE_VOLUME@
+      - @fastcdr_LIB_DIR_COMPOSE_VOLUME@
+      - @CMAKE_INSTALL_PREFIX_COMPOSE_VOLUME@
+      @TINYXML2_LIB_DIR_COMPOSE_VOLUME@
+    environment:
+      @PATH_ENVIRONMENT_VARIABLE_COMPOSE@
+      EXAMPLE_DIR: @EXAMPLE_PREFIX_DIR_COMPOSE@/benchmark/@EXAMPLE_SUFFIX_DIR_COMPOSE@
+      FASTDDS_DEFAULT_PROFILES_FILE: @FASTDDS_DEFAULT_PROFILES_FILE_PREFIX_COMPOSE@/benchmark/benchmark_profile.xml
+      PUBLISHER_ADDITIONAL_ARGUMENTS: ${PUB_ARGS}
+    command: @SHELL_EXECUTABLE@ -c "@COMMAND_EXAMPLE_DIR_PREFIX_COMPOSE@/benchmark@FILE_EXTENSION@ publisher $${PUBLISHER_ADDITIONAL_ARGUMENTS}"
```
Mario-DL, line 46:
```suggestion
    command: @SHELL_EXECUTABLE@ -c "@COMMAND_EXAMPLE_DIR_PREFIX_COMPOSE@/benchmark@FILE_EXTENSION@ publisher --reliable --transient-local $${PUBLISHER_ADDITIONAL_ARGUMENTS}"
```

test/examples/benchmark.compose.yml
```diff
@@ -0,0 +1,46 @@
...
+      - @PROJECT_BINARY_DIR_COMPOSE_VOLUME@
+      - @fastcdr_LIB_DIR_COMPOSE_VOLUME@
+      - @CMAKE_INSTALL_PREFIX_COMPOSE_VOLUME@
+      @TINYXML2_LIB_DIR_COMPOSE_VOLUME@
+    environment:
+      @PATH_ENVIRONMENT_VARIABLE_COMPOSE@
+      EXAMPLE_DIR: @EXAMPLE_PREFIX_DIR_COMPOSE@/benchmark/@EXAMPLE_SUFFIX_DIR_COMPOSE@
+      FASTDDS_DEFAULT_PROFILES_FILE: @FASTDDS_DEFAULT_PROFILES_FILE_PREFIX_COMPOSE@/benchmark/benchmark_profile.xml
+      SUBSCRIBER_ADDITIONAL_ARGUMENTS: ${SUB_ARGS}
+    command: @SHELL_EXECUTABLE@ -c "@COMMAND_EXAMPLE_DIR_PREFIX_COMPOSE@/benchmark@FILE_EXTENSION@ subscriber $${SUBSCRIBER_ADDITIONAL_ARGUMENTS}"
```
Mario-DL, line 30: In order to avoid flakiness, add the `--reliable` and `--transient-local` arguments
```suggestion
    command: @SHELL_EXECUTABLE@ -c "@COMMAND_EXAMPLE_DIR_PREFIX_COMPOSE@/benchmark@FILE_EXTENSION@ subscriber --reliable --transient-local $${SUBSCRIBER_ADDITIONAL_ARGUMENTS}"
```

Mario-DL (changes requested):
Just couple of comments

Mario-DL (approved):
LGTM with green CI

rsanchez15 (approved):

Javgilavi:
@Mergifyio backport 3.1.x 3.0.x

PR #5540, opened 2025-01-07
[22537] Fix EDP reliability timings

fujitatomoya:
## Description

happened to find this miscalculation during https://github.com/ros2/rmw_fastrtps/issues/789.
i believe that comments are the correct values, so that default Duration_t adjusted accordingly in milliseconds.

@mergifyio backport 3.1.x 3.0.x 2.14.x 2.10.x

## Contributor Checklist

- _N/A_: Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
    
## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/builtin/discovery/endpoint/EDPSimple.cpp
+++ b/src/cpp/rtps/builtin/discovery/endpoint/EDPSimple.cpp
@@ -56,9 +56,9 @@ namespace rtps {
 
 // Default configuration values for EDP entities.
 static const dds::Duration_t edp_heartbeat_period{1, 0}; // 1 second
-static const dds::Duration_t edp_nack_response_delay{0, 100 * 1000 }; // 100 milliseconds
-static const dds::Duration_t edp_nack_supression_duration{0, 10 * 1000}; // 10 milliseconds
-static const dds::Duration_t edp_heartbeat_response_delay{0, 10 * 1000}; // 10 milliseconds
+static const dds::Duration_t edp_nack_response_delay{0, 100 * 1000 * 1000 }; // 100 milliseconds
+static const dds::Duration_t edp_nack_supression_duration{0, 10 * 1000 * 1000}; // 10 milliseconds
+static const dds::Duration_t edp_heartbeat_response_delay{0, 10 * 1000 * 1000}; // 10 milliseconds
 
 static const int32_t edp_reader_initial_reserved_caches = 1;
 static const int32_t edp_writer_initial_reserved_caches = 20;
--- a/test/blackbox/common/DDSBlackboxTestsStatistics.cpp
+++ b/test/blackbox/common/DDSBlackboxTestsStatistics.cpp
@@ -762,7 +762,7 @@ TEST(DDSStatistics, correct_deletion_upon_delete_contained_entities)
                 {
                     writer->init();
                     ASSERT_TRUE(writer->isInitialized());
-                    writer->wait_discovery();
+                    writer->wait_discovery(std::chrono::seconds(3));
                     writer->send(*data, 10);
                     writer->destroy();
                 }));
--- a/test/blackbox/common/DDSBlackboxTestsStatistics.cpp
+++ b/test/blackbox/common/DDSBlackboxTestsStatistics.cpp
@@ -762,8 +762,8 @@ TEST(DDSStatistics, correct_deletion_upon_delete_contained_entities)
                 {
                     writer->init();
                     ASSERT_TRUE(writer->isInitialized());
-                    writer->wait_discovery(std::chrono::seconds(3));
-                    writer->send(*data, 10);
+                    writer->wait_discovery();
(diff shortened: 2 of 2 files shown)


src/cpp/rtps/builtin/discovery/endpoint/EDPSimple.cpp
```diff
@@ -56,9 +56,9 @@ namespace rtps {
 
 // Default configuration values for EDP entities.
 static const dds::Duration_t edp_heartbeat_period{1, 0}; // 1 second
-static const dds::Duration_t edp_nack_response_delay{0, 100 * 1000 }; // 100 milliseconds
-static const dds::Duration_t edp_nack_supression_duration{0, 10 * 1000}; // 10 milliseconds
-static const dds::Duration_t edp_heartbeat_response_delay{0, 10 * 1000}; // 10 milliseconds
```
fujitatomoya, line 61: these are actually `microseconds`, not `milliseconds`.

fujitatomoya:
@MiguelCompany can you take a look?

src/cpp/rtps/builtin/discovery/endpoint/EDPSimple.cpp
```diff
@@ -56,9 +56,9 @@ namespace rtps {
...
+static const dds::Duration_t edp_nack_response_delay{0, 100 * 1000000 }; // 100 milliseconds
+static const dds::Duration_t edp_nack_supression_duration{0, 10 * 1000000}; // 10 milliseconds
+static const dds::Duration_t edp_heartbeat_response_delay{0, 10 * 1000000}; // 10 milliseconds
```
MiguelCompany, line 61: The values changed in commit 39fa6011d84e311060cb698d5b4e0ac4efc96163. The comment was updated, but the value was wrongly converted.

I prefer the old format (multiplying twice by 1000):

```suggestion
static const dds::Duration_t edp_nack_response_delay{0, 100 * 1000 * 1000 }; // 100 milliseconds
static const dds::Duration_t edp_nack_supression_duration{0, 10 * 1000 * 1000}; // 10 milliseconds
static const dds::Duration_t edp_heartbeat_response_delay{0, 10 * 1000 * 1000}; // 10 milliseconds
```

MiguelCompany (changes requested):

fujitatomoya:
@MiguelCompany took your suggestion, can you check and approve?

MiguelCompany (approved):

MiguelCompany:
@Mergifyio backport 3.1.x 3.0.x 2.14.x 2.10.x

MiguelCompany:
@fujitatomoya `BlackboxTests_DDS_PIM.DDSStatistics.correct_deletion_upon_delete_contained_entities` has failed both in Ubuntu and Windows, so it seems like a regression. Could you take a look?

test/blackbox/common/DDSBlackboxTestsStatistics.cpp
```diff
@@ -762,7 +762,7 @@ TEST(DDSStatistics, correct_deletion_upon_delete_contained_entities)
                 {
                     writer->init();
                     ASSERT_TRUE(writer->isInitialized());
-                    writer->wait_discovery();
+                    writer->wait_discovery(std::chrono::seconds(3));
```
fujitatomoya, line 765: https://github.com/eProsima/Fast-DDS/actions/runs/12665512557/job/35295559797?pr=5540 does not tell much information besides `correct_deletion_upon_delete_contained_entities` test is failing...

my assumption is that this PR extends the EDP time compared to the configuration before, actually 1000 times slower in the computation time.
that makes this specific case takes more time for endpoint discovery, and then all the test reader and writer threads cannot join in time during `correct_deletion_upon_delete_contained_entities` test.

actually this test is already flaky depending on the thread dispatching scheduling.

test requires 2 participants in the same topic path, if one of the writer threads is not dispatched just yet, that could be waiting on `writer->wait_discovery()` forever, and eventually that thread cannot be joined at the end.

having the timeout for `writer->wait_discovery()` should mitigate the flaky situation, and most likely this is just fine for the test.

@MiguelCompany @Mario-DL (since you are the author who added this test) what do you think?

Mario-DL: I agree that the failure has to do with the test but setting a timeout for the `wait_discovery()` may cause an infinite wait in the `block_for_at_least()` in the reader's side.
Instead, I would suggest increasing the time between samples writes (`writer->send(*data, 200);`) that gives enough time for discovering and message reception. I also tested that the test keeps reproducing the original statistics regression with this change.

MiguelCompany (approved):
LGTM with green CI

Mario-DL (approved):

fujitatomoya:
@Mario-DL thanks!

PR #5553, opened 2025-01-09
[22202] JSON to DynamicData deserializer

juanlofer-eprosima:
## Description

This PR is a follow-up of https://github.com/eProsima/Fast-DDS/pull/4780, in which a `DynamicData` to JSON serializer was introduced. Here a new utility method is introduced to perform the inverse transformation, i.e. from JSON to `DynamicData`. In addition, `json_serialize` method has been modified to:

- Handle enums with base type other that int32
- Improve wide characters and strings handling
- Increase overall robustness
- Decrease 'complexity' through higher encapsulation

## Contributor Checklist

    - Related documentation PR: https://github.com/eProsima/Fast-DDS-docs/pull/1093

## Reviewer Checklist


src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
```diff
@@ -0,0 +1,1385 @@
...
+            {
+                EPROSIMA_LOG_ERROR(XTYPES_UTILS, "Error encountered while deserializing TK_CHAR16 member: expected 1-character string.");
+                return RETCODE_BAD_PARAMETER;
+            }
+
+            // Convert UTF-8 value to wstring
+            std::string j_string = j.get<std::string>();
+            std::wstring aux_wstring({L'\0'});
+#if defined(MINGW_COMPILER)
+            // WARNING: it is the user responsibility to set the appropiate UTF-8 locale before calling this method
```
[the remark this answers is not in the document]
juanlofer-eprosima: Good catch buddy.

src/cpp/fastdds/xtypes/utils.cpp
```diff
@@ -81,6 +91,38 @@ ReturnCode_t json_serialize(
...
+
+    ReturnCode_t ret;
+    if (RETCODE_OK !=
+            (ret =
+            json_deserialize(j, traits<DynamicType>::narrow<DynamicTypeImpl>(dynamic_type), format, data)))
+    {
+        EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                "Error encountered while performing JSON to DynamicData deserialization.");
+        data.reset(); // Reset data in case of error
+    }
```
Carlosespicur, line 122:
```suggestion
    ReturnCode_t ret = json_deserialize(j, traits<DynamicType>::narrow<DynamicTypeImpl>(dynamic_type), format, data);
    if (RETCODE_OK != ret)
    {
        EPROSIMA_LOG_ERROR(XTYPES_UTILS,
                "Error encountered while performing JSON to DynamicData deserialization.");
        data.reset(); // Reset data in case of error
    }
```

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
```diff
@@ -0,0 +1,1386 @@
...
+                    "' from JSON: get_member_by_name failed.");
+            break;
+        }
+
+        if (RETCODE_OK != (ret = json_deserialize_member(it.value(), type_member, format, data)))
+        {
+            EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                    "Error encountered while deserializing " << kind_str << " member '" << it.key() << "'.");
+            break;
+        }
```
Carlosespicur, line 141: Apply in the rest of file:

```suggestion
        ret = data->enclosing_type()->get_member_by_name(type_member, it.key());

        if (RETCODE_OK != ret)
        {
            EPROSIMA_LOG_ERROR(XTYPES_UTILS,
                    "Error encountered while deserializing " << kind_str << " member '" << it.key() <<
                    "' from JSON: get_member_by_name failed.");
            break;
        }

        ret = json_deserialize_member(it.value(), type_member, format, data);

        if (RETCODE_OK != ret)
        {
            EPROSIMA_LOG_ERROR(XTYPES_UTILS,
                    "Error encountered while deserializing " << kind_str << " member '" << it.key() << "'.");
            break;
        }
```

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
```diff
@@ -0,0 +1,1397 @@
...
+    return ret;
+}
+
+ReturnCode_t json_deserialize_member(
+        const nlohmann::json& j,
+        const traits<DynamicTypeMember>::ref_type& type_member,
+        DynamicDataJsonFormat format,
+        traits<DynamicDataImpl>::ref_type& data) noexcept
+{
+    MemberDescriptorImpl& member_desc =
```
Carlosespicur, line 152: NIT: Perhaps we could check here that `type_member` is not nil (assert), and don't rely only in `json_deserialize_aggregate` checks.

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
```diff
@@ -0,0 +1,1397 @@
...
+
+    return json_deserialize_aggregate(j, format, data_impl);
+}
+
+ReturnCode_t json_deserialize_aggregate(
+        const nlohmann::json& j,
+        DynamicDataJsonFormat format,
+        traits<DynamicDataImpl>::ref_type& data) noexcept
+{
+    std::string kind_str = (data->enclosing_type()->get_kind() == TK_STRUCTURE) ? "structure" : "bitset";
```
Carlosespicur, line 107: NIT: Same here for `data`. Same in the rest of functions

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
```diff
@@ -0,0 +1,1397 @@
...
+                ReturnCode_t ret;
+                if (TK_BITSET == data->enclosing_type()->get_kind())
+                {
+                    const auto value = numeric_get<uint8_t>(j);
+                    if (value != 0 && value != 1)
+                    {
+                        EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                                "Error encountered while deserializing TK_BOOLEAN member: expected 0 or 1, got " <<
+                                value);
+                        return RETCODE_BAD_PARAMETER;
```
Carlosespicur: I think it is redundant with the try-catch block, perhaps we could simply throw a `std::invalid_argument` exception.

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
```diff
@@ -0,0 +1,1378 @@
...
+            std::wstring aux_wstring({L'\0'});
+#if defined(MINGW_COMPILER)
+            // WARNING: it is the user responsibility to set the appropriate UTF-8 locale before calling this method
+            int size_needed = std::mbstowcs(nullptr, j_string.c_str(), 0);
+            if (size_needed < 0)
+            {
+                EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                        "Error encountered while deserializing TK_CHAR16 member: invalid UTF-8 string.");
+                return RETCODE_BAD_PARAMETER;
+            }
```
Carlosespicur: `std::mbstowcs` returns a `size_t`, so this check is invalid:


```suggestion
            size_t size_needed = std::mbstowcs(nullptr, j_string.c_str(), 0);
            if (size_needed == static_cast<size_t>(-1))
            {
                EPROSIMA_LOG_ERROR(XTYPES_UTILS,
                        "Error encountered while deserializing TK_CHAR16 member: invalid UTF-8 string.");
                return RETCODE_BAD_PARAMETER;
            }
```

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
```diff
@@ -0,0 +1,1376 @@
...
+            std::wstring value;
+#if defined(MINGW_COMPILER)
+            // WARNING: it is the user responsibility to set the appropriate UTF-8 locale before calling this method
+            int size_needed = std::mbstowcs(nullptr, j_string.c_str(), 0);
+            if (size_needed < 0)
+            {
+                EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                        "Error encountered while deserializing TK_STRING16 member: invalid UTF-8 string.");
+                return RETCODE_BAD_PARAMETER;
+            }
```
Carlosespicur: Same here

src/cpp/fastdds/xtypes/serializers/json/dynamic_data_json.cpp
```diff
@@ -746,20 +432,40 @@ ReturnCode_t json_serialize_basic_member(
...
+                std::string utf8_value("\0", 1);
+#if defined(MINGW_COMPILER)
+                // WARNING: it is the user responsibility to set the appropriate UTF-8 locale before calling this method
+                int size_needed = std::wcstombs(nullptr, aux_wstring_value.c_str(), 0);
+                if (size_needed < 0)
+                {
+                    EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                            "Error encountered while serializing TK_CHAR16 member to JSON: encountered invalid character.");
+                    return RETCODE_BAD_PARAMETER;
+                }
```
Carlosespicur: Same situation as discussed in `json_dynamic_data.cpp`
```suggestion
                size_t size_needed = std::wcstombs(nullptr, aux_wstring_value.c_str(), 0);
                if (size_needed == static_cast<size_t>(-1))
                {
                    EPROSIMA_LOG_ERROR(XTYPES_UTILS,
                            "Error encountered while serializing TK_CHAR16 member to JSON: encountered invalid character.");
                    return RETCODE_BAD_PARAMETER;
                }
```

src/cpp/fastdds/xtypes/serializers/json/dynamic_data_json.cpp
```diff
@@ -790,17 +496,38 @@ ReturnCode_t json_serialize_basic_member(
             if (RETCODE_OK == ret)
             {
                 // Insert UTF-8 converted value
-#ifdef MINGW_COMPILER
                 std::string utf8_value;
-                int size_needed = std::wcstombs(nullptr, value.data(), 0);
-                if (size_needed > 0)
+#ifdef MINGW_COMPILER
+                // WARNING: it is the user responsibility to set the appropriate UTF-8 locale before calling this method
+                int size_needed = std::wcstombs(nullptr, value.c_str(), 0);
```
Carlosespicur: Same here

src/cpp/fastdds/xtypes/serializers/json/dynamic_data_json.cpp
```diff
@@ -895,46 +549,281 @@ ReturnCode_t json_serialize_basic_member(
...
+    assert(enum_type->get_kind() == TK_ENUM);
+    auto enclosing_type_impl = traits<DynamicType>::narrow<DynamicTypeImpl>(enum_type->get_all_members_by_index().at(
+                        0)->get_descriptor().type()); // Unfortunately DynamicDataImpl::get_enclosing_typekind is private
+    if (nullptr == enclosing_type_impl)
+    {
+        EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                "Error encountered while serializing TK_ENUM member to JSON: null enclosing type.");
+        return RETCODE_BAD_PARAMETER;
+    }
+    TypeKind enclosing_kind = enclosing_type_impl->get_kind();
```
Carlosespicur, line 595:
```suggestion
```

src/cpp/fastdds/xtypes/serializers/json/dynamic_data_json.cpp
```diff
@@ -991,6 +880,209 @@ ReturnCode_t json_serialize_array(
...
+            EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                    "Error encountered while serializing map member's member to JSON: invalid member id.");
+            return RETCODE_BAD_PARAMETER;
+        }
+
+        auto it = id_to_key.find(id);
+        if (it == id_to_key.end())
+        {
+            EPROSIMA_LOG_ERROR(XTYPES_UTILS,
+                    "Error encountered while serializing map member's member to JSON: key not found.");
```
Carlosespicur, line 928: As `id_to_key` is only used to find the id, perhaps we could just reuse the `key_to_id` map:
```suggestion
        auto it = std::find_if(key_to_id.begin(), key_to_id.end(), [&id](const auto& value){value.second == id});
```

test/unittest/dds/xtypes/serializers/json/DynDataJSONTests.cpp
```diff
@@ -59,15 +59,28 @@ void test_generic()
 
             ASSERT_NE(dyn_data, nullptr);
 
+            // Test DynamicData to JSON serialization
```
Carlosespicur, line 67: Perhaps should we add here some negative tests (for example, trying to create a DynamicData from a JSON with invalid values)

Carlosespicur (changes requested):

src/cpp/fastdds/xtypes/utils.cpp
juanlofer-eprosima: I'd prefer to keep this style unless there is a good reason to change.

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
juanlofer-eprosima: Same as above.

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
juanlofer-eprosima: I will consider it, I'd say I am both trusting the right checks are done in the higher level methods, but also that DynamicData API results are valid (in this case no nullptrs are given) when the corresponding return codes are OK. If following your proposal, I would be needing to add many more checks than this one.

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
juanlofer-eprosima: Same as above.

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
juanlofer-eprosima: I don't think I understood your point.

src/cpp/fastdds/xtypes/serializers/json/dynamic_data_json.cpp
juanlofer-eprosima: But I believe the current implementation is more efficient. If I am not wrong, the current implementation is O(NlogN) and the one you are proposing would be O(N²).

test/unittest/dds/xtypes/serializers/json/DynDataJSONTests.cpp
juanlofer-eprosima: I think you are right, however there are many possible cases, which ones do you think I should cover?

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
juanlofer-eprosima: I believe it is correct, since implicit casting would be applied when -1 is returned. However with your proposed change, the failing case would be missed when casting -1 to size_t. Well not exactly, it would capture it, but would not be able to differentiate the failing case with the case where the maximum value is requested (unless this value is reserved by definition, would need to check that).

src/cpp/fastdds/xtypes/serializers/json/dynamic_data_json.cpp
Carlosespicur: Yes, you're right. Closing this

test/unittest/dds/xtypes/serializers/json/DynDataJSONTests.cpp
Carlosespicur: For example, deserializing a corrupt json and check that `numeric_get` errors are handled correctly

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
juanlofer-eprosima: After investigating, I will apply your suggestion. The failing case was correctly captured as a -1 in the int variable, however we could be overflowing for a high return value. Also, it seems that SIZE_MAX is indeed reserved for the failing case.

src/cpp/fastdds/xtypes/serializers/json/json_dynamic_data.cpp
juanlofer-eprosima: I finally decided not to apply this suggestion, since after all the specific (de)serializers are just methods to encapsulate blocks that were originally in a huge unique (de)serialize function, were the security checks are made. They are implementation methods that are not expected to be used elsewhere, if that changes at some point I'd say to apply your suggestion then.

Carlosespicur:
LGTM

Carlosespicur (approved):

Issue #5214, opened 2024-09-10
[21614] Security tip of Fastdds Discovery Server is incorrect


chunyisong:
### Is there an already existing issue for this?

- [X] I have searched the existing issues

### Expected behavior

Security tip should reflect the actual status of "security feature(As far as I know, it should be for secure or safe communication)".

### Current behavior

Security tip is determined by server-id

### Steps to reproduce

![image](https://github.com/user-attachments/assets/b3d709a0-fe72-4e86-9b2f-2d1dca1d8286)

### Fast DDS version/commit

FastDDS v3.0.0

### Platform/Architecture

Ubuntu Focal 20.04 amd64

### Transport layer

UDPv4, TCPv4

JesusPoderoso:
Hi @chunyisong, thanks for your report.
The eProsima team will review and reproduce the issue and come back with some feedback in the following days.

cferreiragonz:
Hi @chunyisong, the fix will be available with the next patch 3.1.3.

PR #5702, opened 2025-03-12
[21614] Fix Security tag in DS CLI

cferreiragonz:
## Description

This PR fixes Issue (https://github.com/eProsima/Fast-DDS/issues/5214) in which the SECURITY field does not show the real state of the Discovery Server initialized.

In 3.2.0 the CLI display was modified and 3.0.2 was the EOL, so no backports are needed.

## Contributor Checklist

- _N/A_ Changes are API compatible. 
    
## Reviewer Checklist

### DIFF
--- a/tools/fds/server.cpp
+++ b/tools/fds/server.cpp
@@ -595,9 +595,9 @@ int fastdds_discovery_server(
 #endif // ifndef _WIN32
 
         bool has_security = false;
-        auto check_qos = pServer->get_qos();
+        const auto& check_qos = pServer->get_qos();
         // Access the properties of the participant QoS
-        for (auto prop : check_qos.properties().properties())
+        for (const auto& prop : check_qos.properties().properties())
         {
             if (prop.name() == "dds.sec.auth.builtin.PKI-DH.identity_ca" && prop.value() != "")
             {
--- a/tools/fds/server.cpp
+++ b/tools/fds/server.cpp
@@ -571,7 +571,6 @@ int fastdds_discovery_server(
         }
     }
 
-    fastdds::rtps::GuidPrefix_t guid_prefix = participantQos.wire_protocol().prefix;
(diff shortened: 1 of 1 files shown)


tools/fds/server.cpp
```diff
@@ -596,9 +595,15 @@ int fastdds_discovery_server(
 #endif // ifndef _WIN32
 
         bool has_security = false;
-        if (guid_prefix != pServer->guid().guidPrefix)
+        auto check_qos = pServer->get_qos();
+        // Access the properties of the participant QoS
+        for (auto prop : check_qos.properties().properties())
```
MiguelCompany, line 600:
```suggestion
        const auto& check_qos = pServer->get_qos();
        // Access the properties of the participant QoS
        for (const auto& prop : check_qos.properties().properties())
```

MiguelCompany (changes requested):

tools/fds/server.cpp
```diff
@@ -596,9 +595,15 @@ int fastdds_discovery_server(
...
         {
-            has_security = true;
+            if (prop.name() == "dds.sec.auth.builtin.PKI-DH.identity_ca" && prop.value() != "")
```
juanlofer-eprosima, line 602: NIT: perhaps we should check if all the mandatory security properties are present.

juanlofer-eprosima (changes requested):

tools/fds/server.cpp
cferreiragonz: I assumed that if this property was set but there were still some missing properties, the participant creation would fail and we would never reach this point of the code. Not sure if this would be the case

MiguelCompany (approved):

tools/fds/server.cpp
juanlofer-eprosima: You were right, it fails if only setting one of them.

PR #5764, opened 2025-04-14
[22814] Improve DS routines

cferreiragonz:
## Description

This PR aims to improve efficiency of DS routines by reducing the number of messages sent (Data(p), Data(r) and Data(w)). It does so by:

- Avoiding re-sending the server's Data(p) to relevant participants every time its own Data(p) is added to the `pdp_queue` (when a new client is discovered). Now a direct message to the new discovered client is sent instead of adding the server's Data(p) to the queue, which led to multiple sends to all direct clients. 
- Sending just one Data(p) and Data(r/w) to direct clients by considering a three-state value (unmatched, sent & matched) for `DiscoveryParticipantsAckStatus` instead of just a boolean. In this way the Discovery DB avoids re-sending repeated data when processing topics.

Related PR: 

- https://github.com/eProsima/Discovery-Server/pull/115

@Mergifyio backport 3.1.x 2.14.x 2.10.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
+++ b/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
@@ -282,8 +282,8 @@ void DiscoveryDataBase::update_change_and_unmatch_(
     changes_to_release_.push_back(entity.update_and_unmatch(new_change));
     // Manually set relevant participants ACK status of this server, and of the participant that sent the
     // change, to 1. This way, we avoid backprogation of the data.
-    entity.add_or_update_ack_participant(server_guid_prefix_, ParticipantState::MATCHED);
-    entity.add_or_update_ack_participant(new_change->writerGUID.guidPrefix, ParticipantState::MATCHED);
+    entity.add_or_update_ack_participant(server_guid_prefix_, ParticipantState::ACKED);
+    entity.add_or_update_ack_participant(new_change->writerGUID.guidPrefix, ParticipantState::ACKED);
 }
 
 void DiscoveryDataBase::add_ack_(
@@ -307,7 +307,7 @@ void DiscoveryDataBase::add_ack_(
             // database has been updated, so this ACK is not relevant anymore
             if (it->second.change()->write_params.sample_identity() == change->write_params.sample_identity())
             {
-                it->second.add_or_update_ack_participant(acked_entity, ParticipantState::MATCHED);
+                it->second.add_or_update_ack_participant(acked_entity, ParticipantState::ACKED);
             }
         }
     }
@@ -322,7 +322,7 @@ void DiscoveryDataBase::add_ack_(
             // database has been updated, so this ACK is not relevant anymore
             if (it->second.change()->write_params.sample_identity() == change->write_params.sample_identity())
             {
-                it->second.add_or_update_ack_participant(acked_entity, ParticipantState::MATCHED);
+                it->second.add_or_update_ack_participant(acked_entity, ParticipantState::ACKED);
             }
         }
     }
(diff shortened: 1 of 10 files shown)


src/cpp/rtps/builtin/discovery/database/DiscoveryParticipantsAckStatus.hpp
```diff
@@ -45,15 +45,25 @@ class DiscoveryParticipantsAckStatus
 
     ~DiscoveryParticipantsAckStatus() = default;
 
+    enum class ParticipantState : uint8_t
```
Mario-DL, line 48: What do you think about changing the naming here to
```
enum class DiscoveryInfoState
{
  PENDING_SEND, // Data(P) ....
  WAITING_ACK,
  ACKED
};
```

I think it could be more congruent with the context

src/cpp/rtps/builtin/discovery/database/DiscoveryParticipantsAckStatus.hpp
```diff
@@ -45,15 +45,25 @@ class DiscoveryParticipantsAckStatus
...
             const GuidPrefix_t& guid_p,
-            bool status);
+            ParticipantState status);
 
     void remove_participant(
             const GuidPrefix_t& guid_p);
 
     void unmatch_all();
 
+    bool is_sent(
```
Mario-DL, line 64: If the previous comment is accepted, this would be either a `is_pending_send()` or even `is_waiting_ack()`

src/cpp/rtps/builtin/discovery/participant/PDPServer.cpp
```diff
@@ -1620,7 +1620,26 @@ void PDPServer::send_announcement(
             EPROSIMA_LOG_ERROR(RTPS_PDP_SERVER, "Error sending announcement from server to clients");
         }
     }
+}
+
+void PDPServer::send_own_pdp(
+        ParticipantProxyData* pdata)
+{
+    std::vector<GUID_t> remote_readers;
```
Mario-DL, line 1628: Should we add the `if (!backup_in_progress())` at the beginning of the method ?

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1481,23 +1483,30 @@ bool DiscoveryDataBase::process_dirty_topics()
                         // Check the status of the writer in `readers_[reader]::relevant_participants_builtin_ack_status`.
```
Mario-DL, line 1483: I think that we can move here the `readers_it = readers_.find(reader)` and prevent the `find()` if participants are not matched

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1513,23 +1522,30 @@ bool DiscoveryDataBase::process_dirty_topics()
                         // Check the status of the reader in `writers_[writer]::relevant_participants_builtin_ack_status`.
```
Mario-DL, line 1522: Same here would be with the `writers_it`

Mario-DL (changes requested):
The approach looks good to me.
Apart from the suggestions below it would be nice if we could  add a test. We can make use of [test_UDPTransport](https://github.com/eProsima/Fast-DDS/blob/master/src/cpp/rtps/transport/test_UDPv4Transport.h) to detect that the Data(P,W,R) are only sent one time.

cferreiragonz:
In spite of the changes requested, I have added as first two commits the tests to check each scenario solved by this PR.

src/cpp/rtps/builtin/discovery/participant/PDPServer.cpp
Mario-DL: Then I think we can leave it as it is

Mario-DL (approved):
LGTM, 
Please double check the contributor checklist. I think we could backport this improvement

cferreiragonz:
@Mergifyio backport 3.1.x 2.14.x 2.10.x

PR #5780, opened 2025-04-23
[23088] Solve Discovery Server race conditions

juanlofer-eprosima:
## Description

This PR attempts to fix a couple of race conditions detected in scenarios where a large number of clients are present:
- PDP and EDP messages are stored in two queues which are not atomically swapped, which could create situations such as processing a data r/w of a data P already received but to be processed in the next server routine iteration.
- When a data UP is received (or manually inserted if a participant is dropped), it remains in the internal participants map but with its change updated. The item will be deleted only after the data UP is acked by all participants. If a new data P is received before this occurs, it will not be correctly processed. As a result, when the data UP is acked the item will be deleted from the participants map, and if a data r/w was to be processed afterwards, this would fail and print one of these errors "Reader/Writer has no associated participant. Skipping" or "Matching unexisting participant from reader/writer".
- When a reader/writer is being inserted to the database, checking if the corresponding participant is present is done after insertion, instead of as a first step. And even if the participant is present, it should be checked if it is alive to abort otherwise.

@Mergifyio backport 3.1.x 2.14.x 2.10.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
+++ b/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
@@ -654,7 +654,7 @@ void DiscoveryDataBase::match_new_server_(
         }
     }
     // The resources needed for TCP new connections are created during the matching process when the
-    // DATA(p) is receieved by each server.
+    // DATA(p) is received by each server.
 
     // Create virtual endpoints
     create_virtual_endpoints_(participant_prefix);
@@ -782,11 +782,46 @@ void DiscoveryDataBase::update_participant_from_change_(
 {
     fastdds::rtps::GUID_t change_guid = guid_from_change(ch);
 
+    assert(ch->kind == eprosima::fastdds::rtps::ALIVE);
+
+    // If the change corresponds to a previously removed participant, update map with new data and behave as if it was a
+    // new participant. Remove also the old change from the disposals collection, if it was added just before
+    if (participant_info.change()->kind != eprosima::fastdds::rtps::ALIVE)
+    {
+        // If it is local and server we have to create virtual endpoints, except for our own server
+        if (change_guid.guidPrefix != server_guid_prefix_ && !change_data.is_client() && change_data.is_local())
+        {
+            // Match new server and create virtual endpoints
+            match_new_server_(change_guid.guidPrefix, change_data.is_superclient());
+        }
+
+        // Update the change data
+        participant_info.participant_change_data(change_data);
+
+        // Remove old change from disposals if it was added just before to avoid sending data UP
+        auto it = std::find(disposals_.begin(), disposals_.end(), participant_info.change());
+        if (it != disposals_.end())
+        {
+            disposals_.erase(it);
+        }
+
+        // Update change. This should add the UNALIVE change to changes_to_release_, which should later both remove the
+        // change from the writer's history and release the change
+        update_change_and_unmatch_(ch, participant_info);
+
+        // Treat as a new participant found
+        new_updates_++;
+        if (change_guid.guidPrefix != server_guid_prefix_)
+        {
+            server_acked_by_all(false);
+        }
+    }
+
     // Specific case when a Data(P) from an entity A known as remote comes from the very entity A (we have
     // the Data(P) because of other server B, but now it arrives from A itself)
     // The entity A changes to local
     // Must be local data, or else it is a remote endpoint and should not be changed
-    if (change_guid.guidPrefix != server_guid_prefix_ && change_data.is_local() &&
+    else if (change_guid.guidPrefix != server_guid_prefix_ && change_data.is_local() &&
             DiscoveryDataBase::participant_data_has_changed_(participant_info, change_data))
     {
         // If the participant changes to server local, virtual endpoints must be added
@@ -807,7 +842,7 @@ void DiscoveryDataBase::update_participant_from_change_(
         server_acked_by_all(false);
 
         // It is possible that this Data(P) is in our history if it has not been acked by all
-        // In this case we have to resent it with the new update
+        // In this case we have to resend it with the new update
         if (!participant_info.is_acked_by_all())
         {
             add_pdp_to_send_(ch);
--- a/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
+++ b/src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
@@ -956,12 +956,14 @@ void DiscoveryDataBase::create_writers_from_change_(
         {
             EPROSIMA_LOG_ERROR(DISCOVERY_DATABASE,
                     "Writer " << writer_guid << " has no associated participant. Skipping");
+            changes_to_release_.push_back(ch); // Release change so it can be reused
             return;
         }
         else if (writer_part_it->second.change()->kind != fastdds::rtps::ChangeKind_t::ALIVE)
         {
             EPROSIMA_LOG_WARNING(DISCOVERY_DATABASE,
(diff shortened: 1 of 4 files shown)


src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -786,11 +780,46 @@ void DiscoveryDataBase::update_participant_from_change_(
 {
     fastdds::rtps::GUID_t change_guid = guid_from_change(ch);
 
+    assert(ch->kind == eprosima::fastdds::rtps::ALIVE);
+
+    // If the change corresponds to a previously removed participant, update map with new data and behave as if it was a
+    // new participant. Remove also the old change from the disposals collection, if it was added just before
+    if (participant_info.change()->kind != eprosima::fastdds::rtps::ALIVE)
```
juanlofer-eprosima, line 787: There is code repetition in this block and the two ones below, but there are some subtle differences I was not able to understand so I took the most conservative approach.

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -914,6 +943,25 @@ void DiscoveryDataBase::create_writers_from_change_(
...
+        // NOTE: Processing a data w should always be preceded by the reception and processing of its corresponding
+        // participant. However, one may receive a data w just after the participant has been removed, case in which the
+        // former should no longer be processed.
+        std::map<eprosima::fastdds::rtps::GuidPrefix_t, DiscoveryParticipantInfo>::iterator writer_part_it =
+                participants_.find(writer_guid.guidPrefix);
+        if (writer_part_it == participants_.end())
+        {
+            EPROSIMA_LOG_ERROR(DISCOVERY_DATABASE,
+                    "Writer " << writer_guid << " has no associated participant. Skipping");
+            return;
```
juanlofer-eprosima, line 956: Should we add the change to be release in the failing cases?

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -780,11 +780,46 @@ void DiscoveryDataBase::update_participant_from_change_(
...
+    {
+        // If it is local and server we have to create virtual endpoints, except for our own server
+        if (change_guid.guidPrefix != server_guid_prefix_ && !change_data.is_client() && change_data.is_local())
+        {
+            // Match new server and create virtual endpoints
+            match_new_server_(change_guid.guidPrefix, change_data.is_superclient());
+        }
+
+        // Update the change data
+        participant_info.participant_change_data(change_data);
```
cferreiragonz, line 797: If this case only occurs when a non-modified server (meaning it is the same server that initially connected to us) is discovered again, then this should not be necessary as the `DiscoveryParticipantChangeData` should be equal. 

If this is not the case and we can fall here in different scenarios, maybe its better just to delete the participant map entry, remove from disposals queue any unsent change and just call to `create_new_participant_from_change_` .

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -786,11 +780,46 @@ void DiscoveryDataBase::update_participant_from_change_(
...
+        // change from the writer's history and release the change
+        update_change_and_unmatch_(ch, participant_info);
+
+        // Treat as a new participant found
+        new_updates_++;
+        if (change_guid.guidPrefix != server_guid_prefix_)
+        {
+            server_acked_by_all(false);
+        }
+    }
```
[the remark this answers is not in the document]
cferreiragonz: In case of a CLIENT or SUPER_CLIENT, this is done in `PDPServer::assignRemoteEndpoints`.  In case of a SERVER, it is done in `match_new_server_` . So it should not be needed

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1032,6 +1069,25 @@ void DiscoveryDataBase::create_readers_from_change_(
     // The reader was NOT known by the database
     else
     {
+        // Check if corresponding participant is known, abort otherwise
+        // NOTE: Processing a data r should always be preceded by the reception and processing of its corresponding
```
cferreiragonz, line 1073:
```suggestion
        // NOTE: Processing a DATA(r) should always be preceded by the reception and processing of its corresponding
```

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -914,6 +943,25 @@ void DiscoveryDataBase::create_writers_from_change_(
     // The writer was NOT known by the database
     else
     {
+        // Check if corresponding participant is known, abort otherwise
+        // NOTE: Processing a data w should always be preceded by the reception and processing of its corresponding
```
cferreiragonz, line 947:
```suggestion
        // NOTE: Processing a DATA(w) should always be preceded by the reception and processing of its corresponding
```

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
cferreiragonz: I agree that this is the best approach. Although we might clarify a bit more that this case only occurs when a participant has been marked as removed but it has not yet received all the clients' ACK of the Data(Up). Just to make clearer how we can receive a Data(p) of a removed participant when it is in the `participants_` map

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
cferreiragonz: As we are checking the `participants_` map before creating the `tmp_writer` and adding it to the `writers_` map I would say no. We have not make a copy of the pointer yet.
However, I don't know why it was not being done before and when was this change being released. Maybe this was a source of error and performance decay.

cferreiragonz (changes requested):
Changes look good to me, I will proceed to make some tests now.
Leaving some suggestions tho. Also, we should rebase this PR.

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
juanlofer-eprosima: I thought it is better to update the discovery participant data in case the list of locators changes, which could happen in scenarios where the GUID is fixed and network is dynamic (e.g. Kubernetes).

And I would prefer to leave the method as it is because:
- I would like to keep the removal of items from the participants map dedicated to the acknowledgements check subtask.
- I wouldn't want to overcomplicate the code.
- It is more efficient to update a map entry than to remove + insert.

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
juanlofer-eprosima: I have checked there are no leaks without adding the change to the release collection. For the successful case (the associated participant is in the map), changes are not released until the pool is destroyed, same situation as with the inconsistent database scenario. As you point out, perhaps this is something that should be changed not to waste resources, but not in this PR.

juanlofer-eprosima: I eventually tested the release of changes corresponding to endpoints with no associated participants, and it seems to be fine (changes are released in the server routine, instead of on destruction), so I am including this as well.

PR #5795, opened 2025-05-05
[23078] Transmit only non-default QoS in EDP messages

cferreiragonz:
## Description

This PR introduces new features and behavior changes:

- QoS policies are now only added to EDP messages if they use a non-default value. Template method `should_be_sent` has been added to verify it. 
- `HistoryQoS` policy is now included in EDP messages in case it does not use the default value. 
- New property `fastdds.serialize_optional_qos` to extend EDP messages with eProsima Extensions QoS policies.

## Contributor Checklist

- _N/A_ Changes are API compatible. 
    
## Reviewer Checklist


src/cpp/rtps/builtin/data/WriterProxyData.cpp
```diff
@@ -242,79 +242,78 @@ uint32_t WriterProxyData::get_serialized_size(
         // PID_PERSISTENCE_GUID
         ret_val += 4 + PARAMETER_GUID_LENGTH;
     }
-    if (durability.send_always() || durability.hasChanged)
```
Mario-DL, line 245: Now that `hasChanged`, `send_always_` and `send_always()` are deprecated, would it be possible to clean up all references (basically assignments) to them?
Also, it may be a good idea to add a FASTDDS_TODO_BEFORE(4,..) remove them from `QosPolicy` class

src/cpp/rtps/builtin/data/WriterProxyData.cpp
```diff
@@ -352,6 +351,76 @@ uint32_t WriterProxyData::get_serialized_size(
...
+            });
+
+        if (serialize_opt_qos_property != properties.end())
+        {
+            if (serialize_opt_qos_property->second() == "true")
+            {
+                serialize_optional_qos = true;
+            }
+        }
+    }
```
Mario-DL, line 378: As we internally agreed, we could move the property to be in the `participant` and leave a boolean `should_send_optional_qos` in the Reader/Writer proxy datas to be filled after the proxy datas creations.

We should also check for "1" value

src/cpp/fastdds/core/policy/QosPoliciesSerializer.hpp
```diff
@@ -63,6 +63,22 @@ class QosPoliciesSerializer
         return ParameterSerializer<QosPolicy>::cdr_serialized_size(qos_policy);
     }
 
+    static bool should_be_sent(
```
Mario-DL, line 66: In order to distinguish reader from writer here we could add a `const EntityId& entity_id`

src/cpp/fastdds/core/policy/QosPoliciesSerializer.hpp
```diff
@@ -1296,8 +1330,11 @@ inline bool QosPoliciesSerializer<RTPSEndpointQos>::add_to_cdr_message(
         const RTPSEndpointQos& qos_policy,
         rtps::CDRMessage_t* cdr_message)
 {
-    bool valid = true;
+    // Add common to cdr_message
+    bool valid = rtps::CDRMessage::addUInt16(cdr_message, PID_RTPS_ENDPOINT);
+    valid &= rtps::CDRMessage::addUInt16(cdr_message, static_cast<uint16_t>(cdr_serialized_size(qos_policy) - 4));
```
Mario-DL, line 1335: For the next major we could also add a `TODO` of making all eprosima extended Qos to inherit from `QosPolicy`. Currently only some of them are doing so. What do you think ?

test/unittest/rtps/builtin/BuiltinDataSerializationTests.cpp
```diff
@@ -2767,6 +2767,188 @@ TEST(BuiltinDataSerializationTests, optional_qos_extensions_writer)
     ASSERT_EQ(wpd.writer_resource_limits->reader_filters_allocation.increment, 6u);
 }
 
+/**
+ * This test checks that a correct serialization is obtained
+ * when non-default eProsima's optional qos extensions are used in ReaderProxyData.
+ * History Qos is additionally tested.
+ */
+TEST(BuiltinDataSerializationTests, optional_qos_extensions_reader_serialize_non_default_values)
```
Mario-DL, line 2775: Apart from these unittests which look good to me. It would be nice if we could add a couple of BB tests that:

* Check that only non-default qos are sent 
* Property of optional qos at DDS level

We may use the test transport for that.

Mario-DL (changes requested):
Good job overall, leaving some notes which we basically commented internally.
In addition, this PR should be paired with one in docs describing the new property

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -464,24 +464,24 @@ ReturnCode_t DataWriterImpl::enable()
...
     {
-        std::istringstream partition_string(*endpoint_partitions);
-        std::string partition_name;
-        wqos.m_partition.clear();
+        EPROSIMA_LOG_ERROR(DATA_WRITER, "Could not register writer on discovery protocols");
+        writer_->set_listener(nullptr);
+        RTPSDomain::removeRTPSWriter(writer_);
+        writer_ = nullptr;
+        history_.reset();
+        release_payload_pool();
```
cferreiragonz, line 481: Not sure about these lines. It was not being done before. Should we add them or not?

cferreiragonz:
In order to be able to transfer DDS QosPolicies to RTPS layer, new methods for `register_writer` and `register_reader` have been created, which use `PublicationBuiltinTopicData` and `SubscriptionBuiltinTopicData` as arguments instead of `WriterQos` and `ReaderQos`, respectively. Old methods, however, have not been deleted to avoid API breaks, but will be deprecated in next major as it was planned.

The `should_send_opt_qos` argument added to methods in commit d01128ba9 can be replaced by adding a new boolean attribute to `PublicationBuiltinTopicData` and `SubscriptionBuiltinTopicData` in next minor too.

src/cpp/rtps/builtin/BuiltinProtocols.h
```diff
@@ -141,6 +141,23 @@ class BuiltinProtocols
...
+     * Add a local writer to the BuiltinProtocols.
+     *
+     * @param writer                  Pointer to the RTPSWriter
+     * @param topic                   Information regarding the topic where the writer is registering
+     * @param pub_builtin_topic_data  Information on the publication endpoint
+     * @param should_send_opt_qos     True if optional QoS policies should be sent.
+     *
+     * @return True if correct.
+     */
+    bool add_writer(
```
MiguelCompany, line 155: Would be better to return a code instead of a boolean

src/cpp/rtps/builtin/BuiltinProtocols.h
```diff
@@ -156,6 +173,23 @@ class BuiltinProtocols
...
+     * Add a local reader to the BuiltinProtocols.
+     *
+     * @param rtps_reader             Pointer to the RTPSReader.
+     * @param topic                   Information regarding the topic where the writer is registering
+     * @param sub_builtin_topic_data  Information on the subscription endpoint
+     * @param should_send_opt_qos     True if optional QoS policies should be sent.
+     * @param content_filter          Optional content filtering information.
+     * @return True if correct.
+     */
+    bool add_reader(
```
MiguelCompany, line 186: Would be better to return a code instead of a boolean

src/cpp/rtps/builtin/data/PublicationBuiltinTopicData.cpp
```diff
@@ -27,6 +27,7 @@ namespace rtps {
 PublicationBuiltinTopicData::PublicationBuiltinTopicData()
 {
     reliability.kind = dds::RELIABLE_RELIABILITY_QOS;
+    durability.kind = dds::TRANSIENT_LOCAL_DURABILITY_QOS;
```
MiguelCompany, line 30: Wrong. The default is volatile

src/cpp/rtps/builtin/data/PublicationBuiltinTopicData.cpp
```diff
@@ -39,6 +40,7 @@ PublicationBuiltinTopicData::PublicationBuiltinTopicData(
     partition.set_max_size(static_cast<uint32_t>(data_limits.max_partitions));
     data_sharing.set_max_domains(static_cast<uint32_t>(data_limits.max_datasharing_domains));
     reliability.kind = dds::RELIABLE_RELIABILITY_QOS;
+    durability.kind = dds::TRANSIENT_LOCAL_DURABILITY_QOS;
```
MiguelCompany, line 43: Same here

MiguelCompany (changes requested):

src/cpp/fastdds/core/policy/QosPoliciesSerializer.hpp
```diff
@@ -84,6 +125,20 @@ class QosPoliciesSerializer
...
+        const DurabilityQosPolicy& qos_policy,
+        bool is_writer)
+{
+    if (is_writer)
+    {
+        // Writer is Transient Local by default
+        return (qos_policy.kind != TRANSIENT_LOCAL_DURABILITY_QOS);
+    }
+    // Reader is Volatile by default
+    return (qos_policy.kind != VOLATILE_DURABILITY_QOS);
```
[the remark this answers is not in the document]
cferreiragonz: As argued internally, durability will always be sent

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -283,33 +283,21 @@ ReturnCode_t DataReaderImpl::enable()
...
         filter_property = &content_topic->filter_property;
     }
-    if (!subscriber_->rtps_participant()->register_reader(
-                reader_,
-                topic_desc,
-                rqos,
-                filter_property))
+
+    RTPSParticipantImpl* rtps = RTPSDomainImpl::find_local_participant(subscriber_->rtps_participant()->getGuid());
+    if (rtps->register_reader(reader_, topic_desc, subscription_data, filter_property) != RETCODE_OK)
```
MiguelCompany, line 300: Return the code returned by `register_reader`

MiguelCompany (changes requested):

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -1428,6 +1438,44 @@ bool RTPSParticipantImpl::register_reader(
...
+            });
+
+        if (serialize_opt_qos_property != m_att.properties.properties().end())
+        {
+            if (serialize_opt_qos_property->value() == "true" ||
+                    serialize_opt_qos_property->value() == "1")
+            {
+                should_send_opt_qos = true;
+            }
+        }
```
Mario-DL, line 1474: I think we could simplify this by taking advantage of existing methods, and implement the `PropertyParser::as_bool()` 
```suggestion
        const Property* const serialize_optional_qos_property =
            PropertyPolicyHelper::get_property(m_att.properties, fastdds::dds::parameter_serialize_optional_qos);
        
        if (serialize_optional_qos_property != nullptr)
        {
            should_send_opt_qos = PropertyParser::as_bool(*serialize_optional_qos_property, exception("..."));
        }
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -464,25 +464,25 @@ ReturnCode_t DataWriterImpl::enable()
...
-    if (!is_data_sharing_compatible_)
+    PublicationBuiltinTopicData publication_data;
+    if (get_publication_builtin_topic_data(publication_data) != RETCODE_OK)
     {
-        wqos.data_sharing.off();
+        EPROSIMA_LOG_ERROR(DATA_WRITER, "Error getting publication data. RTPS Writer not enabled.");
+        return RETCODE_ERROR;
     }
-    if (endpoint_partitions)
+    RTPSParticipantImpl* rtps = RTPSDomainImpl::find_local_participant(publisher_->rtps_participant()->getGuid());
```
Mario-DL, line 473: Would it be possible to expose the new `register_writer` and `register_reader` methods in the `RTPSParticipant` ? In that way this would simply be instead of doing a global search for the participant. Lets leverage that we already have a pointer to the RTPSParticipant in the Publisher.
Same suggestion would apply in the Subscriber
```
 publisher_->rtps_participant()->register_writer(writer_, topic_desc, publication_data);
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
Mario-DL: I think we could simply log an error and return the `register_writer_code`, but not cleaning. Upon a successful creation, resources are already tracked in the lists of `RTPSDomain` and `RTPSParticipant` (i.e m_RTPSParticipants, m_allWritersLists, m_allReadersLists). Hence, it should take care of cleaning everything by itself at destruction.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -464,25 +464,25 @@ ReturnCode_t DataWriterImpl::enable()
...
+    if (register_writer_code != RETCODE_OK)
     {
-        std::istringstream partition_string(*endpoint_partitions);
-        std::string partition_name;
-        wqos.m_partition.clear();
-
-        while (std::getline(partition_string, partition_name, ';'))
-        {
-            wqos.m_partition.push_back(partition_name.c_str());
-        }
```
Mario-DL, line 481: I think I dont fully understand why we are removing the partitions checks from here

Mario-DL (changes requested):
Nice effort with the tests and general cleanup.
Leaving some suggestions

src/cpp/fastdds/publisher/DataWriterImpl.cpp
Mario-DL: I think we could simply log an error and return the register_writer_code, but not cleaning. Upon a successful creation, resources are already tracked in the lists of RTPSDomain and RTPSParticipant (i.e m_RTPSParticipants, m_allWritersLists, m_allReadersLists). Hence, it should take care of cleaning everything by itself at destruction.

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
cferreiragonz: I have added the method `as_bool` to the `ProperyParser`

src/cpp/fastdds/publisher/DataWriterImpl.cpp
cferreiragonz: Would not that break ABI? We will be modifying `RTPSParticipant.hpp`.

Mario-DL: It should not be breaking ABI since we are only extending public API and keeping the old methods unaltered

Mario-DL: We have another alternative of reaching the `rtps_participant_impl()`, which would be 
`writer_->get_participant_impl()` but honestly, I think that adding the new register_* methods in the public API of the RTPSParticipant is useful

Mario-DL (approved):
LGTM with green CI 🚀

MiguelCompany (approved):

PR #5814, opened 2025-05-14
[23042] Bugfix: Properly delete secure endpoints if registration fails

Mario-DL:
## Description

This PR fixes a bug in which, when secure writers or readers failed to register, cleanup is not properly performed.

No need for backport since `local_actions_on_.._removed()` was introduced in `3.2.x`.

## Contributor Checklist

- NO Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
    
- NO Applicable backports have been included in the description.

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
+++ b/src/cpp/rtps/participant/RTPSParticipantImpl.cpp
@@ -975,6 +975,7 @@ bool RTPSParticipantImpl::create_writer(
         if (!m_security_manager.register_local_writer(SWriter->getGuid(),
                 param.endpoint.properties, SWriter->getAttributes().security_attributes()))
         {
+            SWriter->local_actions_on_writer_removed();
             delete(SWriter);
             return false;
         }
@@ -984,6 +985,7 @@ bool RTPSParticipantImpl::create_writer(
         if (!m_security_manager.register_local_builtin_writer(SWriter->getGuid(),
                 SWriter->getAttributes().security_attributes()))
         {
+            SWriter->local_actions_on_writer_removed();
             delete(SWriter);
             return false;
         }
@@ -1111,6 +1113,7 @@ bool RTPSParticipantImpl::create_reader(
         if (!m_security_manager.register_local_reader(SReader->getGuid(),
                 param.endpoint.properties, SReader->getAttributes().security_attributes()))
         {
+            SReader->local_actions_on_reader_removed();
             delete(SReader);
             return false;
         }
@@ -1120,6 +1123,7 @@ bool RTPSParticipantImpl::create_reader(
         if (!m_security_manager.register_local_builtin_reader(SReader->getGuid(),
                 SReader->getAttributes().security_attributes()))
         {
+            SReader->local_actions_on_reader_removed();
             delete(SReader);
             return false;
         }
--- a/test/blackbox/common/RTPSBlackboxTestsSecurity.cpp
+++ b/test/blackbox/common/RTPSBlackboxTestsSecurity.cpp
@@ -1,0 +1,100 @@
+// Copyright 2025 Proyectos y Sistemas de Mantenimiento SL (eProsima).
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


test/blackbox/common/RTPSBlackboxTestsSecurity.cpp
```diff
@@ -0,0 +1,100 @@
...
+                "file://" + std::string(certs_path) + "/maincacert.pem"));
+    props.emplace_back(Property("dds.sec.access.builtin.Access-Permissions.governance",
+            "file://" + std::string(certs_path) + "/" + governance_file));
+    props.emplace_back(Property("dds.sec.access.builtin.Access-Permissions.permissions",
+            "file://" + std::string(certs_path) + "/" + permissions_file));
+}
+
+/**
+ * This test checks the a StatefulWriter created with security
+ * with a wrong permissions on topic, should be properly cleaned up
```
cferreiragonz, line 69:
```suggestion
 * This test checks that a StatefulWriter created with security,
 * but with wrong permissions on topic, is properly cleaned up
```

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -1111,6 +1113,7 @@ bool RTPSParticipantImpl::create_reader(
         if (!m_security_manager.register_local_reader(SReader->getGuid(),
                 param.endpoint.properties, SReader->getAttributes().security_attributes()))
         {
+            SReader->local_actions_on_reader_removed();
```
cferreiragonz, line 1116: This does not have any real effect. Test passes without them.

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
```diff
@@ -975,6 +975,7 @@ bool RTPSParticipantImpl::create_writer(
         if (!m_security_manager.register_local_writer(SWriter->getGuid(),
                 param.endpoint.properties, SWriter->getAttributes().security_attributes()))
         {
+            SWriter->local_actions_on_writer_removed();
```
cferreiragonz, line 978: Should we add a call to `local_actions_on_writer_removed()` on the [StatefulWriter Destructor](https://github.com/eProsima/Fast-DDS/blob/ea69181efaa4315303da70a2df467849f8689e4e/src/cpp/rtps/writer/StatefulWriter.cpp#L253) to ensure that events are always destroyed, instead of manually calling it here?

cferreiragonz (changes requested):
If calling `local_actions_on_writer_removed()` from StatefulWriter destructor does not have any side-effects I would try to follow that approach.

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
Mario-DL: Yes, in the reader side it does not have any effect, but to be congruent I think we should call it too

src/cpp/rtps/participant/RTPSParticipantImpl.cpp
Mario-DL: The `local_actions_on_writer_removed()` comes from the former `deinit()` that was explicitly moved out from the `StatefulPersistentWriter` destructor, [in this commit](https://github.com/eProsima/Fast-DDS/commit/7458386c25c221adc99c44e78e2ad58896fe01c3#diff-680b6e42c54bd512d44efd64560acefa5890be8e6d8022f80fb61fdc1ad816d3L45) 

However I dont see any reason why we should not call it from the `StatafulWriter`

MiguelCompany: Calling virtual functions inside destructors is dangerous. We could have a proxy method that receives a pointer to a writer and performs the call to `local_actions_on_writer_removed()` and then deletes the writer

Mario-DL:
This latter commit only changes the test description, maybe with the former ci is enough

cferreiragonz (approved):

cferreiragonz:
CI passed in previous commit caeb9dd25c887755e4b3e7016f6fb524f5dafe43

PR #5849, opened 2025-06-03
[23253] Split windows workflow in different test suites

Mario-DL:
## Description

This PR splits the windows reusable workflow in building and testing. In turn, testing was divided in three jobs:
* `blackbox`: including `BlackboxTests|ParticipantTests|SecureDiscoverServer|SimpleCommunication|system.`
* `examples`: including `example_tests`
* `unittests-I`: including `DDSSQLFilterValue`. This test suite runs in parallel
* `unittests-II`: excluding the rest of the former ones

 Mind that the `fastdds_build` step does not build with `-DFASTDDS_EXAMPLE_TESTS` , instead it build with with `DFASTDDS_COMPILE_EXAMPLES` since generating the docker image for testing takes too long and it has to be regenerated again in the `fastdds_test` proper job again. 

@Mergifyio backport 2.14.x

## Contributor Checklist

- [X] Commit messages follow the project guidelines. 
- [X] The code follows the style guidelines of this project. 
- **N/A** Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- **N/A** Any new/modified methods have been properly documented using Doxygen. 
- **N/A** Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
- **N/A** Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- **N/A** Changes are API compatible. 
- **N/A** New feature has been added to the `versions.md` file (if applicable).
- **N/A** New feature has been documented/Current behavior is correctly described in the documentation. 
    
- [X] Applicable backports have been included in the description.

## Reviewer Checklist

- [x] The PR has a milestone assigned.
- [x] The title and description correctly express the PR's purpose.
- [x] Check contributor checklist is correct.
- [ ] If this is a critical bug fix, backports to the critical-only supported branches have been requested.
- [x] Check CI results: changes do not issue any warning.
- [x] Check CI results: failing tests are unrelated with the changes.

PR #5851, opened 2025-06-04
[22593] Include all example tests in windows CI

Mario-DL:
## Description

This PR includes all the remaining examples into the windows ci testing.
Some considerations worth mentioning:

* `test_configuration` was splitted into `3` suite of tests in order not to exceed the `300` timeout seconds configured in colcon.
* IPv6 test cases were disabled in both Windows and Ubuntu. Windows does not support it and Docker in Ubuntu CI supports only the ipv6 localhost (I tried to enable the experimental features but gave error). Since Docker in Windows neither supports `network_mode: host`, some examples were refactored to use custom `network` with fixed ips (for instance, the `discovery_server` example one) 
* Some test cases were disabled in some examples in windows to avoid flakiness. Specially those making use of SHM or DataSharing among dockers. The issue will keep being being investigated in a follow up PR.
* The Python `subprocess.check_output` does not properly work in Windows, when the `timeout` expires, subprocesses are not killed. This led me to also disable the `timeout` test cases for some of the example tests.
* Some extra additions were made to the `Publishers` to force waiting for `ack` in the first sample and mitigate flakiness.

**Note: This PR should be merged after** 
* #5849

## Contributor Checklist

- **N/A** Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- **N/A** Changes are API compatible. 
    
## Reviewer Checklist


examples/cpp/benchmark/PublisherApp.cpp
```diff
@@ -507,6 +506,18 @@ bool PublisherApp::publish()
...
         }
     }
+
+    if (benchmark_.index() == 0u)
+    {
+        ReturnCode_t acked = RETCODE_ERROR;
+        do
+        {
+            dds::Duration_t acked_wait{1, 0};
+            acked = writer_->wait_for_acknowledgments(acked_wait);
```
rsanchez15, line 516: Is there any timeout in case this never returns RETCODE_OK?

rsanchez15 (changes requested):

examples/cpp/benchmark/PublisherApp.cpp
Mario-DL: No in this case, the idea is to ensure that the first sample is acknowledged. However, the user can stop the application at any time with `Cntrl+C` which IMO is fair enough, should we then include an additional timeout here ?

Mario-DL:
In the `delivery_mechanisms` example I forced to `ack` the first and the last `sample` (just if `samples >0`). Hopefully it completely solves the flakiness of the test. In my local I run it 20 times without failure 🤞🏻

rsanchez15 (approved):

PR #5853, opened 2025-06-04
[23269] Branch out 3.3.x

Mario-DL:
## Description

This PR brings the branch out for 3.3.x including:

* Update mirror workflow to 3.3.x
* Include 3.2.x as supported branch for backporting improvements and fixes
* Add 3.2.x nightly workflows

*Note: Bumping project's version is left for the corresponding bump to 3.3.0*

## Contributor Checklist

- **N/A** The code follows the style guidelines of this project. 
- **N/A** Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- **N/A** Changes are API compatible. 
    
## Reviewer Checklist

- _N/A_ Check CI results: changes do not issue any warning.
- _N/A_ Check CI results: failing tests are unrelated with the changes.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -21,7 +21,7 @@
     In case of bug fixes, please provide the list of supported branches where this fix should be also merged.
     Please uncomment following line, adjusting the corresponding target branches for the backport.
 -->
-<!-- @Mergifyio backport 2.14.x -->
+<!-- @Mergifyio backport 3.2.x 2.14.x -->
 
 <!--
     In case of critical bug fix, please uncomment following line, adjusting the corresponding LTS target branches for the backport.
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         dest_branch:
-          - '3.2.x'
+          - '3.3.x'
           - '3.x'
     steps:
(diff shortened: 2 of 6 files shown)


MiguelCompany:
> Note: Bumping project's version is left for the corresponding bump to 3.3.0

We should do this so the generated libraries already have a new version in the filename.
Also to allow generated code to require version 3.3 when necessary.

Mario-DL:
> > Note: Bumping project's version is left for the corresponding bump to 3.3.0
> 
> We should do this so the generated libraries already have a new version in the filename. Also to allow generated code to require version 3.3 when necessary.

In that case, I think we should either address, remove or forward to 3.4.x all the [FASTDDS_TODO_BEFORE(3,3)](https://github.com/eProsima/Fast-DDS/blob/f0ce10aa91d645c15c9129e7b105310a4001c0bc/src/cpp/fastdds/rpc/ReplierImpl.cpp#L105) otherwise we will fail

MiguelCompany:
> In that case, I think we should either address, remove or forward to 3.4.x all the [FASTDDS_TODO_BEFORE(3,3)](https://github.com/eProsima/Fast-DDS/blob/f0ce10aa91d645c15c9129e7b105310a4001c0bc/src/cpp/fastdds/rpc/ReplierImpl.cpp#L105) otherwise we will fail

You are right, let's leave the bump till we have implemented those.

PR #5861, opened 2025-06-09
[23264] Feature: `DataWriter` sample prefilter

Mario-DL:
## Description

This PR brings the `prefiltering` feature in the `DataWriter`. This operation lets user to filter out destination readers from a particular sample based on the `SerializedPayload` and/or `WriteParams` by implementing the `IContentFilter` interface. 

It also extends the `FilteredSampleInfo` and the `WriteParams` structures to include a new member `UserWriteData` meant to be extended by the user.

## Contributor Checklist

- NO Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
    - Related documentation PR: eProsima/Fast-DDS-docs#1077
- NO Applicable backports have been included in the description.

## Reviewer Checklist

### DIFF
--- a/src/cpp/fastdds/publisher/DataWriterImpl.cpp
+++ b/src/cpp/fastdds/publisher/DataWriterImpl.cpp
@@ -347,9 +347,12 @@ ReturnCode_t DataWriterImpl::enable()
     bool filtering_enabled =
             qos_.liveliness().lease_duration.is_infinite() &&
             (0 < qos_.writer_resource_limits().reader_filters_allocation.maximum);
-    if (filtering_enabled)
     {
-        reader_filters_.reset(new ReaderFilterCollection(qos_.writer_resource_limits().reader_filters_allocation));
+        std::lock_guard<std::mutex> lock(filters_mtx_);
+        if (filtering_enabled && !reader_filters_)
+        {
+            reader_filters_.reset(new ReaderFilterCollection(qos_.writer_resource_limits().reader_filters_allocation));
+        }
     }
 
     // Set Datawriter's DataRepresentationId taking into account the QoS.
@@ -1504,24 +1507,8 @@ ReturnCode_t DataWriterImpl::set_sample_prefilter(
         std::shared_ptr<IContentFilter> prefilter)
 {
     {
-        std::lock_guard<std::mutex> lock(sample_prefilter_mutex_);
+        std::lock_guard<std::mutex> lock(filters_mtx_);
         sample_prefilter_ = prefilter;
-    }
-
-    if (writer_ != nullptr)
-    {
-        if (!reader_filters_)
-        {
-            std::lock_guard<RecursiveTimedMutex> lock(writer_->getMutex());
-            if (prefilter)
-            {
-                writer_->reader_data_filter(this);
-            }
-            else
-            {
-                writer_->reader_data_filter(nullptr);
-            }
-        }
     }
 
     return RETCODE_OK;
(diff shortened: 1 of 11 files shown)


src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -347,9 +347,12 @@ ReturnCode_t DataWriterImpl::enable()
...
             qos_.liveliness().lease_duration.is_infinite() &&
             (0 < qos_.writer_resource_limits().reader_filters_allocation.maximum);
-    if (filtering_enabled)
     {
-        reader_filters_.reset(new ReaderFilterCollection(qos_.writer_resource_limits().reader_filters_allocation));
+        std::lock_guard<std::mutex> lock(filters_mtx_);
+        if (filtering_enabled && !reader_filters_)
+        {
+            reader_filters_.reset(new ReaderFilterCollection(qos_.writer_resource_limits().reader_filters_allocation));
+        }
```
MiguelCompany: No need to check the value of `reader_filters_`, since it is only assigned here, and we cannot enable twice:
```suggestion
        if (filtering_enabled)
        {
            std::lock_guard<std::mutex> lock(filters_mtx_);
            reader_filters_.reset(new ReaderFilterCollection(qos_.writer_resource_limits().reader_filters_allocation));
        }
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1500,6 +1503,17 @@ ReturnCode_t DataWriterImpl::get_publication_matched_status(
     return RETCODE_OK;
 }
 
+ReturnCode_t DataWriterImpl::set_sample_prefilter(
+        std::shared_ptr<IContentFilter> prefilter)
+{
+    {
+        std::lock_guard<std::mutex> lock(filters_mtx_);
+        sample_prefilter_ = prefilter;
+    }
```
MiguelCompany:
```suggestion
    std::lock_guard<std::mutex> lock(filters_mtx_);
    sample_prefilter_ = prefilter;
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1500,6 +1503,17 @@ ReturnCode_t DataWriterImpl::get_publication_matched_status(
...
+
+    return RETCODE_OK;
```
MiguelCompany, line 1517: When standard filtering is not enabled, we are not setting the filter in the rtps writer, so we should set it here if necessary.

```suggestion
    if (prefilter)
    {
        writer_->reader_data_filter(this);
    }

    return RETCODE_OK;
```

test/blackbox/common/DDSBlackboxTestsDataWriter.cpp
```diff
@@ -1296,6 +1296,248 @@ TEST_P(DDSDataWriter, datawriter_sends_non_default_qos_optional)
...
+
+    auto data = default_helloworld_data_generator();
+
+    reader.startReception(data);
+
+    writer.send(data, 50, &write_params);
+    // Reader should have received the samples
+    ASSERT_EQ(reader.block_for_all(std::chrono::seconds(1)), 5u);
+}
+
```
MiguelCompany, line 1599: We need a test that checks that the prefilter works when the standard filtering is disabled (i.e. `reader_filters_allocation.maximum = 0`)

MiguelCompany (changes requested):

src/cpp/fastdds/publisher/DataWriterImpl.cpp
Mario-DL: Instead, we will always register the `DataWriterImpl` as the `reader_data_filter` implementer within the `enable()` method

PR #5890, opened 2025-06-19
[22973] Support modules in IDL Parser

Carlosespicur:
## Description

This PR adds support for scoped types in the IDL Parser. The main changes are as follows:

1. Introduced the `ModuleStack` class to manage hierarchies of nested modules.

2. Updated usage of `DynamicTypeBuilder::get_descriptor`. Since it returns a copy of the Type Descriptor instead of a `shared_ptr` to the original, the previous logic for updating the descriptor after builder instantiation was incorrect. This logic has been moved before the builder instantiation.

3. Internal builder maps now store types using fully qualified names, enabling support for types with the same name in different scopes.

4. Enabled previously commented API functions related to module tree management.

5. Fixed `resolve_scope` method.

6. Added new test cases not covered in `dds-types-test`:
   - Nested modules  
   - Scope resolution  
   - Scoped union discriminators

## Contributor Checklist

Related documentation PR:
- https://github.com/eProsima/Fast-DDS-docs/pull/1090
- [ ] Applicable backports have been included in the description.

## Reviewer Checklist

### DIFF
--- a/src/cpp/fastdds/xtypes/dynamic_types/idl_parser/IdlParser.hpp
+++ b/src/cpp/fastdds/xtypes/dynamic_types/idl_parser/IdlParser.hpp
@@ -394,9 +394,6 @@ struct action<scoped_name>
     {
         auto module = ctx->modules().current();
         std::string identifier_name = in.string();
-        // const std::string& name_space = module->scope();
-        // const std::string& scoped_identifier_name = name_space.empty() ?
-        //     identifier_name : name_space + "::" + identifier_name;
 
         if (state.count("arithmetic_expr"))
         {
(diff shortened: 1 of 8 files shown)


src/cpp/fastdds/xtypes/dynamic_types/idl_parser/IdlParser.hpp
```diff
@@ -1382,24 +1459,33 @@ struct action<struct_forward_dcl>
...
         const std::string& struct_name = state["struct_name"];
-        if (module.has_symbol(struct_name, false))
+
+        if (struct_name.find("::") != std::string::npos)
+        {
+            return; // Cannot add a symbol with scoped name.
+        }
+
+        const std::string& name_space = module->scope();
+        const std::string scoped_struct_name = name_space.empty() ? struct_name : name_space + "::" + struct_name;
```
MiguelCompany: This pattern repeats a lot. I would create a `std::string create_scoped_name(const std::string& plain_name)` method in `Module` class that will:
1. Throw a `runtime_error` exception when `plain_name` contains `::`.
2. Return the corresponding scoped name.

MiguelCompany: If possible, we should also add negative tests for this.

MiguelCompany (changes requested):

PR #5899, opened 2025-06-23
[23361] Feature: RPC enhanced discovery

Mario-DL:
## Description

This PR addresses the `FASTDDS_TODO_BEFORE(3,3)` feature of enhancing the discovery algorithm for RPC. 

## Contributor Checklist

- **NO** Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
    
## Reviewer Checklist

### DIFF
--- a/test/blackbox/api/dds-pim/ReqRepHelloWorldReplier.cpp
+++ b/test/blackbox/api/dds-pim/ReqRepHelloWorldReplier.cpp
@@ -41,7 +41,8 @@ ReqRepHelloWorldReplier::ReqRepHelloWorldReplier()
     , service_(nullptr)
     , participant_(nullptr)
     , initialized_(false)
-    , matched_(0)
+    , pub_matched_(0)
+    , sub_matched_(0)
 {
     request_processor_ = [](
         eprosima::fastdds::dds::rpc::RequestInfo& info,
@@ -68,7 +69,8 @@ ReqRepHelloWorldReplier::ReqRepHelloWorldReplier(
     , service_(nullptr)
     , participant_(nullptr)
     , initialized_(false)
-    , matched_(0)
+    , pub_matched_(0)
+    , sub_matched_(0)
     , request_processor_(request_processor)
 {
 }
@@ -139,17 +141,25 @@ void ReqRepHelloWorldReplier::wait_discovery()
 
     cvDiscovery_.wait(lock, [&]()
             {
-                return matched_ > 1;
+                return pub_matched_ > 0 && sub_matched_ > 0;
             });
 
     std::cout << "Replier discovery finished..." << std::endl;
 }
 
-void ReqRepHelloWorldReplier::matched()
+void ReqRepHelloWorldReplier::matched(
+        bool is_pub)
 {
     std::unique_lock<std::mutex> lock(mutexDiscovery_);
-    ++matched_;
-    if (matched_ > 1)
+    if (is_pub)
+    {
+        ++pub_matched_;
+    }
+    else
+    {
+        ++sub_matched_;
+    }
+    if (pub_matched_ > 0 && sub_matched_ > 0)
     {
         cvDiscovery_.notify_one();
     }
@@ -208,9 +218,9 @@ void ReqRepHelloWorldReplier::process_status_changes()
 
                     // status.current_count_change is shadowed by the internal entity listeners
                     // so check also the current_count
-                    if (status.current_count_change > 0 || status.current_count > 0)
+                    if (status.current_count_change > 0 || status.current_count > (int32_t)pub_matched_)
                     {
-                        matched();
+                        matched(true);
                     }
                 }
                 else if (status_changes.is_active(StatusMask::subscription_matched()))
@@ -230,9 +240,9 @@ void ReqRepHelloWorldReplier::process_status_changes()
 
                     // status.current_count_change is shadowed by the internal entity listeners
                     // so check also the current_count
-                    if (status.current_count_change > 0 || status.current_count > 0)
+                    if (status.current_count_change > 0 || status.current_count > (int32_t)sub_matched_)
                     {
-                        matched();
+                        matched(false);
                     }
                 }
                 else if (status_changes.is_active(StatusMask::data_available()))
--- a/test/blackbox/api/dds-pim/ReqRepHelloWorldReplier.hpp
+++ b/test/blackbox/api/dds-pim/ReqRepHelloWorldReplier.hpp
@@ -70,7 +70,8 @@ public:
 
     void wait_discovery();
 
-    void matched();
+    void matched(
+            bool is_pub);
 
     eprosima::fastdds::dds::ReplierQos create_replier_qos();
 
@@ -92,7 +93,8 @@ private:
 
     std::mutex mutexDiscovery_;
     std::condition_variable cvDiscovery_;
-    unsigned int matched_;
+    unsigned int pub_matched_;
+    unsigned int sub_matched_;
 
     // Entity status changes are managed using the WaitSet on a different thread
     // The main thread remains blocked until the requester matches with the replier
--- a/test/blackbox/api/dds-pim/ReqRepHelloWorldRequester.cpp
+++ b/test/blackbox/api/dds-pim/ReqRepHelloWorldRequester.cpp
@@ -44,7 +44,8 @@ ReqRepHelloWorldRequester::ReqRepHelloWorldRequester()
     , service_(nullptr)
     , participant_(nullptr)
     , initialized_(false)
-    , matched_(0)
+    , pub_matched_(0)
+    , sub_matched_(0)
 {
 }
 
@@ -155,17 +156,26 @@ void ReqRepHelloWorldRequester::wait_discovery()
 
     cvDiscovery_.wait(lock, [&]()
             {
-                return matched_ > 1;
+                return pub_matched_ > 0 && sub_matched_ > 0;
             });
 
     std::cout << "Requester discovery finished..." << std::endl;
 }
 
-void ReqRepHelloWorldRequester::matched()
+void ReqRepHelloWorldRequester::matched(
+        bool is_pub)
 {
     std::unique_lock<std::mutex> lock(mutexDiscovery_);
-    ++matched_;
-    if (matched_ > 1)
+    if (is_pub)
+    {
+        ++pub_matched_;
+    }
+    else
+    {
+        ++sub_matched_;
+    }
+
+    if (pub_matched_ > 0 && sub_matched_ > 0)
     {
         cvDiscovery_.notify_one();
     }
@@ -327,9 +337,9 @@ void ReqRepHelloWorldRequester::process_status_changes()
 
                     // status.current_count_change is shadowed by the internal entity listeners
                     // so check also the current_count
-                    if (status.current_count_change > 0 || status.current_count > 0)
+                    if (status.current_count_change > 0 || status.current_count > (int32_t)pub_matched_)
                     {
-                        matched();
+                        matched(true);
                     }
                 }
                 else if (status_changes.is_active(StatusMask::subscription_matched()))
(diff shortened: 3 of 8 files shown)


src/cpp/fastdds/rpc/ReplierImpl.cpp
```diff
@@ -137,6 +137,13 @@ ReturnCode_t ReplierImpl::take_request(
         return RETCODE_PRECONDITION_NOT_MET;
     }
 
+    // For taking the request is enough to check if the replier is fully matched
+    if (!is_fully_matched())
+    {
+        EPROSIMA_LOG_ERROR(REPLIER, "Trying to take a request with an unmatched requester");
```
Mario-DL: I think I should make this at least a warning since it could spuriously fail on applications shutdown

include/fastdds/dds/publisher/DataWriter.hpp
```diff
@@ -613,6 +613,23 @@ class DataWriter : public DomainEntity
...
+     * This operation sets the key of the DataReader that is related to this DataWriter.
+     * This is used to establish a relationship between a DataReader and a DataWriter
+     * in the context of RPC over DDS.
+     *
+     * @warning This operation is only valid if the entity is not enabled.
+     *
+     * @param [in] related_reader_guid The GUID of the DataReader to set as related.
+     *
+     * @return RETCODE_OK if the key is set successfully.
+     * @return RETCODE_ERROR if this entity is enabled.
```
MiguelCompany: I would use ILEGAL_OPERATION instead of ERROR

include/fastdds/dds/publisher/DataWriter.hpp
```diff
@@ -613,6 +613,23 @@ class DataWriter : public DomainEntity
...
+     * @return RETCODE_BAD_PARAMETER if the provided GUID is unknown
+     * or does not correspond to a DataReader.
+     */
+    FASTDDS_EXPORTED_API ReturnCode_t set_related_datareader_key(
+            const rtps::GUID_t& related_reader_guid);
```
MiguelCompany: I would pass the reader directly, and check that they belong to the same participant (return PRECONDITION_NOT_MET if from different participants)

```suggestion
    FASTDDS_EXPORTED_API ReturnCode_t set_related_datareader(
            const DataReader* related_reader);
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1517,6 +1517,28 @@ ReturnCode_t DataWriterImpl::set_sample_prefilter(
...
 
+ReturnCode_t DataWriterImpl::set_related_datareader_key(
+        const rtps::GUID_t& related_reader_guid)
+{
+    ReturnCode_t ret = RETCODE_ERROR;
+
+    if (nullptr == writer_)
+    {
+        if (related_reader_guid != c_Guid_Unknown &&
+                related_reader_guid.entityId.is_reader())
```
MiguelCompany: Check it has our same prefix.

src/cpp/fastdds/rpc/ReplierImpl.cpp
MiguelCompany: I think we just don't need this check. Since the topic is volatile, we will take whatever is in the reader's history.

src/cpp/fastdds/rpc/ReplierImpl.cpp
```diff
@@ -157,6 +162,13 @@ ReturnCode_t ReplierImpl::take_request(
         return RETCODE_PRECONDITION_NOT_MET;
     }
 
+    // For taking the request is enough to check if the replier is fully matched
+    if (!is_fully_matched())
+    {
+        EPROSIMA_LOG_WARNING(REPLIER, "Trying to take a request with an unmatched requester");
+        return RETCODE_PRECONDITION_NOT_MET;
+    }
+
```
MiguelCompany: Same here:

```suggestion
```

src/cpp/fastdds/rpc/RequesterImpl.cpp
```diff
@@ -96,36 +97,49 @@ ReturnCode_t RequesterImpl::send_request(
             info.related_sample_identity.sequence_number() = wparams.related_sample_identity().sequence_number();
         }
     }
+
+    std::cout << "Request sent with identity: " << info.sample_identity << std::endl;
+
```
MiguelCompany: Debug print?

src/cpp/fastdds/rpc/ReplierImpl.cpp
```diff
@@ -302,22 +326,30 @@ ReplierImpl::RequesterMatchStatus ReplierImpl::requester_match_status(
...
+        if (related_guid.entityId.is_reader())
+        {
+            SubscriptionBuiltinTopicData sub_data;
+            reply_topic_matched = RETCODE_OK == replier_writer_->get_matched_subscription_data(sub_data, related_guid);
+        }
+        else
+        {
+            // At least, demand that the number of matched entities on the replier
+            // writer is equal to the ones on the requester reader.
+            reply_topic_matched = is_fully_matched();
```
MiguelCompany: This should use the `related_datareader_key`. So I would change line 331 to `if (related_guid.entityId.is_reader() && (info.sample_identity.writer_guid() != related_guid))` and always fallback to the code below in the `else`

src/cpp/fastdds/rpc/ReplierImpl.cpp
```diff
@@ -340,6 +372,52 @@ bool ReplierImpl::is_fully_matched() const
...
+        const RequestInfo& info) const
+{
+    WaitSet waitset;
+
+    // Add condition for changes in the replier reader matched status
+    StatusCondition& reader_condition = replier_reader_->get_statuscondition();
+    // Do not overwrite the original mask, since it is used by the server
+    StatusMask new_mask_reader, original_mask_reader = new_mask_reader = reader_condition.get_enabled_statuses();
+    new_mask_reader |= StatusMask::subscription_matched();
+    reader_condition.set_enabled_statuses(new_mask_reader);
```
MiguelCompany: There could be a race here if this method is called from different threads. It might be a good idea if the ReplierImpl becomes a listener of both endpoints, and just notifies on a CV when the matched status of any of them is updated.

src/cpp/fastdds/rpc/RequesterImpl.cpp
```diff
@@ -96,36 +97,49 @@ ReturnCode_t RequesterImpl::send_request(
...
         return RETCODE_PRECONDITION_NOT_MET;
     }
 
+    // For taking the reply is enough to check if the requester is fully matched
+    if (!is_fully_matched())
+    {
+        EPROSIMA_LOG_WARNING(REQUESTER, "Trying to take a reply with an unmatched requester");
+        return RETCODE_PRECONDITION_NOT_MET;
+    }
+
```
MiguelCompany: Not sure about this. Perhaps we could just check if we are matched with at least one reply writer.

MiguelCompany (changes requested):

src/cpp/fastdds/publisher/DataWriterImpl.cpp
Mario-DL: Since we are now passing the `DataReader` directly, I am removing the check for `is_reader()`

src/cpp/fastdds/rpc/RequesterImpl.cpp
Mario-DL: Wops, yes

src/cpp/fastdds/rpc/ReplierImpl.cpp
Mario-DL: Since the `is_fully_matched()` method will remain unusued, I will remove it from the `ReplierImpl`

src/cpp/fastdds/rpc/ReplierImpl.cpp
Mario-DL: Alright. I will apply the same for the `RequesterImpl`

include/fastdds/dds/publisher/DataWriter.hpp
Mario-DL: Unittest also included for this case

src/cpp/fastdds/rpc/ReplierImpl.cpp
Mario-DL: A detail to bear in mind here. 
For any application (for instance the `ReqRep` in the BB tests) directly using the `Replier` and the `Requester` (which is discouraged, only the generated code should do that). If the application attaches through a `WaitSet` to any status conditions and after gets its status (i.e `get_subscription_matched_status` `get_publication_matched_status`), the `current_count_change`  and `total_count_change` values are going to be shadowed by the internal listeners since those are called first and the `current_count_change` and `total_count_change` are immediately resetted after retrieving the statuses the first time for the listeners.

src/cpp/fastdds/rpc/RequesterImpl.cpp
Mario-DL: I added a `replier_match_status()` mimicking the one in the `ReplierImpl` so that we have a method to know it they are partially, fully or unmatched. Also, I removed the `is_fully_matched()` method from the `RequesterImpl` because it remains unused.

include/fastdds/dds/publisher/DataWriter.hpp
```diff
@@ -620,15 +621,16 @@ class DataWriter : public DomainEntity
      *
      * @warning This operation is only valid if the entity is not enabled.
      *
-     * @param [in] related_reader_guid The GUID of the DataReader to set as related.
+     * @param [in] related_reader The GUID of the DataReader to set as related.
```
MiguelCompany:
```suggestion
     * @param [in] related_reader Pointer to the DataReader to set as related.
```

include/fastdds/dds/subscriber/DataReader.hpp
```diff
@@ -1095,15 +1096,16 @@ class DataReader : public DomainEntity
      *
      * @warning This operation is only valid if the entity is not enabled.
      *
-     * @param [in] related_writer_guid The GUID of the DataWriter to set as related.
+     * @param [in] related_writer The DataWriter to set as related.
```
MiguelCompany:
```suggestion
     * @param [in] related_writer Pointer to the DataWriter to set as related.
```

src/cpp/fastdds/rpc/ReplierImpl.cpp
```diff
@@ -261,7 +264,7 @@ ReturnCode_t ReplierImpl::create_dds_entities(
 
     // Create writer for the Reply topic
     replier_writer_ =
-            service_->get_publisher()->create_datawriter(service_->get_reply_topic(), qos.writer_qos, nullptr);
+            service_->get_publisher()->create_datawriter(service_->get_reply_topic(), qos.writer_qos, this);
```
MiguelCompany: Specify the relevant status in the status mask.

MiguelCompany: Applies elsewhere

src/cpp/fastdds/rpc/ReplierImpl.cpp
```diff
@@ -357,53 +351,29 @@ ReplierImpl::RequesterMatchStatus ReplierImpl::requester_match_status(
...
+                [this]()
+                {
+                    return matched_status_changed_.load();
+                });
+
+        // Reset the matched status changed flag
+        matched_status_changed_.store(false);
 
         // Check if the replier is fully matched
         requester_status = requester_match_status(info);
```
MiguelCompany: This should only be done if the wait call succeeds

MiguelCompany (changes requested):

src/cpp/fastdds/rpc/ReplierImpl.cpp
Mario-DL: Well, that makes the `requester_match_status()` to be invoked more frequently after the first time that `matched_status_changed ` becomes `true` (maybe without any news in the status), but Im OK with it

MiguelCompany: I ment both things: setting the flag to false, and updating requeser_status

Mario-DL: Now I understood what you meant.  For the initial comment I interpreted that only when `wait_for_matching` succeeds. But no, you mean checking the output of the `wait_for` and only if succeeeds, reset that flag and update the status. That makes totally sense to me

PR #5921, opened 2025-07-03
Flaky test 2 14 x

EugenioCollado:
## Description

Since update in CI to new Windows version, the test `BlackboxTests_DDS_PIM.DDSStatus.sample_lost_be_dw_be_dr_fragments` is failing consistently (it was flaky before the update). This PR tries to fix this by increasing the time btw consecutive writes.

## Contributor Checklist

## Reviewer Checklist


src/cpp/rtps/transport/asio_helpers.hpp
```diff
@@ -53,7 +53,7 @@ struct asio_helpers
         assert(initial_buffer_value >= minimum_buffer_value);
 
         final_buffer_value = initial_buffer_value;
-        while (final_buffer_value > minimum_buffer_value)
+        while (final_buffer_value >= minimum_buffer_value)
```
MiguelCompany: No need for this, since a final attempt with `minimum_buffer_value` is performed after the while loop is finished.

test/blackbox/api/fastrtps_deprecated/PubSubReader.hpp
```diff
@@ -1102,10 +1102,10 @@ class PubSubReader
     }
 
     PubSubReader& socket_buffer_size(
-            uint32_t sockerBufferSize)
+            uint32_t socketBufferSize)
```
MiguelCompany, line 1105: Same typo exists in dds-pim

test/blackbox/common/DDSBlackboxTestsListeners.cpp
```diff
@@ -776,7 +776,13 @@ void sample_lost_test_dr_init(
         PubSubReader<T>& reader,
         std::function<void(const eprosima::fastdds::dds::SampleLostStatus& status)> functor)
 {
-    reader.sample_lost_status_functor(functor)
+    auto testTransport = std::make_shared<test_UDPv4TransportDescriptor>();
+    testTransport->sendBufferSize = SAMPLE_LOST_TEST_BUFFER_SIZE;
+    testTransport->receiveBufferSize = SAMPLE_LOST_TEST_BUFFER_SIZE;
+
+    reader.disable_builtin_transport()
+            .add_user_transport_to_pparams(testTransport)
```
MiguelCompany: Is this really necessary?
I think `reader.socket_buffer_size(SAMPLE_LOST_TEST_BUFFER_SIZE)` would be enough.

test/blackbox/common/DDSBlackboxTestsListeners.cpp
```diff
@@ -788,9 +794,6 @@ void sample_lost_test_init(
         PubSubWriter<T>& writer,
         std::function<void(const eprosima::fastdds::dds::SampleLostStatus& status)> functor)
 {
-    reader.socket_buffer_size(SAMPLE_LOST_TEST_BUFFER_SIZE);
```
MiguelCompany, line 791: ... so I think we can keep the line for the reader.

The line for the writer does not harm, but I am ok with removing it.

src/cpp/rtps/transport/asio_helpers.hpp
EugenioCollado: But in that final attempt, the condition `(option.value() >= value_to_set) && (option.value() <= max_value)` is checked, which can fail if the socket buffer size actually set is higher than the requested value. For example, in many tests the buffer size is requested as 1024 and the max message size is also 1024. This corner case causes the code to jump directly to that final check, and since `socket.set_option` often sets a value around 2K, the method fails because that value exceeds ´max_value´ (´max_value = static_cast<int32_t>(initial_buffer_value)´).

A better solution might be to remove the upper-limit check against `max_value`, since that logic isn’t applied in the main while loop.

test/blackbox/common/DDSBlackboxTestsListeners.cpp
EugenioCollado: You are right, this is a hot fix done when the builtin transports buffer size was not being correctly set.

test/blackbox/common/DDSBlackboxTestsListeners.cpp
EugenioCollado: I think it’s best to remove that line, as it could lead future readers to believe the entity is using built-in transports.

src/cpp/rtps/transport/asio_helpers.hpp
MiguelCompany: > A better solution might be to remove the upper-limit check against `max_value`, since that logic isn’t applied in the main while loop.

This is better IMO.

test/blackbox/common/DDSBlackboxTestsListeners.cpp
MiguelCompany: Might be also worth moving the line for the reader inside `sample_lost_test_dr_init`

EugenioCollado:
@MiguelCompany all suggestions applied. The CI fail is unrelated as it is about the `test/blackbox/types/Data100kb.cxx` which is not even modified in this PR

src/cpp/rtps/transport/asio_helpers.hpp
```diff
@@ -86,11 +86,9 @@ struct asio_helpers
         socket.set_option(BufferOptionType(value_to_set), ec);
         if (!ec)
         {
-            // Last attempt was successful. Get the actual value set.
```
MiguelCompany, line 89: Please leave the comment

MiguelCompany (changes requested):
Just one NIT. Does not need to run CI again

src/cpp/rtps/transport/asio_helpers.hpp
EugenioCollado: Done in [97a6472](https://github.com/eProsima/Fast-DDS/pull/5921/commits/97a6472bb56ff9eb8895b4a5541c0bb605bd12a2)

MiguelCompany (approved):

MiguelCompany:
CI passed for 759e5f96d1637962069ec4402594dc10ab271684

PR #5988, opened 2025-08-28
[23610] Fix duplicated transport when using specific XML with CLI

cferreiragonz:
## Description

When an XML file with an user transport and a custom metatrafficUnicastLocatorList is passed to the Fast DDS CLI through the `-x` command, two TCP transports will be added to `user_transports` in the end. This PR fixes this behavior by skipping the transport creation if there is already an existing transport created.

@Mergifyio backport 3.3.x 3.2.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/tools/fds/CliDiscoveryManager.cpp
+++ b/tools/fds/CliDiscoveryManager.cpp
@@ -566,6 +566,11 @@ bool CliDiscoveryManager::add_tcp_servers()
 
 void CliDiscoveryManager::configure_transports()
 {
+    if (serverQos.transport().user_transports.size() > 0)
+    {
+        // User transports already configured in XML file
+        return;
+    }
     if (!serverQos.wire_protocol().builtin.metatrafficUnicastLocatorList.has_kind<LOCATOR_KIND_UDPv4>())
     {
         serverQos.transport().use_builtin_transports = false;
--- a/test/system/tools/fds/CMakeLists.txt
+++ b/test/system/tools/fds/CMakeLists.txt
@@ -37,6 +37,7 @@ if(Python3_Interpreter_FOUND)
         test_fast_discovery_several_server_ids
         test_fast_discovery_invalid_locator
         test_fast_discovery_non_existent_profile
+        test_fast_discovery_tcp_via_XML
     )
 
     if(SECURITY)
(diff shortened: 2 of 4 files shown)


tools/fds/CliDiscoveryManager.cpp
```diff
@@ -566,6 +566,11 @@ bool CliDiscoveryManager::add_tcp_servers()
 
 void CliDiscoveryManager::configure_transports()
 {
+    if (serverQos.transport().user_transports.size() > 0)
```
juanlofer-eprosima, line 569: I am wondering if it could be problematic to mix an XML profile and CLI arguments. Perhaps if using XML configuration we should forget about CLI args (move `get_cli_ports_and_ips` under `options[XML_FILE] == nullptr` condition)? Otherwise when defining IP/ports via CLI I believe the ones in the XML would be cleared.

cferreiragonz: Yes, but that is intended behavior. As stated in [documentation](https://fast-dds.docs.eprosima.com/en/latest/fastddscli/cli/cli.html#discovery-server-cli-standard-mode):

> The CLI options override XML configuration for that specific parameter.

The idea is to be able to mix both XML and CLI configurations to, for example, load security profiles from XML files while being able to use the CLI to select ports and IPs.

juanlofer-eprosima: Ok, I see how mixing can be useful. However I believe you could be having locators for which a transport is not created (would that produce an error?), and create transports for kinds while the locators defined in the XML for those kinds are cleared. Maybe we could fix the first situation (if it's indeed troublesome, the second does not seem problematic) by checking if there exists a transport for each locator kind, and create it if there is none. Or we can just move forward and merge :)

cferreiragonz: The CLI automatically creates transports for the specified arguments. This PR modifies this behavior so that if the user specifically define any transport in the XML, the transport creation is skipped. If you define one transport in an XML file and pass non compatible arguments to the CLI (i.e. defining a TCP transport and passing an UDP locator) this will raise an error as the Participant creation will fail. I believe this is user responsibility. The **same error** will be obtained if this configuration is replicated in a single XML file, with a transport of one type and using locators of a different type.

Fast DDS does not automatically fix this miss-configuration when obtained through XML files, and I believe neither we should fix it here.

juanlofer-eprosima:
@Mergifyio backport 3.3.x 3.2.x

MiguelCompany:
@Mergifyio backport 3.3.x 3.2.x

PR #6016, opened 2025-09-08
Hotfix/23289: deadline timer fixes

zesk1999:
Only check DataWriterImpl.cpp and DataWriterImpl.hpp, the rest of the files I changed are examples used for testing.

## Contributor Checklist

- [ ] Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
    
- [ ] Applicable backports have been included in the description.

## Reviewer Checklist

### DIFF
--- a/src/cpp/fastdds/publisher/DataWriterImpl.cpp
+++ b/src/cpp/fastdds/publisher/DataWriterImpl.cpp
@@ -59,7 +59,7 @@
 #ifdef FASTDDS_STATISTICS
 #include <statistics/fastdds/domain/DomainParticipantImpl.hpp>
 #include <statistics/types/monitorservice_types.hpp>
-#endif //FASTDDS_STATISTICS
+#endif // FASTDDS_STATISTICS
 
 using namespace eprosima::fastdds;
 using namespace eprosima::fastdds::rtps;
@@ -431,7 +431,6 @@ ReturnCode_t DataWriterImpl::enable()
     }
 
     writer_ = BaseWriter::downcast(writer);
-
     // Set DataWriterImpl as the implementer of the
     // IReaderDataFilter interface
     writer_->reader_data_filter(this);
@@ -439,12 +438,8 @@ ReturnCode_t DataWriterImpl::enable()
     // In case it has been loaded from the persistence DB, rebuild instances on history
     history_->rebuild_instances();
 
-    deadline_timer_ = new TimedEvent(publisher_->rtps_participant()->get_resource_event(),
-                    [&]() -> bool
-                    {
-                        return deadline_missed();
-                    },
-                    qos_.deadline().period.to_ns() * 1e-6);
+    std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
+    configure_deadline_timer_locked_();
 
     lifespan_timer_ = new TimedEvent(publisher_->rtps_participant()->get_resource_event(),
                     [&]() -> bool
@@ -687,8 +682,8 @@ ReturnCode_t DataWriterImpl::check_write_preconditions(
         type_.get()->compute_key(data, instance_handle, is_key_protected);
     }
 
-    //Check if the Handle is different from the special value HANDLE_NIL and
-    //does not correspond with the instance referred by the data
+    // Check if the Handle is different from the special value HANDLE_NIL and
+    // does not correspond with the instance referred by the data
     if (handle.isDefined() && handle != instance_handle)
     {
         return RETCODE_PRECONDITION_NOT_MET;
@@ -1040,6 +1035,7 @@ ReturnCode_t DataWriterImpl::perform_create_new_change(
         }
     }
 
+    // create_change seeds the next per-instance deadline and reschedules the timer for the next sample.
     CacheChange_t* ch = history_->create_change(change_kind, handle);
     if (ch != nullptr)
     {
@@ -1072,7 +1068,7 @@ ReturnCode_t DataWriterImpl::perform_create_new_change(
             return RETCODE_TIMEOUT;
         }
 
-        if (qos_.deadline().period != dds::c_TimeInfinite)
+        if (deadline_timer_ != nullptr && qos_.deadline().period.to_ns() > 0 && qos_.deadline().period != dds::c_TimeInfinite)
         {
             if (!history_->set_next_deadline(
                         handle,
@@ -1183,7 +1179,7 @@ void DataWriterImpl::publisher_qos_updated()
 {
     if (writer_ != nullptr)
     {
-        //NOTIFY THE BUILTIN PROTOCOLS THAT THE WRITER HAS CHANGED
+        // NOTIFY THE BUILTIN PROTOCOLS THAT THE WRITER HAS CHANGED
         WriterQos wqos = qos_.get_writerqos(get_publisher()->get_qos(), topic_->get_qos());
         publisher_->rtps_participant()->update_writer(writer_, wqos);
     }
@@ -1218,6 +1214,8 @@ ReturnCode_t DataWriterImpl::set_qos(
         return RETCODE_IMMUTABLE_POLICY;
     }
 
+    const dds::Duration_t prev_deadline = qos_.deadline().period;
+
     set_qos(qos_, qos_to_set, !enabled);
 
     if (enabled)
@@ -1237,20 +1235,17 @@ ReturnCode_t DataWriterImpl::set_qos(
             writer_->update_attributes(w_att);
         }
 
-        //Notify the participant that a Writer has changed its QOS
+        // Notify the participant that a Writer has changed its QOS
         WriterQos wqos = qos_.get_writerqos(get_publisher()->get_qos(), topic_->get_qos());
         publisher_->rtps_participant()->update_writer(writer_, wqos);
 
-        // Deadline
-        if (qos_.deadline().period != dds::c_TimeInfinite)
+
+        std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
+
+        // If the deadline period actually changed, (re)configure the timer.
+        if (qos_.deadline().period != prev_deadline)
         {
-            deadline_duration_us_ =
-                    duration<double, std::ratio<1, 1000000>>(qos_.deadline().period.to_ns() * 1e-3);
-            deadline_timer_->update_interval_millisec(qos_.deadline().period.to_ns() * 1e-6);
-        }
-        else
-        {
-            deadline_timer_->cancel_timer();
+            configure_deadline_timer_locked_();
         }
 
         // Lifespan
@@ -1334,7 +1329,7 @@ void DataWriterImpl::InnerDataWriterListener::on_offered_incompatible_qos(
 
 #ifdef FASTDDS_STATISTICS
     notify_status_observer(statistics::StatusKind::INCOMPATIBLE_QOS);
-#endif //FASTDDS_STATISTICS
+#endif // FASTDDS_STATISTICS
 
     data_writer_->user_datawriter_->get_statuscondition().get_impl()->set_status(notify_status, true);
 }
@@ -1373,7 +1368,7 @@ void DataWriterImpl::InnerDataWriterListener::on_liveliness_lost(
 
 #ifdef FASTDDS_STATISTICS
     notify_status_observer(statistics::StatusKind::LIVELINESS_LOST);
-#endif //FASTDDS_STATISTICS
+#endif // FASTDDS_STATISTICS
 
     data_writer_->user_datawriter_->get_statuscondition().get_impl()->set_status(notify_status, true);
 }
@@ -1417,7 +1412,7 @@ void DataWriterImpl::InnerDataWriterListener::notify_status_observer(
     }
 }
 
-#endif //FASTDDS_STATISTICS
+#endif // FASTDDS_STATISTICS
 
 ReturnCode_t DataWriterImpl::wait_for_acknowledgments(
         const dds::Duration_t& max_wait)
@@ -1551,9 +1546,14 @@ ReturnCode_t DataWriterImpl::set_related_datareader(
     return ret;
 }
 
+// deadline_timer_reschedule returns true if it could compute and set a new interval; false if there’s no pending deadline.
 bool DataWriterImpl::deadline_timer_reschedule()
 {
     assert(qos_.deadline().period != dds::c_TimeInfinite);
+    if (deadline_timer_ == nullptr)
+    {
+        return false;
+    }
 
     std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
 
@@ -1569,11 +1569,102 @@ bool DataWriterImpl::deadline_timer_reschedule()
     return true;
 }
 
+void DataWriterImpl::configure_deadline_timer_locked_()
+{
+
+    // Tear down any existing timer
+    if (deadline_missed_status_.total_count >= std::numeric_limits<int32_t>::max())
+    {
+        if (deadline_timer_)              // if it exists, shut it down
+        {
+            deadline_timer_->cancel_timer();
+            delete deadline_timer_;
+            deadline_timer_ = nullptr;
+        }
+        return;                            // do not create a new timer
+    }
+
+    // Tear down any existing timer (normal reconfigure path)
+    if (deadline_timer_ != nullptr)
+    {
+        deadline_timer_->cancel_timer();
+        delete deadline_timer_;
+        deadline_timer_ = nullptr;
+    }
+
+    // Refresh cached duration and guards
+    deadline_duration_us_ =
+            std::chrono::duration<double, std::ratio<1, 1000000>>(qos_.deadline().period.to_ns() * 1e-3);
+
+    // Create the timer with a safe callback
+    deadline_timer_ = new TimedEvent(
+        publisher_->rtps_participant()->get_resource_event(),
+        [this]() -> bool
+        {
+            std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
+
+            // If QoS is 0 or infinite, warn once, self-disable, and stop.
+            if ((qos_.deadline().period.to_ns() == 0) || (qos_.deadline().period == dds::c_TimeInfinite))
+            {
+                
+                EPROSIMA_LOG_WARNING(
+                    DATA_WRITER,
+                    "Deadline period is "
+                        << (qos_.deadline().period.to_ns() == 0 ? "0" : "infinite")
+                        << ", it will be ignored from now on. Timer is going to be canceled.");   
+
+                if (deadline_timer_ != nullptr)
+                {
+                    deadline_timer_->cancel_timer();
+                }
+                return false;
+            }
+
+            // Normal path
+            return deadline_missed();
+        },
+        0.0);
+
+    // Initial arming logic
+
+    if ((qos_.deadline().period.to_ns() > 0) && (qos_.deadline().period != dds::c_TimeInfinite))
+    {
+        // Only arm if a next deadline is already computed; otherwise keep disabled to avoid spurious fire.
+        std::chrono::steady_clock::time_point next_tp;
+        if (history_->get_next_deadline(timer_owner_, next_tp))
+        {
+            auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                          next_tp - std::chrono::steady_clock::now()).count();
+            if (ms < 0) ms = 0;
+
+            deadline_timer_->update_interval_millisec(static_cast<double>(ms));
+            deadline_timer_->restart_timer();
+        }
+        else
+        {
+            deadline_timer_->cancel_timer(); // keep it created but idle
+        }
+    }
+    else
+    {
+        // 0 or infinite -> fire once to log and self-disable
+        deadline_timer_->restart_timer();
+    }
+}
+
+
 bool DataWriterImpl::deadline_missed()
 {
     assert(qos_.deadline().period != dds::c_TimeInfinite);
 
     std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
+    
+    if (deadline_missed_status_.total_count >= std::numeric_limits<int32_t>::max() && deadline_timer_ != nullptr)
+    {
+        EPROSIMA_LOG_WARNING(DATA_WRITER, "Maximum number of deadline missed messages reached. Stopping deadline timer.");
+        deadline_timer_->cancel_timer();
+        return false; 
+    }
 
     deadline_missed_status_.total_count++;
     deadline_missed_status_.total_count_change++;
@@ -1588,7 +1679,7 @@ bool DataWriterImpl::deadline_missed()
 
 #ifdef FASTDDS_STATISTICS
     writer_listener_.notify_status_observer(statistics::StatusKind::DEADLINE_MISSED);
-#endif //FASTDDS_STATISTICS
+#endif // FASTDDS_STATISTICS
 
     user_datawriter_->get_statuscondition().get_impl()->set_status(notify_status, true);
 
--- a/src/cpp/fastdds/publisher/DataWriterImpl.hpp
+++ b/src/cpp/fastdds/publisher/DataWriterImpl.hpp
@@ -771,6 +771,8 @@ protected:
 
 private:
 
+    void configure_deadline_timer_locked_();
+
     void create_history(
             const std::shared_ptr<IPayloadPool>& payload_pool,
             const std::shared_ptr<IChangePool>& change_pool);
--- a/examples/cpp/configuration/CLIParser.hpp
+++ b/examples/cpp/configuration/CLIParser.hpp
@@ -647,18 +647,17 @@ public:
                 {
                     try
                     {
-                        int input = std::stoi(argv[i]);
-                        if (input < 1 || static_cast<long>(input) > static_cast<long>(max_duration))
+                        // Always interpret as milliseconds (decimals allowed)
+                        double ms = std::stod(argv[i]);
+                        if (ms < 0.0 || ms > static_cast<double>(max_duration))
                         {
-                            throw std::out_of_range("deadline argument " + std::string(
-                                              argv[i]) + " out of range [1, " + std::to_string(
-                                              max_duration) + "].");
+                            throw std::out_of_range(
+                                "deadline argument " + std::string(argv[i]) +
+                                " out of range [0, " + std::to_string(max_duration) + "] ms.");
                         }
-                        else
-                        {
-                            config.pub_config.deadline = static_cast<uint32_t>(input);
-                            config.sub_config.deadline = static_cast<uint32_t>(input);
-                        }
+                        uint32_t deadline_ms = static_cast<uint32_t>(std::llround(ms));
+                        config.pub_config.deadline = deadline_ms;
+                        config.sub_config.deadline = deadline_ms;
                     }
                     catch (const std::invalid_argument& e)
                     {
@@ -678,6 +677,7 @@ public:
                     print_help(EXIT_FAILURE);
                 }
             }
+
             else if (arg == "--disable-positive-ack")
             {
                 config.pub_config.disable_positive_ack = true;
--- a/examples/cpp/configuration/PublisherApp.cpp
+++ b/examples/cpp/configuration/PublisherApp.cpp
@@ -125,6 +125,18 @@ PublisherApp::PublisherApp(
     {
         DataWriterQos writer_qos = DATAWRITER_QOS_DEFAULT;
         publisher_->get_default_datawriter_qos(writer_qos);
+        // Helper: milliseconds (double) -> Duration_t (sec + nsec), rounded to nearest ns
+        auto ms_to_duration = [](double ms) -> eprosima::fastdds::dds::Duration_t
+        {
+            if (ms == 0.0) {
+                return eprosima::fastdds::dds::Duration_t{0, 0}; // preserve zero => triggers warning & disables timer
+            }
+            long long ns  = static_cast<long long>(std::llround(ms * 1e6)); // 1 ms = 1e6 ns
+            int32_t  sec  = static_cast<int32_t>(ns / 1000000000LL);
+            uint32_t nsec = static_cast<uint32_t>(ns - static_cast<long long>(sec) * 1000000000LL);
+            return eprosima::fastdds::dds::Duration_t{sec, nsec};
+        };
+
         writer_qos.publish_mode().kind = config.publish_mode;
         writer_qos.reliability().kind = config.reliability;
         writer_qos.durability().kind = config.durability;
@@ -141,10 +153,8 @@ PublisherApp::PublisherApp(
                       "DataWriter initialization failed: ownership strength is only valid with exclusive ownership");
         }
         writer_qos.ownership_strength().value = config.ownership_strength;
-        if (config.deadline > 0)
-        {
-            writer_qos.deadline().period = eprosima::fastdds::dds::Duration_t(config.deadline * 1e-3);
-        }
+        writer_qos.deadline().period = ms_to_duration(static_cast<double>(config.deadline));
+        
         writer_qos.reliable_writer_qos().disable_positive_acks.enabled = config.disable_positive_ack;
         if (config.ack_keep_duration > 0)
         {
@@ -175,6 +185,22 @@ PublisherApp::PublisherApp(
     {
         throw std::runtime_error("DataWriter initialization failed");
     }
+    // right after you confirm writer_ was created successfully
+    eprosima::fastdds::dds::DataWriter* w = writer_;
+    std::thread([w]() {
+        std::this_thread::sleep_for(std::chrono::seconds(3));
+
+        eprosima::fastdds::dds::DataWriterQos q;
+        w->get_qos(q);
+
+        // set deadline to 200 ms
+        q.deadline().period = eprosima::fastdds::dds::Duration_t{0, 200000000u}; // 0s + 200,000,000 ns
+
+        auto rc = w->set_qos(q);
+        (void)rc; // silence -Werror=unused-variable
+    }).detach();
+
+
 }
 
 PublisherApp::~PublisherApp()
--- a/examples/cpp/configuration/main.cpp
+++ b/examples/cpp/configuration/main.cpp
@@ -42,6 +42,8 @@ int main(
         int argc,
         char** argv)
 {
+    Log::SetVerbosity(Log::Warning);
+    Log::SetVerbosity(Log::Kind::Info);
     auto ret = EXIT_SUCCESS;
     CLIParser::configuration_config config = CLIParser::parse_cli_options(argc, argv);
     uint16_t samples = 0;


zesk1999:
Only check DataWriterImpl.cpp and DataWriterImpl.hpp, the rest of the files I changed are examples used for testing.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1040,6 +1035,7 @@ ReturnCode_t DataWriterImpl::perform_create_new_change(
         }
     }
 
+    // create_change seeds the next per-instance deadline and reschedules the timer for the next sample.
```
juanlofer-eprosima:
```suggestion
    // create_change seeds the next per-instance deadline and reschedules the timer for the next sample
```

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -24,6 +26,50 @@
 
 using namespace eprosima::fastdds;
 using namespace eprosima::fastdds::rtps;
+using eprosima::fastdds::dds::Duration_t;
+using eprosima::fastdds::dds::RELIABLE_RELIABILITY_QOS;
+using eprosima::fastdds::dds::KEEP_LAST_HISTORY_QOS;
+using eprosima::fastdds::dds::RETCODE_OK;
+using eprosima::fastdds::dds::StatusMask;
```
juanlofer-eprosima: I believe these lines are not required.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -24,6 +26,50 @@
...
+  #include <fastdds/dds/log/Log.hpp>
+  #include <fastdds/dds/log/OStreamConsumer.hpp>
+  namespace fastlog = eprosima::fastdds::dds;
+#elif __has_include(<fastrtps/log/Log.h>)
+  #include <fastrtps/log/Log.h>
+  #include <fastrtps/log/OStreamConsumer.h>
+  namespace fastlog = eprosima::fastrtps;
+#else
+  #error "No Fast DDS/RTPS log header found"
+#endif
```
juanlofer-eprosima: Just include the newer version.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -24,6 +26,50 @@
...
+
+// Consumer: counts matching warnings and mirrors them to an ostringstream
```
juanlofer-eprosima: But the ostringstream is not used right?

juanlofer-eprosima: I would consider moving this implementation to the test where it is used. Alternatively, you could make the class more generic (having a counter per category, storing logs if configured, performing the "needle" check with API, etc.) and add it to test/utils, it could be useful for future tests.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +355,67 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
     EXPECT_GE(writer.missed_deadlines(), 1u);
 }
 
+/**
+ * Regression test for the zero-deadline period bug.
```
juanlofer-eprosima, line 318: Add reference to Redmine issue.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +355,67 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+    writer.wait_discovery();
+    reader.wait_discovery();
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(150));
+
+    // To directly read the status struct
+    eprosima::fastdds::dds::OfferedDeadlineMissedStatus st{};
+    ASSERT_EQ(writer.get_offered_deadline_missed_status(st), eprosima::fastdds::dds::RETCODE_OK);
+
+    EXPECT_EQ(st.total_count, std::numeric_limits<int32_t>::max()) << "Expected the max value after a zero-deadline warning.";
```
juanlofer-eprosima:
```suggestion
    EXPECT_EQ(st.total_count, std::numeric_limits<uint32_t>::max()) << "Expected the max value after a zero-deadline warning.";
```

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +355,67 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
     EXPECT_GE(writer.missed_deadlines(), 1u);
 }
 
+/**
+ * Regression test for the zero-deadline period bug.
+ * Creating a DataWriter with a deadline of 0.
+ * Checking if a warning is logged exactly once, the timer is cancelled without missed deadline
+ * messages and a total count set to max integer.
+ */
+TEST_P(DeadlineQos, DeadlineZeroWithoutDeadlineMissedMessages)
```
juanlofer-eprosima:
```suggestion
TEST_P(DeadlineQos, ZeroDeadlinePeriod)
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -431,20 +431,15 @@ ReturnCode_t DataWriterImpl::enable()
     }
 
     writer_ = BaseWriter::downcast(writer);
-
```
juanlofer-eprosima, line 434: NIT: personally I believe it's better to leave the original author's format style unless it seems unintentional or creates confusion.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
     return true;
 }
 
+/**
+ * This function tears down any existing deadline timer and creates a new one
+ * configured for the current deadline period. If the period is 0 or infinite,
+ * the timer is created to fire only once to log a warning and then cancel itself.
+ */
```
juanlofer-eprosima: Please move this to the header file.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+void DataWriterImpl::configure_deadline_timer_locked_()
```
juanlofer-eprosima: We usually use "_nts" instead. Anyway, I believe you can make this method thread safe since the mutex is recursive.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1218,6 +1214,8 @@ ReturnCode_t DataWriterImpl::set_qos(
         return RETCODE_IMMUTABLE_POLICY;
     }
 
+    const dds::Duration_t prev_deadline = qos_.deadline().period;
```
juanlofer-eprosima: NIT: I would store the whole previous qos struct, you could also use if below in the lifespan section.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1237,20 +1235,22 @@ ReturnCode_t DataWriterImpl::set_qos(
...
-        // Deadline
-        if (qos_.deadline().period != dds::c_TimeInfinite)
-        {
-            deadline_duration_us_ =
-                    duration<double, std::ratio<1, 1000000>>(qos_.deadline().period.to_ns() * 1e-3);
-            deadline_timer_->update_interval_millisec(qos_.deadline().period.to_ns() * 1e-6);
-        }
-        else
+
+        std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
```
juanlofer-eprosima: I believe the mutex should be taken at the beginning of the method not here. Probably `enable` should also be protected.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+ * configured for the current deadline period. If the period is 0 or infinite,
+ * the timer is created to fire only once to log a warning and then cancel itself.
+ */
+void DataWriterImpl::configure_deadline_timer_locked_()
+{
+    if (deadline_timer_ != nullptr)
+    {
+        deadline_timer_->cancel_timer();
+        delete deadline_timer_;
+        deadline_timer_ = nullptr;
```
juanlofer-eprosima: Wouldn't it be better to update the object if already exists?

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+    }
+
+    // If we had previously hit the max, keep timer disabled until QoS changes reset the count.
+    if (deadline_missed_status_.total_count >= std::numeric_limits<int32_t>::max())
+    {
+        return; // no new timer
+    }
```
juanlofer-eprosima: I believe this condition will never be given. If that's the case, I would this block to make the code more clear.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: And maybe it is not that useful to destroy the object, leaving the timer cancelled might be enough and simplify the solution.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+            std::chrono::duration<double, std::ratio<1, 1000000>>(qos_.deadline().period.to_ns() * 1e-3);
+
+    // Create the timer with a safe callback
+    deadline_timer_ = new TimedEvent(
+        publisher_->rtps_participant()->get_resource_event(),
+        [this]() -> bool
+        {
+            std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
+
+            if ((qos_.deadline().period.to_ns() == 0) || (qos_.deadline().period == dds::c_TimeInfinite))
```
juanlofer-eprosima: I don't think the count should be set to max for the infinite case.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+            {
+                
+                EPROSIMA_LOG_WARNING(
+                    DATA_WRITER,
+                    "Deadline period is "
+                        << (qos_.deadline().period.to_ns() == 0 ? "0" : "infinite")
+                        << ", it will be ignored from now on. Timer is going to be canceled."); 
+                deadline_missed_status_.total_count = std::numeric_limits<int32_t>::max();
```
juanlofer-eprosima: I would move this to outside the callback.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+
+                if (deadline_timer_ != nullptr)
+                {
+                    deadline_timer_->cancel_timer();
+                }
+                return false;
```
juanlofer-eprosima: So when deadline is set to 0, the listener callback is never triggered right? I thought we wanted it to be triggered once. Hence, I would move this to `deadline_missed` method.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+            }
+
+            // Normal path
+            return deadline_missed();
+        },
+        0.0);
```
juanlofer-eprosima: Why don't you create the timer with the value you already computed? Even if you modify the period just below, the timer can be triggered nevertheless.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+    {
+        // Only arm if a next deadline is already computed; otherwise keep disabled to avoid spurious fire.
+        std::chrono::steady_clock::time_point next_tp;
+        if (history_->get_next_deadline(timer_owner_, next_tp))
+        {
+            auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(
+                          next_tp - std::chrono::steady_clock::now()).count();
+            if (ms < 0) ms = 0;
+
+            deadline_timer_->update_interval_millisec(static_cast<double>(ms));
```
juanlofer-eprosima: Can't you use `deadline_timer_reschedule`?

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+}
+
+
 bool DataWriterImpl::deadline_missed()
 {
     assert(qos_.deadline().period != dds::c_TimeInfinite);
 
     std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
 
+    if (deadline_missed_status_.total_count >= std::numeric_limits<int32_t>::max() && deadline_timer_ != nullptr)
```
juanlofer-eprosima: I believe the second condition can never occur.

juanlofer-eprosima: And regarding the first one, I would say the idea is that it's also never given, the timer should have been cancelled.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,15 +1574,109 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+        deadline_timer_->cancel_timer();
+        return false; 
+    }
+
     deadline_missed_status_.total_count++;
     deadline_missed_status_.total_count_change++;
     deadline_missed_status_.last_instance_handle = timer_owner_;
+
+    // If we just reached the max -> log ONCE, stop timer, and bail.
+    if (deadline_missed_status_.total_count == std::numeric_limits<int32_t>::max())
```
juanlofer-eprosima:
```suggestion
    if (deadline_missed_status_.total_count == std::numeric_limits<uint32_t>::max())
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
zesk1999: I believe the case when the deadline period is infinite could cause issues. That's why I left the separate cases for infinite and any other number.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: Please remove spaces.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include <thread>
```
juanlofer-eprosima, line 15: Do not remove header, we follow the "include what you use policy", and there are a couple of sleeps in this cpp.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: After more careful review, I believe a new mutex protecting the `qos_` attribute would be required. However, I am not optimistic that it won't produce ABBAs.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -431,20 +431,16 @@ ReturnCode_t DataWriterImpl::enable()
...
     // In case it has been loaded from the persistence DB, rebuild instances on history
     history_->rebuild_instances();
 
-    deadline_timer_ = new TimedEvent(publisher_->rtps_participant()->get_resource_event(),
-                    [&]() -> bool
-                    {
-                        return deadline_missed();
-                    },
-                    qos_.deadline().period.to_ns() * 1e-6);
+    std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
```
juanlofer-eprosima:
```suggestion
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1072,7 +1069,7 @@ ReturnCode_t DataWriterImpl::perform_create_new_change(
             return RETCODE_TIMEOUT;
         }
 
-        if (qos_.deadline().period != dds::c_TimeInfinite)
+        if (deadline_timer_ != nullptr && qos_.deadline().period.to_ns() > 0 && qos_.deadline().period != dds::c_TimeInfinite && deadline_missed_status_.total_count < std::numeric_limits<uint32_t>::max())
```
juanlofer-eprosima: I would remove the first condition if eventually the timer gets always constructed regardless of the deadline period value.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1218,10 +1215,16 @@ ReturnCode_t DataWriterImpl::set_qos(
...
     }
 
+    // Take a snapshot of the current QoS before mutating it
+    const DataWriterQos old_qos = qos_;
+
     set_qos(qos_, qos_to_set, !enabled);
 
     if (enabled)
     {
+        std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
```
juanlofer-eprosima: As already stated, I believe this is still not thread safe; the three instructions above should also be protected, and since it's not always possible to rely on the writer's mutex, I believe a new one for protecting the qos attribute is needed.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1551,9 +1555,14 @@ ReturnCode_t DataWriterImpl::set_related_datareader(
     return ret;
 }
 
+// deadline_timer_reschedule returns true if it could compute and set a new interval; false if there’s no pending deadline.
```
juanlofer-eprosima: This should go into the header.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1551,9 +1555,14 @@ ReturnCode_t DataWriterImpl::set_related_datareader(
...
 bool DataWriterImpl::deadline_timer_reschedule()
 {
     assert(qos_.deadline().period != dds::c_TimeInfinite);
+    if (deadline_timer_ == nullptr || deadline_missed_status_.total_count >= std::numeric_limits<uint32_t>::max())
+    {
+        return false;
+    }
```
juanlofer-eprosima: You don't need to check these conditions I believe, but should be done before calling the method. In any case, you could add the conditions to the assert.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
         return false;
     }
 
+    if (next_deadline_us <= std::chrono::steady_clock::now())
+    {
+        // Ignore uninitialized or stale deadlines; don't arm to avoid a spurious first miss
+        return false;
+    }
```
juanlofer-eprosima: TODO: double check before merging whether this is actually needed or the condition was given due to a flaw already solved.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+    std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
+
+    // Create the timer once 
+    if (deadline_timer_ == nullptr)
+    {
+        deadline_timer_ = new TimedEvent(
+            publisher_->rtps_participant()->get_resource_event(),
+            [this]() -> bool
+            {
+                std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
```
juanlofer-eprosima:
```suggestion
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+
+                // Normal path: process a deadline miss. This function will
+                // update counters, notify, and ask for reschedule as needed.
```
juanlofer-eprosima:
```suggestion
```

juanlofer-eprosima: It's better to document what the method does from the method itself, since its behavior could be changed and comments like this remain stale.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+                return deadline_missed();
+            },
+            (qos_.deadline().period == dds::c_TimeInfinite) ? std::numeric_limits<double>::max() : qos_.deadline().period.to_ns() * 1e-6
+            // In case of deadline period = infinite, multiplying it by 1e-6 could cause a problem
```
juanlofer-eprosima: Perhaps you could create the timer always with maximum value, and then update it with the actual value if it's a valid one.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+    }
+
+    // Handle "infinite" and "zero" outside the callback
+    if (qos_.deadline().period == dds::c_TimeInfinite)
+    {
+        deadline_timer_->cancel_timer();
+        return;
+    }
+
+    deadline_duration_us_ = std::chrono::duration<double, std::ratio<1, 1000000>>(qos_.deadline().period.to_ns() * 1e-3);
```
juanlofer-eprosima: Maybe it's better to keep the maximum value when infinite, as it was the case before.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+        deadline_timer_->restart_timer();
+    }
+    else
+    {
+        // Keep the timer object around but idle if there's no pending deadline
+        deadline_timer_->cancel_timer();
+    }
+}
+
+void DataWriterImpl::notify_deadline_missed_no_increment_()
```
juanlofer-eprosima:
```suggestion
void DataWriterImpl::notify_deadline_missed_()
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+{
+    std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
```
juanlofer-eprosima: In this case I believe it makes more sense to make the method nts, as it's clearly not an standalone method.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+    if (qos_.deadline().period.to_ns() == 0)
+    {
+        EPROSIMA_LOG_WARNING(
+            DATA_WRITER,
+            "Deadline period is 0, it will be ignored from now on. Timer is going to be canceled.");
+
+        // Bump once and notify listener exactly once.
+        notify_deadline_missed_no_increment_();
+
+        deadline_missed_status_.total_count = std::numeric_limits<uint32_t>::max();
```
juanlofer-eprosima: Shouldn't this be done before notifying?

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: And also, when the period is 0, I believe there would exist a race condition in which the timer could be triggered before getting canceled.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: And I also think we should set the `total_count_change` to be the maximum.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+        deadline_timer_->cancel_timer();
+        return;
+    }
+
+    if (deadline_timer_reschedule())
+    {
+        deadline_timer_->restart_timer();
```
juanlofer-eprosima: Cancel before as well? That's the way it was being done in other sections of the code.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1578,19 +1661,20 @@ bool DataWriterImpl::deadline_missed()
...
+        deadline_timer_->cancel_timer();
+        return false; // do not reschedule
     }
 
 #ifdef FASTDDS_STATISTICS
     writer_listener_.notify_status_observer(statistics::StatusKind::DEADLINE_MISSED);
-#endif //FASTDDS_STATISTICS
-
-    user_datawriter_->get_statuscondition().get_impl()->set_status(notify_status, true);
+#endif // FASTDDS_STATISTICS
```
juanlofer-eprosima: I believe this should go into the notification method.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -2448,4 +2532,4 @@ bool DataWriterImpl::is_relevant(
 
 } // namespace dds
 } // namespace fastdds
-} // namespace eprosima
+} // namespace eprosima
```
juanlofer-eprosima: We always have an EOL on purpose.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1564,11 +1573,85 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+
+        // Bump once and notify listener exactly once.
+        notify_deadline_missed_no_increment_();
+
+        deadline_missed_status_.total_count = std::numeric_limits<uint32_t>::max();
+        deadline_timer_->cancel_timer();
+        return;
+    }
+
+    if (deadline_timer_reschedule())
```
juanlofer-eprosima: I believe there is a problem here, since this method gets the next deadline period from the history, but it may not be updated yet.

test/blackbox/api/dds-pim/PubSubWriter.hpp
```diff
@@ -287,6 +287,12 @@ class PubSubWriter
     typedef TypeSupport type_support;
     typedef typename type_support::type type;
 
+    eprosima::fastdds::dds::ReturnCode_t get_offered_deadline_missed_status(
+            eprosima::fastdds::dds::OfferedDeadlineMissedStatus& st)
+    {
+        return datawriter_->get_offered_deadline_missed_status(st);
+    }
```
juanlofer-eprosima: I don't think you need this method, since the listener is already storing the value. You could modify the listener to store the whole object, or add each of the attributes you need instead. It would be interesting to also test for the value of `total_count_change`.

juanlofer-eprosima (changes requested):
Remember to apply the same fixes to the reader before merging.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
juanlofer-eprosima: Please restore blank line. We separate blocks of include statements on purpose, differentiating on system headers and Fast DDS ones (among others).

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: I actually believe it would be good to add the conditions to the assert, just take into account that you would need to move the mutex up if checking the value of `deadline_missed_status_`.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: Could you please a short comment next to `std::numeric_limits<double>::max()`? It may odd for a fellow developer at first sight.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +310,81 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
 }
 
+/**
+ * Fixes Redmine issue #23289.
+ * Regression test for the zero-deadline period bug.
+ * Creating a DataWriter with a deadline of 0.
+ * Checking if a warning is logged exactly once, the timer is cancelled without missed deadline
+ * messages and a total count set to max integer.
+ */
+TEST_P(DeadlineQos, ZeroDeadlinePeriod)
```
juanlofer-eprosima: This test is great for reproducing the bug, however I am wondering if we should also add a test for checking the deadline qos update since this PR is changing the behavior for that situation.

src/cpp/fastdds/publisher/DataWriterImpl.hpp
```diff
@@ -642,6 +642,7 @@ class DataWriterImpl : protected rtps::IReaderDataFilter
     bool deadline_missed();
 
     /**
+     * Returns true if it could compute and set a new interval; false if there’s no pending deadline
```
juanlofer-eprosima: Please follow doxygen format (return has a dedicated tag and it's usually placed after brief).

src/cpp/fastdds/publisher/DataWriterImpl.hpp
```diff
@@ -771,6 +772,18 @@ class DataWriterImpl : protected rtps::IReaderDataFilter
 
 private:
 
+    /**
+    * This function tears down any existing deadline timer and creates a new one
+    * configured for the current deadline period. If the period is 0 or infinite,
+    * the timer is created to fire only once to log a warning and then cancel itself.
+    */
```
juanlofer-eprosima: Please update to describe current behavior.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,6 +1571,65 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+        EPROSIMA_LOG_WARNING(
+            DATA_WRITER,
+            "Deadline period is 0, it will be ignored from now on. Timer is going to be canceled.");
+
+        // Bump once and notify listener exactly once.
+        notify_deadline_missed_nts_();
+        return;
+    }
+
+    deadline_timer_->cancel_timer();
```
juanlofer-eprosima: Cancel? I was expecting `update_interval_millisec`.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,6 +1571,65 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+{
+    StatusMask notify_status = StatusMask::offered_deadline_missed();
+    if (auto* listener = get_listener_for(notify_status))
+    {
+        listener->on_offered_deadline_missed(user_datawriter_, deadline_missed_status_);
+    }
+    #ifdef FASTDDS_STATISTICS
+    writer_listener_.notify_status_observer(statistics::StatusKind::DEADLINE_MISSED);
+#endif // FASTDDS_STATISTICS
+    user_datawriter_->get_statuscondition().get_impl()->set_status(notify_status, true);
```
juanlofer-eprosima, line 1631: NIT: please preserve the original formatting, I find this compact version harder to read.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,28 +1576,88 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+
+void DataWriterImpl::notify_deadline_missed_nts_()
+{
     StatusMask notify_status = StatusMask::offered_deadline_missed();
-    auto listener = get_listener_for(notify_status);
-    if (nullptr != listener)
+    if (auto* listener = get_listener_for(notify_status))
     {
         listener->on_offered_deadline_missed(user_datawriter_, deadline_missed_status_);
-        deadline_missed_status_.total_count_change = 0;
```
juanlofer-eprosima, line 1586: I don't think this line should be deleted, although I am not sure it is valid for the infinite value case (I think it does but again, not sure).

src/cpp/fastdds/publisher/DataWriterImpl.hpp
```diff
@@ -643,6 +644,7 @@ class DataWriterImpl : protected rtps::IReaderDataFilter
 
     /**
      * @brief A method to reschedule the deadline timer
+     * @return true if it could compute and set a new interval, if there’s a pending deadline
```
juanlofer-eprosima:
```suggestion
     * @return true if deadline rescheduling succeeded, false otherwise
```

src/cpp/fastdds/publisher/DataWriterImpl.hpp
```diff
@@ -771,6 +773,19 @@ class DataWriterImpl : protected rtps::IReaderDataFilter
 
 private:
 
+    /**
+     * (Re)configures the deadline timer:
+     *  Create once, parked with a huge interval (idle).
```
juanlofer-eprosima:
```suggestion
```

src/cpp/fastdds/publisher/DataWriterImpl.hpp
```diff
@@ -771,6 +773,19 @@ class DataWriterImpl : protected rtps::IReaderDataFilter
...
+     *  In case of deadline period ∞ cancel it, for 0 warn and notify once; set counts to max and
+     *  for values >0 store period.
```
juanlofer-eprosima:
```suggestion
     *  In case of infinite deadline period cancel it, for 0 warn and notify once (with max counts), and
     *  for non-infinite positive values store period.
```

src/cpp/fastdds/publisher/DataWriterImpl.hpp
```diff
@@ -771,6 +773,19 @@ class DataWriterImpl : protected rtps::IReaderDataFilter
...
+     */
+    void configure_deadline_timer_();
+
+    /**
+     * Notifies listeners that a deadline has been missed without touching the counters.
```
juanlofer-eprosima: I would remove the "without touching the counters" part, if we eventually reset the change counter within the method (still to decide).

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1218,10 +1215,16 @@ ReturnCode_t DataWriterImpl::set_qos(
...
 
+    // Take a snapshot of the current QoS before mutating it
+    const DataWriterQos old_qos = qos_;
+
     set_qos(qos_, qos_to_set, !enabled);
 
     if (enabled)
     {
+        std::unique_lock<RecursiveTimedMutex> lock(writer_->getMutex());
+        // Locks after we checked that writer exists
```
juanlofer-eprosima: Don't take the mutex here, it is not thread safe as it is but let's tackle that task in other PR with the rest of changes needed.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,28 +1576,88 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+
+    if (qos_.deadline().period.to_ns() == 0)
+    {
+        deadline_timer_->cancel_timer();
+
+        deadline_missed_status_.total_count = std::numeric_limits<uint32_t>::max();
+        deadline_missed_status_.total_count_change = std::numeric_limits<uint32_t>::max();
+        EPROSIMA_LOG_WARNING(
+            DATA_WRITER,
+            "Deadline period is 0, it will be ignored from now on. Timer is going to be canceled.");
```
juanlofer-eprosima:
```suggestion
            "Deadline period is 0, it will be ignored from now on.");
```

juanlofer-eprosima (changes requested):

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: To be handled in a future PR.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
juanlofer-eprosima: To be done in another PR.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -72,6 +74,29 @@ class DeadlineQos : public testing::TestWithParam<communication_type>
...
+    // Update deadline period on a native DataWriter
+    static void set_deadline_seconds(
+            eprosima::fastdds::dds::DataWriter& native,
+            double seconds)
+    {
+        eprosima::fastdds::dds::DataWriterQos q;
+        ASSERT_EQ(native.get_qos(q), eprosima::fastdds::dds::RETCODE_OK);
+        q.deadline().period = seconds_to_duration(seconds);
+        ASSERT_EQ(native.set_qos(q), eprosima::fastdds::dds::RETCODE_OK);
+    }
```
juanlofer-eprosima: I don't think you need these methods, Time_t already has a constructor taking seconds.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +334,169 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+    auto prev = consumer_ptr->count();
+    std::this_thread::sleep_for(std::chrono::milliseconds(100));
+    EXPECT_EQ(consumer_ptr->count(), prev) << "Timer should be canceled; no more warnings";
+
+    eprosima::fastdds::dds::DataWriter& native = writer.get_native_writer();
+
+    DeadlineQos::set_deadline_seconds(native, 0.15); // Update 0 -> finite: counters must reset, no new warning
+
+    eprosima::fastdds::dds::OfferedDeadlineMissedStatus st{};
+    std::this_thread::sleep_for(std::chrono::milliseconds(150));
```
juanlofer-eprosima, line 381: Isn't this too tight? Couldn't the timer sometimes fire once?

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +334,169 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+/**
+ * Testing Redmine issue #23289.
+ * Regression test for the deadline period change at runtime.
+ * Creating a DataWriter with a small deadline so missed deadlines are registered.
+ * Changing deadline period to a large value to infer that there are no missed deadlines anymore.
+ * Checking total count and count change when changing the deadline
+ */
+TEST_P(DeadlineQos, ChangeDeadlineAtRuntime)
+{
+    // Writer/Reader setup (same pattern as your other tests)
```
juanlofer-eprosima:
```suggestion
```

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +334,169 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+    ASSERT_TRUE(reader.isInitialized());
+    writer.wait_discovery();
+    reader.wait_discovery();
+
+    eprosima::fastdds::dds::DataWriter& native = writer.get_native_writer();
+
+    // Arm the per-instance timer
+    KeyedHelloWorld sample;
+    sample.key(1);
+    ASSERT_EQ(native.write(&sample), eprosima::fastdds::dds::RETCODE_OK);
```
juanlofer-eprosima: I don't see the purpose of this part.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +334,169 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+    eprosima::fastdds::dds::OfferedDeadlineMissedStatus st{};
+    std::this_thread::sleep_for(std::chrono::milliseconds(150));
+    ASSERT_EQ(native.get_offered_deadline_missed_status(st), eprosima::fastdds::dds::RETCODE_OK);
+    EXPECT_GT(st.total_count, 0u);
+    EXPECT_GT(st.total_count_change, 0u);
+
+    // Switch to a large deadline
+    DeadlineQos::set_deadline_seconds(native, 5000.0);
+
+    // Immediately after the change, there can be one last miss in flight
```
juanlofer-eprosima: Maybe it would be better to make the deadline period greater and be stricter (not allow for any single extra deadline misses).

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
zesk1999: I wrote once to create the instance so deadline misses could be recorded. It is changed now.

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1072,8 +1068,10 @@ ReturnCode_t DataWriterImpl::perform_create_new_change(
             return RETCODE_TIMEOUT;
         }
 
-        if (qos_.deadline().period != dds::c_TimeInfinite)
+        if (qos_.deadline().period.to_ns() > 0 && qos_.deadline().period != dds::c_TimeInfinite &&
+                deadline_missed_status_.total_count < std::numeric_limits<uint32_t>::max())
         {
+            std::cout << "Create new change..." << std::endl;
```
juanlofer-eprosima:
```suggestion
```

src/cpp/fastdds/publisher/DataWriterImpl.cpp
```diff
@@ -1569,28 +1569,94 @@ bool DataWriterImpl::deadline_timer_reschedule()
...
+    assert(qos_.deadline().period != dds::c_TimeInfinite);
+
+    deadline_missed_status_.total_count++;
+    deadline_missed_status_.total_count_change++;
+    deadline_missed_status_.last_instance_handle = timer_owner_;
+
+    if (deadline_missed_status_.total_count == 3)
+    {
+        std::cout << "STOP" << std::endl;
+    }
```
juanlofer-eprosima:
```suggestion
```

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +313,185 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+    const auto post_change = writer.missed_deadlines_change();
+
+    EXPECT_EQ(pre_total, post_total) << "The total count should not change, as the timer was canceled.";
+    EXPECT_EQ(pre_change, post_change) << "The total_count_change should not change, as the timer was canceled.";
+    EXPECT_EQ(cptr->matched_global(), prev) << "No extra warnings after cancel.";
+
+    auto q = writer.get_qos();
+    q.deadline().period = Duration_t(0.1);
+
+    std::cout << "Setting non-zero deadline period" << std::endl;
```
juanlofer-eprosima: Please remove all couts.

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -1283,9 +1322,31 @@ bool DataReaderImpl::deadline_missed()
...
+    deadline_missed_status_.total_count++;
+    deadline_missed_status_.total_count_change++;
+    deadline_missed_status_.last_instance_handle = timer_owner_;
+
+    notify_deadline_missed_nts_();
+
+    // If we just reached the max -> log ONCE, stop timer, and bail
+    if (deadline_missed_status_.total_count == std::numeric_limits<uint32_t>::max())
+    {
+        EPROSIMA_LOG_WARNING(DATA_READER,
```
juanlofer-eprosima, line 1345: Piggyback request: please change all logs in this file to use category "DATA_READER".

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -1957,14 +2018,14 @@ ReturnCode_t DataReaderImpl::delete_contained_entities()
     // Check pending ReadConditions
     for (detail::ReadConditionImpl* impl : read_conditions_)
     {
-        // should be alive
+        // Should be alive
```
juanlofer-eprosima, line 2021: FYI it is totally fine for our style not to start with a capital letter in code comments. In fact, I am a fan of respecting the original author's stylistic choices and only make modifications to correct typos, bugs or unclear statements. Anyways that's just my personal preference.

test/utils/GenericLogCounter.hpp
```diff
@@ -0,0 +1,100 @@
+#pragma once
+
+#include <fastdds/dds/log/Log.hpp>
+#include <atomic>
+#include <array>
+#include <mutex>
+#include <string>
+#include <vector>
```
juanlofer-eprosima:
```suggestion
#include <atomic>
#include <array>
#include <mutex>
#include <string>
#include <vector>

#include <fastdds/dds/log/Log.hpp>
```

juanlofer-eprosima: Always add the copyright at the beginning of new files.

test/utils/GenericLogCounter.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace testutils {
+
+/** Header-only log consumer for tests.
+ * - Counts per Log::Kind (Warning, Error, Info, ...)
+ * - Optional storage of full entries (disabled by default)
+ * - "Needle" matcher for exact message occurrences
+ */
+class GenericLogCounter : public eprosima::fastdds::dds::LogConsumer
```
juanlofer-eprosima:
```suggestion
class LogCounter : public eprosima::fastdds::dds::LogConsumer
```

test/utils/GenericLogCounter.hpp
```diff
@@ -0,0 +1,100 @@
...
+        if (store_)
+        {
+            std::lock_guard<std::mutex> lk(m_);
+            entries_.push_back(e);
+        }
+    }
+
+private:
+
+    std::array<std::atomic<size_t>, 8> counts_{};
```
juanlofer-eprosima: I think a dictionary makes more sense here.

test/utils/LogCounter.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+private:
+
+    std::array<std::atomic<size_t>, 8> counts_{};
+    std::atomic<size_t> matched_{0};
+
+    bool store_;
+    mutable std::mutex m_;
+    std::string needle_;
+    std::vector<Log::Entry> entries_;
```
juanlofer-eprosima, line 113: More ideas for the future (future developers can extend your class if needed): split entries per category, allow several needles, needle per category, set a log kind threshold.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -22,9 +22,13 @@
 #include "PubSubReader.hpp"
 #include "PubSubWriter.hpp"
 
+#include "../../utils/GenericLogCounter.hpp"
```
juanlofer-eprosima:
```suggestion
#include <GenericLogCounter.hpp>
```

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -22,9 +22,13 @@
...
+
 using namespace eprosima::fastdds;
 using namespace eprosima::fastdds::rtps;
 
+namespace fastlog = eprosima::fastdds::dds;
```
juanlofer-eprosima: This is kind of weird, did you mean "namespace fastlog = eprosima::fastdds::dds::Log;" ?

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +313,185 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+ * Regression test for the zero-deadline period bug.
+ * Creating a DataWriter with a deadline of 0.
+ * Checking if a warning is logged exactly once, the timer is cancelled without missed deadline
+ * messages and a total count and count change set to max integer.
+ * Checking warnings, total count and count change when changing the deadline.
+ */
+TEST_P(DeadlineQos, ZeroDeadlinePeriodWriter)
+{
+    auto consumer = std::make_unique<testutils::GenericLogCounter>(/*store=*/ false);
+    auto* cptr = consumer.get();
```
juanlofer-eprosima: Why do you need to get and use the raw pointer instead of the unique pointer?

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
```diff
@@ -309,6 +313,185 @@ TEST_P(DeadlineQos, KeyedTopicBestEffortReaderVolatileWriterSetDeadline)
...
+TEST_P(DeadlineQos, ZeroDeadlinePeriodReader)
+{
+    PubSubWriter<KeyedHelloWorldPubSubType> writer(TEST_TOPIC_NAME);
+    PubSubReader<KeyedHelloWorldPubSubType> reader(TEST_TOPIC_NAME);
+
+    writer.durability_kind(eprosima::fastdds::dds::VOLATILE_DURABILITY_QOS);
+    writer.reliability(eprosima::fastdds::dds::RELIABLE_RELIABILITY_QOS);
+    reader.reliability(eprosima::fastdds::dds::RELIABLE_RELIABILITY_QOS);
+
+    writer.deadline_period(0.0).init();
```
juanlofer-eprosima, line 419: Add a comment saying that needs to be 0 in order to fulfill the matching rule.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
juanlofer-eprosima, line 1: If you still have some strength, it would be great to add a test of the finite to 0 update case.

juanlofer-eprosima (changes requested):

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
zesk1999: It is already there.

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
```diff
@@ -288,7 +283,7 @@ ReturnCode_t DataReaderImpl::enable()
     SubscriptionBuiltinTopicData subscription_data;
     if (get_subscription_builtin_topic_data(subscription_data) != RETCODE_OK)
     {
-        EPROSIMA_LOG_ERROR(DATA_WRITER, "Error getting subscription data. RTPS Writer not enabled.");
+        EPROSIMA_LOG_ERROR(DATA_READER, "Error getting subscription data. RTPS Writer not enabled.");
```
juanlofer-eprosima: Please fix the typo as well.

src/cpp/fastdds/subscriber/DataReaderImpl.cpp
juanlofer-eprosima: Please also change SUBSCRIBER -> DATA_READER

test/utils/GenericLogCounter.hpp
juanlofer-eprosima: Please also rename file.

test/utils/GenericLogCounter.hpp
juanlofer-eprosima: For the copyright, use the year in which the file is created.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
juanlofer-eprosima: I still think you don't need that extra variable.

test/blackbox/common/BlackboxTestsDeadlineQos.cpp
juanlofer-eprosima: I actually meant a dedicated test, but it is good enough for me as it is.

juanlofer-eprosima:
https://github.com/Mergifyio backport 3.3.x 3.2.x 2.14.x

PR #6066, opened 2025-09-30
[23655] Solve TCP disconnect while incomplete read deadlock

juanlofer-eprosima:
## Description

This PR attempts to solve the deadlock produced when a read operation never completes (e.g. peer blocks before ending a transmission), and the TCP channel resource is disconnected (normally on closure while cleaning the associated transport).
Summing up, the solution consists in shutting down the socket before acquiring the culprit read mutex, so that any ongoing write/read operations are aborted even if not completed.

In addition, some other changes in this PR target general thread safety and attempt to improve overall resilience:
- New "disconnecting" state to protect from concurrent connect/disconnects.
- Assert that socket options are only configured before connection is established.

@Mergifyio backport 3.3.x 3.2.x 2.14.x

## Contributor Checklist

- _N/a_ Any new/modified methods have been properly documented using Doxygen. 
- _N/a_ Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
- _N/a_ New feature has been added to the `versions.md` file (if applicable).
- _N/a_ New feature has been documented/Current behavior is correctly described in the documentation. 
    
## Reviewer Checklist

- _N/a_ If this is a critical bug fix, backports to the critical-only supported branches have been requested.

### DIFF
--- a/src/cpp/rtps/transport/TCPChannelResource.cpp
+++ b/src/cpp/rtps/transport/TCPChannelResource.cpp
@@ -66,7 +66,7 @@ TCPChannelResource::TCPChannelResource(
     , parent_(parent)
     , locator_()
     , waiting_for_keep_alive_(false)
-    , connection_status_(eConnectionStatus::eConnected)
+    , connection_status_(eConnectionStatus::eDisconnected)
     , tcp_connection_type_(TCPConnectionType::TCP_ACCEPT_TYPE)
 {
 }
@@ -376,6 +376,9 @@ void TCPChannelResource::set_socket_options(
         asio::basic_socket<asio::ip::tcp>& socket,
         const TCPTransportDescriptor* options)
 {
+    // Options setting should be done before connection is established
+    assert(!connected());
+
     uint32_t minimum_value = options->maxMessageSize;
 
     // Set the send buffer size
--- a/src/cpp/rtps/transport/TCPChannelResource.h
+++ b/src/cpp/rtps/transport/TCPChannelResource.h
@@ -53,7 +53,8 @@ protected:
 
     enum eConnectionStatus
     {
-        eDisconnected = 0,
+        eDisconnecting = -1,        // Transition state to disconnected, used to protect from concurrent (dis)connects.
+        eDisconnected = 0,          // Initial state.
         eConnecting,                // Output -> Trying connection.
         eConnected,                 // Output -> Send bind message.
         eWaitingForBind,            // Input -> Waiting for the bind message.
@@ -71,7 +72,6 @@ protected:
     std::vector<uint16_t> pending_logical_output_ports_; // Must be accessed after lock pending_logical_mutex_
     std::vector<uint16_t> logical_output_ports_;
     std::condition_variable_any logical_output_ports_updated_cv;
-    std::mutex read_mutex_;
     std::recursive_mutex pending_logical_mutex_;
     std::atomic<eConnectionStatus> connection_status_;
 
@@ -111,6 +111,17 @@ public:
     bool connection_established()
     {
         return connection_status_ == eConnectionStatus::eEstablished;
+    }
+
+    bool connected()
+    {
+        return connection_status_ >= eConnectionStatus::eConnected;
+    }
+
+    bool disconnected()
+    {
+        // NOTE: covers both eDisconnected and eDisconnecting states
+        return connection_status_ <= eConnectionStatus::eDisconnected;
     }
 
     eConnectionStatus connection_status()
@@ -275,7 +286,7 @@ protected:
      * @param socket Socket on which to set the options.
      * @param options Descriptor with the options to set.
      */
-    static void set_socket_options(
+    void set_socket_options(
             asio::basic_socket<asio::ip::tcp>& socket,
             const TCPTransportDescriptor* options);
 
--- a/src/cpp/rtps/transport/TCPChannelResourceBasic.cpp
+++ b/src/cpp/rtps/transport/TCPChannelResourceBasic.cpp
@@ -191,7 +191,8 @@ void TCPChannelResourceBasic::set_options(
 
 void TCPChannelResourceBasic::cancel()
 {
-    socket_->cancel(); // thread safe with respect to asio's read and write methods
+    std::error_code ec;
+    socket_->cancel(ec); // thread safe with respect to asio's read and write methods
 }
 
 void TCPChannelResourceBasic::close()
@@ -202,13 +203,15 @@ void TCPChannelResourceBasic::close()
     std::unique_lock<std::mutex> read_lk(read_mutex_, std::defer_lock);
     std::lock(send_lk, read_lk); // Pre C++17 alternative to std::scoped_lock
 
-    socket_->close();
+    std::error_code ec;
+    socket_->close(ec);
 }
 
 void TCPChannelResourceBasic::shutdown(
         asio::socket_base::shutdown_type what)
 {
-    socket_->shutdown(what); // thread safe with respect to asio's read and write methods
+    std::error_code ec;
+    socket_->shutdown(what, ec); // thread safe with respect to asio's read and write methods
 }
 
 } // namespace rtps
--- a/src/cpp/rtps/transport/TCPChannelResourceBasic.cpp
+++ b/src/cpp/rtps/transport/TCPChannelResourceBasic.cpp
@@ -14,8 +14,9 @@
 
 #include <rtps/transport/TCPChannelResourceBasic.h>
 
-#include <future>
 #include <array>
+#include <future>
+#include <mutex>
 
 #include <asio.hpp>
 #include <fastdds/utils/IPLocator.hpp>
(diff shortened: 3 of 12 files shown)


src/cpp/rtps/transport/TCPChannelResource.cpp
```diff
@@ -375,6 +375,9 @@ void TCPChannelResource::set_socket_options(
         asio::basic_socket<asio::ip::tcp>& socket,
         const TCPTransportDescriptor* options)
 {
+    // Options setting should be done before connection is established
```
juanlofer-eprosima, line 379: Please double check this is indeed the case. At least today it is I believe, and if willing to change that in the future, it should be assured the socket is modified in a thread safe manner.

src/cpp/rtps/transport/TCPChannelResourceSecure.cpp
```diff
@@ -339,6 +353,7 @@ void TCPChannelResourceSecure::close()
 void TCPChannelResourceSecure::shutdown(
         asio::socket_base::shutdown_type)
 {
+    // WARNING: This function blocks until receiving the peer’s close_notify
```
juanlofer-eprosima: After some reading I found this possibility, however I am not considering dealing with it in this PR.

src/cpp/rtps/transport/TCPTransportInterface.cpp
```diff
@@ -1744,14 +1746,11 @@ void TCPTransportInterface::SocketConnected(
 
         if (channel)
         {
-            if (!error)
+            if (!error && TCPChannelResource::eConnectionStatus::eConnecting == channel->connection_status())
```
juanlofer-eprosima, line 1749: I changed this condition to something I believe makes more sense, but please double check I am not missing a transition covered by the prior condition.

src/cpp/rtps/transport/TCPChannelResourceBasic.cpp
```diff
@@ -195,23 +186,32 @@ asio::ip::tcp::endpoint TCPChannelResourceBasic::local_endpoint(
...
+    std::error_code ec;
+    socket_->close(ec);
 }
 
 void TCPChannelResourceBasic::shutdown(
         asio::socket_base::shutdown_type what)
 {
-    socket_->shutdown(what);
+    std::error_code ec;
+    socket_->shutdown(what, ec); // thread safe with respect to asio's read and write methods
```
juanlofer-eprosima, line 214: Note this behavior change, I believe it is expected that this class method does not throw exceptions, even if it is not noexcept. Same for cancel and close above, and its secure counterpart.

src/cpp/rtps/transport/TCPChannelResource.h
```diff
@@ -113,6 +113,17 @@ class TCPChannelResource : public ChannelResource
...
     }
 
+    bool connected()
+    {
+        return connection_status_ >= eConnectionStatus::eConnected;
+    }
+
+    bool disconnected()
+    {
+        // NOTE: covers both eDisconnected and eDisconnecting states
```
cferreiragonz, line 123: Is it not redundant to write `NOTE` in a comment? I believe the idiosyncrasy of comments is to act as notes

test/blackbox/common/BlackboxTestsTransportTCP.cpp
```diff
@@ -1591,6 +1593,58 @@ TEST_P(TransportTCP, large_data_tcp_no_frag)
...
+    // Wait for data to be received by the server
+    std::this_thread::sleep_for(std::chrono::seconds(3));
+
+    // Destroy participant (and its TCP transport) while read is ongoing
+    delete replier;
+
+    // Close client's socket
+    socket.shutdown(asio::ip::tcp::socket::shutdown_both, ec);
+    socket.cancel(ec);
+    socket.close(ec);
```
cferreiragonz, line 1645: Should we verify that no unexpected error code is raised?

test/unittest/transport/TCPv4Tests.cpp
```diff
@@ -1536,24 +1536,17 @@ TEST_F(TCPv4Tests, secure_non_blocking_send)
...
-            });
+    // Synchronous socket connection
+    std::error_code ec;
+    asio::connect(secure_socket->lowest_layer(), endpoints, ec);
+    ASSERT_TRUE(!ec);
+
+    // Synchronous handshake
+    asio::ssl::stream_base::handshake_type role = asio::ssl::stream_base::client;
+    secure_socket->handshake(role, ec);
+    ASSERT_TRUE(!ec);
```
cferreiragonz, line 1547: Not sure about this. Method `connect()` of TCPChannelResourceSecure is async. Maybe by making it synchronous we are creating a different scenario than the real application code

test/unittest/transport/TCPv4Tests.cpp
```diff
@@ -2135,9 +2117,9 @@ TEST_F(TCPv4Tests, non_blocking_send)
     bytes_sent = sender_channel_resource->send(nullptr, 0, buffer_list, size, ec);
     ASSERT_EQ(bytes_sent, 0u);
 
-    socket.shutdown(asio::ip::tcp::socket::shutdown_both);
-    socket.cancel();
-    socket.close();
+    socket.shutdown(asio::ip::tcp::socket::shutdown_both, ec);
+    socket.cancel(ec);
+    socket.close(ec);
```
cferreiragonz, line 2124: Again, should we check the ec?

src/cpp/rtps/transport/TCPTransportInterface.cpp
cferreiragonz: It looks good to me too. Only calling this method after a `connect()` operation.

src/cpp/rtps/transport/TCPChannelResourceSecure.cpp
cferreiragonz: "Or until an error occurs"

src/cpp/rtps/transport/TCPChannelResourceSecure.cpp
```diff
@@ -138,19 +132,33 @@ void TCPChannelResourceSecure::connect(
...
-                    socket->lowest_layer().close(ec);
                     socket->async_shutdown([&, socket](const std::error_code&)
                     {
                     });
+
+                    // Close the underlying socket after SSL shutdown
+                    // NOTE: the (async) SSL shutdown may not complete before the socket is closed
+                    socket->lowest_layer().shutdown(asio::ip::tcp::socket::shutdown_both, ec);
+                    socket->lowest_layer().cancel(ec);
+                    socket->lowest_layer().close(ec);
```
cferreiragonz, line 152: What happens if an error code is set?

src/cpp/rtps/transport/TCPChannelResource.h
juanlofer-eprosima: Not for me. I tend to consider general comments to be simple explanations that make interpreting the code easier, and use the "NOTE" introduction to point out something might not be as expected. In this case the expected behavior could have been that it returns true only when in disconnected state.

test/blackbox/common/BlackboxTestsTransportTCP.cpp
juanlofer-eprosima: Do you mean when an error code other than 0 is returned? Or a return code other than 0 or the one corresponding to the "not connected" case? If it's the first case, I don't think so because it might fail in platforms other than Ubuntu.

cferreiragonz: The second case. Any error code that points to an incorrect shutdown

src/cpp/rtps/transport/TCPChannelResource.h
cferreiragonz: And in this case `covers both eDisconnected and eDisconnecting states` cannot be categorized inside the "_simple explanations that make interpreting the code easier_" group? Anyways, both approaches are fine

test/unittest/transport/TCPv4Tests.cpp
juanlofer-eprosima: I don't think that makes sense for this test, we are testing the sending part here. The receiver should be connected before doing any reading, so we should move forward only after that occurs. With such sleep there was enough time to asynchronously connect, but I believe this way is more correct. Had there not been enough time to connect the read operation would fail anyways, but with the change it is explicit where the unexpected condition is given.

test/unittest/transport/TCPv4Tests.cpp
juanlofer-eprosima: Same answer.

src/cpp/rtps/transport/TCPChannelResourceSecure.cpp
juanlofer-eprosima: Nothing, show must go on.

cferreiragonz: My concerning is that these methods raised an exception before. Now we are silently ignoring any possible error. Not sure about how these might affect the whole TCP flow

juanlofer-eprosima: You are right, related to this comment https://github.com/eProsima/Fast-DDS/pull/6066#discussion_r2401600789
Although it seems cancel and close were actually not being called at all.

cferreiragonz (approved):
LGTM, failing tests seems unrelated and Data Races have not been increased comparing TSAN report with last nightly reports

juanlofer-eprosima:
https://github.com/Mergifyio backport 3.3.x 3.2.x 2.14.x

PR #6100, opened 2025-10-14
[23822] Protect PDPStatelessWriter attributes

juanlofer-eprosima:
## Description

This PR attempts to solve the data race produced when a thread clears the attributes of PDPStatelssWriter at mark_all_readers_interested (called in announceParticipantState from events thread, for example), while the values are read from another thread (send_to_fixed_locators called, for example, from the asynchronous send thread → LocatorSelectorSender::send → StatelessWriter::send_nts → PDPStatelessWriter::send_to_fixed_locators).

@Mergifyio backport 3.3.x 3.2.x 2.14.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/builtin/discovery/participant/simple/PDPStatelessWriter.cpp
+++ b/src/cpp/rtps/builtin/discovery/participant/simple/PDPStatelessWriter.cpp
@@ -145,7 +145,8 @@ bool PDPStatelessWriter::is_relevant(
 
 void PDPStatelessWriter::mark_all_readers_interested()
 {
-    std::lock_guard<std::mutex> guard(interested_readers_mutex_);
+    std::lock_guard<RecursiveTimedMutex> guard(mp_mutex);
+    std::lock_guard<std::mutex> _guard(interested_readers_mutex_);
     should_reach_all_destinations_ = true;
     interested_readers_.clear();
     fixed_locators_.clear();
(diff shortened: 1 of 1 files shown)


src/cpp/rtps/builtin/discovery/participant/simple/PDPStatelessWriter.cpp
```diff
@@ -55,6 +55,7 @@ PDPStatelessWriter::PDPStatelessWriter(
 bool PDPStatelessWriter::matched_reader_add_edp(
         const ReaderProxyData& data)
 {
+    std::lock_guard<RecursiveTimedMutex> guard(mp_mutex);
```
juanlofer-eprosima, line 58: Taking the mutex here and in the next two methods does not seem required, however I thought it is better to make this call atomic with respect to the parent methods (taking the mutex), and in the case of this particular method, with respect to `reschedule_all_samples`.

juanlofer-eprosima: Dear reviewer, please also verify that no mutex should be taken in `send_periodic_announcement`

src/cpp/rtps/builtin/discovery/participant/simple/PDPStatelessWriter.cpp
```diff
@@ -106,6 +109,8 @@ bool PDPStatelessWriter::send_to_fixed_locators(
         const uint32_t& total_bytes,
         std::chrono::steady_clock::time_point& max_blocking_time_point) const
 {
+    std::lock_guard<std::mutex> guard(interested_readers_mutex_);
```
juanlofer-eprosima, line 112: Although not reported by TSAN, I am concerned about the modification of `fixed_locators_` in this method only with the new mutex taken, while `reschedule_all_samples` ends up reading its value with only `mp_mutex` taken. I did not know how to solve this, because taking the new mutex in `reschedule_all_samples` would produce a potential deadlock with respect to the locator selector mutex. Surrender isn't weakness, it's the courage to face reality as it is.

MiguelCompany:
@Mergifyio backport 3.3.x 3.2.x 2.14.x

PR #6290, opened 2026-01-30
[24066] Enhance filtering logic for virtual and external endpoints

MiguelCompany:
## Description

This PR ensures that Servers only communicate with their direct clients.

## Contributor Checklist

- _N/A_: Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- _N/A_: Changes are API compatible. 
    
## Reviewer Checklist

- _N/A_: If this is a critical bug fix, backports to the critical-only supported branches have been requested.

### DIFF
--- a/test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
+++ b/test/blackbox/common/DDSBlackboxTestsDiscovery.cpp
@@ -2050,9 +2050,9 @@ TEST(DDSDiscovery, client_server_participants_with_different_domain_ids_discover
     client_domain_3.init();
     EXPECT_TRUE(client_domain_3.isInitialized());
 
-    server_domain_1.wait_discovery(std::chrono::seconds(2));
-    client_domain_2.wait_discovery(std::chrono::seconds(2));
-    client_domain_3.wait_discovery(std::chrono::seconds(2));
+    server_domain_1.wait_discovery(std::chrono::seconds(5));
+    client_domain_2.wait_discovery(2, std::chrono::seconds(5));
+    client_domain_3.wait_discovery(std::chrono::seconds(5));
 
     ASSERT_TRUE(client_domain_2.is_matched());
     ASSERT_TRUE(client_domain_3.is_matched());


src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1197,31 +1199,64 @@ void DiscoveryDataBase::match_writer_reader_(
...
     DiscoveryParticipantInfo& reader_participant_info = p_rit->second;
 
-    // Always exchange all information between all participants (no filtering)
-    // Skip only if it's a virtual endpoint matching with another virtual endpoint
+    // virtual              - needs info and give none
+    // local                - needs info and give info
+    // external             - needs none and give info
+    // writer needs info    = add writer participant in reader ack list
+    // writer give info     = add reader participant in writer ack list
+
```
cferreiragonz: This info is also written below in another comment. I would delete this one

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1197,31 +1199,64 @@ void DiscoveryDataBase::match_writer_reader_(
...
+    // - Virtual endpoints only need info (never give)
+    // - External endpoints only give info (never need)
+    // Additionally, servers do not redirect between remote clients,
+    // so external endpoints only exchange with local endpoints
+    const bool writer_needs_info = writer_is_virtual || writer_is_local;
+    const bool writer_gives_info = writer_is_local || (writer_is_external && reader_is_local);
+
+    const bool reader_needs_info = reader_is_virtual || reader_is_local;
+    const bool reader_gives_info = reader_is_local || (reader_is_external && writer_is_local);
+
```
cferreiragonz: I believe we still might be missing some cases. For example:
If a writer is local (gives and needs) and a reader is virtual (only needs) we should only allow information flow from the writer to the reader. 

However, if the reader is virtual AND local (two servers connected between each other), the reader still gives info to the writer, as `writer_needs_info` is True because it is local and `reader_gives_info` is True because it is also local. Note that an endpoint can be both local and virtual, they are not exclusive.

We could do something like `const bool reader_gives_info = !reader_is_virtual && (reader_is_local || (reader_is_external && writer_is_local));`

cferreiragonz (changes requested):

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1197,31 +1199,58 @@ void DiscoveryDataBase::match_writer_reader_(
...
+    const bool reader_is_external = !reader_is_virtual && !reader_is_local;
+
+    // Determine info exchange based on endpoint types:
+    // - Local endpoints both need and give info
+    // - Virtual endpoints only need info (never give)
+    // - External endpoints only give info (never need)
+    // Additionally, servers do not redirect between remote clients,
+    // so external endpoints only exchange with local endpoints
+    const bool writer_needs_info = writer_is_virtual || writer_is_local;
+    const bool writer_gives_info = writer_is_local || (writer_is_external && reader_is_local);
```
cferreiragonz: Following the same logic as in the `reader_gives_info`, we need to ensure that the writer is not virtual when it gives info:

```suggestion
    const bool writer_gives_info = !writer_is_virtual && (writer_is_local || (writer_is_external && reader_is_local));
```

This prevents the writer giving info to the reader in the case where the writer is external and virtual, and the reader is local and not virtual.

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1197,31 +1199,58 @@ void DiscoveryDataBase::match_writer_reader_(
     }
     DiscoveryParticipantInfo& reader_participant_info = p_rit->second;
 
-    // Always exchange all information between all participants (no filtering)
-    // Skip only if it's a virtual endpoint matching with another virtual endpoint
+    // Servers do not redirect Data(p) between remote clients,
+    // so we skip matching when both endpoints are virtual
```
cferreiragonz:
```suggestion
    // Virtual endpoints do not exchange info
```
I believe this comment or similar is more accurate than mentioning remote clients here

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
```diff
@@ -1775,8 +1804,10 @@ void DiscoveryDataBase::AckedFunctor::operator () (
                     auto remote_server_it = db_->participants_.find(*it);
                     if (remote_server_it == db_->participants_.end())
                     {
-                        EPROSIMA_LOG_INFO(DISCOVERY_DATABASE, "Change " << change_->instanceHandle <<
-                                "check as acked for " << reader_proxy->guid() << " as it has not answered pinging yet");
+                        EPROSIMA_LOG_INFO(DISCOVERY_DATABASE, "Change " << change_->instanceHandle
+                                                                        << "check as acked for "
+                                                                        << reader_proxy->guid()
+                                                                        << " as it has not answered pinging yet");
```
cferreiragonz: Try with this:

```suggestion
                        EPROSIMA_LOG_INFO(DISCOVERY_DATABASE,
                                "Change " << change_->instanceHandle << "check as acked for " << reader_proxy->guid()
                                          << " as it has not answered pinging yet");
```

I don't really understand why uncrustify is doing this

test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -2202,6 +2202,148 @@ TEST_P(Discovery, discovery_server_pdp_messages_sent)
...
+// discovery information about clients connected to a different server from a server
+// different than their own.
+TEST_P(Discovery, discovery_server_no_external_to_external_relay)
+{
+    // Skip test in intraprocess and datasharing mode
+    if (TRANSPORT != GetParam())
+    {
+        GTEST_SKIP() << "Only makes sense on TRANSPORT";
+        return;
+    }
```
cferreiragonz:
```suggestion
TEST(Discovery, discovery_server_no_external_to_external_relay)
{
```
We do not need to use the Parametrized test macro here as we only want to run this test once. TEST_P creates two more tests that will be skipped. I believe other tests in this file suffer for the same issue, this was bad design.

test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -2202,6 +2202,148 @@ TEST_P(Discovery, discovery_server_pdp_messages_sent)
...
+    // Start server 2
+    ASSERT_TRUE(server_2->init_participant());
+
+    // Wait for servers to discover each other
+    server_1->wait_discovery(std::chrono::seconds(5), 1, true);
+    server_2->wait_discovery(std::chrono::seconds(5), 1, true);
+
+    // Record baseline message count after server federation
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    size_t baseline_count = num_messages_to_client_1.load(std::memory_order_seq_cst);
```
cferreiragonz, line 2276: This must always be 0. We could ASSERT it here with something like: `ASSERT_EQ(baseline_count, 0u)`.
Or directly compare `num_messages_to_client_1` with 0 at the end of the test.

test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -2202,6 +2202,148 @@ TEST_P(Discovery, discovery_server_pdp_messages_sent)
...
+
+    // Start server 2
+    ASSERT_TRUE(server_2->init_participant());
+
+    // Wait for servers to discover each other
+    server_1->wait_discovery(std::chrono::seconds(5), 1, true);
+    server_2->wait_discovery(std::chrono::seconds(5), 1, true);
+
+    // Record baseline message count after server federation
+    std::this_thread::sleep_for(std::chrono::seconds(1));
```
cferreiragonz:
```suggestion
```
We don't really need to wait here. Same with the next one. We only wait for discovery which is ensured by `wait_discovery`. The last sleep is the one which can matter

test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -2202,6 +2202,148 @@ TEST_P(Discovery, discovery_server_pdp_messages_sent)
...
+
+    ASSERT_TRUE(client_1.isInitialized());
+
+    std::cout << "Client 1 metatraffic port set to: " << client_1_metatraffic_port << std::endl;
+
+    // Wait for server_1 to discover client_1
+    server_1->wait_discovery(std::chrono::seconds(5), 2, true);  // server_2 + client_1
+
+    // Record message count after client_1 connects
+    std::this_thread::sleep_for(std::chrono::seconds(2));
```
cferreiragonz:
```suggestion
```

test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -2202,6 +2202,148 @@ TEST_P(Discovery, discovery_server_pdp_messages_sent)
...
+    size_t messages_after_client_1 = num_messages_to_client_1.load(std::memory_order_seq_cst);
```
cferreiragonz, line 2303: Should we assert to 0 again here? In this way if the test fails here we are able to know that the issue is before the second client is initialized

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
cferreiragonz: This is what we discussed internally. We need to protect against the case where the writer is virtual and local and the reader is local and not virtual. The writer should not give info in this case.

test/blackbox/common/BlackboxTestsDiscovery.cpp
```diff
@@ -2202,6 +2202,148 @@ TEST_P(Discovery, discovery_server_pdp_messages_sent)
...
+    client_1_qos.builtin.discovery_config.initial_announcements.count = 1;
+    client_1_qos.builtin.metatrafficUnicastLocatorList.push_back(client_1_metatraffic_locator);
+
+    client_1.set_wire_protocol_qos(client_1_qos)
+            .setup_transports(eprosima::fastdds::rtps::BuiltinTransports::UDPv4)
+            .init();
+
+    ASSERT_TRUE(client_1.isInitialized());
+
+    std::cout << "Client 1 metatraffic port set to: " << client_1_metatraffic_port << std::endl;
```
cferreiragonz:
```suggestion
```

test/blackbox/common/BlackboxTestsDiscovery.cpp
raulojeda22: ASSERT_EQ(baseline_count, 0u)

src/cpp/rtps/builtin/discovery/database/DiscoveryDataBase.cpp
cferreiragonz: Your logic is already correct. I would just modify `writer_is_local` variable name as it can be missleading, to `writer_is_local_not_virtual`

Issue #6331, opened 2026-03-15
Fast-DDS crashes in loadDefaultXMLFile() in Docker containers (2.6.x)


PavelGuzenfeld:
## Description

Fast-DDS crashes in `XMLProfileManager::loadDefaultXMLFile()` when running inside x86_64 Docker containers. The crash occurs during DDS participant initialization, before any user code interacts with the XML configuration.

## Environment

- **Architecture:** x86_64
- **Platform:** Docker containers (Ubuntu-based)
- **ROS2 Distro:** Humble (also observed in Jazzy-based containers rebuilding Fast-DDS from the 2.6.x branch)
- **Fast-DDS version:** 2.6.x branch (built from source via `https://github.com/eProsima/Fast-DDS.git`, branch `2.6.x`)
- **Fast-CDR version:** 1.0.x branch

## Workaround

The following environment variables prevent the crash:

```bash
ENV SKIP_DEFAULT_XML_FILE=1
ENV FASTDDS_BUILTIN_TRANSPORTS=DEFAULT
```

Additionally, we also set `LD_PRELOAD=/lib/x86_64-linux-gnu/libgcc_s.so.1` to work around a related libunwind/libgcc_s exception handling conflict that also causes crashes in Docker environments (filed separately as #6330).

## Analysis

- `SKIP_DEFAULT_XML_FILE` was introduced via PR #1350 as an officially supported environment variable, which suggests this class of issue was anticipated.
- Multiple null-dereference fixes in the XML parser have been applied across various Fast-DDS versions (2.6.8, 2.10.0, 2.11.0, 2.12.0, etc.). It is possible that not all of these fixes have been fully backported to the 2.6.x maintenance branch.
- The crash does **not** occur on aarch64 (Jetson Orin) with the same ROS2/Fast-DDS version, suggesting it may be related to platform-specific XML file resolution or library loading behavior in x86_64 Docker environments.

## What we don't have yet

We do not currently have a full stack trace for this crash. The workaround was applied early in our Docker image development. We are happy to reproduce the crash without the workaround and provide a detailed backtrace and core dump if that would be helpful.

## Questions

1. Are all XML parser null-dereference fixes from newer versions (2.10+, 2.11+, 2.12+) fully backported to the 2.6.x branch?
2. Is there a known issue with `loadDefaultXMLFile()` when no default XML file exists in the container filesystem?
3. Is 2.6.x still receiving maintenance fixes, or should we plan to migrate to a newer branch?

## References

[...]

PR #5668, opened 2025-02-27
[22843]+[22844] Solve fuzz XMLParser Null-dereference

juanjo4936:
## Description

This PR solves two Null-dereference issues from the XMLParser found in oss-fuzz. Added regression test, and fixed by adding an error message preventing an empty map to be built.

@Mergifyio backport 3.1.x 2.14.x 2.10.x

## Contributor Checklist

- [x] Commit messages follow the project guidelines. 
- [x] The code follows the style guidelines of this project. 
- [x] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- _N/A_ Any new/modified methods have been properly documented using Doxygen. 
- _N/A_ Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
- [x] Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- [x] Changes are API compatible. 
- _N/A_ New feature has been added to the `versions.md` file (if applicable).
-  _N/A_ New feature has been documented/Current behavior is correctly described in the documentation. 
    
- [x] Applicable backports have been included in the description.

## Reviewer Checklist

- [ ] The PR has a milestone assigned.
- [ ] The title and description correctly express the PR's purpose.
- [ ] Check contributor checklist is correct.
- [ ] If this is a critical bug fix, backports to the critical-only supported branches have been requested.
- [ ] Check CI results: changes do not issue any warning.
- [ ] Check CI results: failing tests are unrelated with the changes.

PR #5080, opened 2024-07-17
[21334] Handle nullptr in parseXMLStructDynamicType

elianalf:
## Description

This PR handles a nullptr in `parseXMLStructDynamicType`

## Contributor Checklist

- [x] Commit messages follow the project guidelines. 
- [x] The code follows the style guidelines of this project. 
- [x] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- [x] Any new/modified methods have been properly documented using Doxygen. 
- N/A Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
- __NO__ Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- __NO__ Changes are API compatible. 
- N/A New feature has been added to the `versions.md` file (if applicable).
- N/A New feature has been documented/Current behavior is correctly described in the documentation. 
    
- N/A Applicable backports have been included in the description.

## Reviewer Checklist

- [x] The PR has a milestone assigned.
- [x] The title and description correctly express the PR's purpose.
- [x] Check contributor checklist is correct.
- **N/A** If this is a critical bug fix, backports to the critical-only supported branches have been requested.
- [x] Check CI results: changes do not issue any warning.
- [x] Check CI results: failing tests are unrelated with the changes.

PR #5007, opened 2024-06-28
[21223] Fix parse XML member dynamic type fuzzer issue

elianalf:
## Contributor Checklist

- [x] Commit messages follow the project guidelines. 
- [x] The code follows the style guidelines of this project. 
- [x] Tests that thoroughly check the new feature have been added/Regression tests checking the bug and its fix have been added; the added tests pass locally 
- [x] Any new/modified methods have been properly documented using Doxygen. 
- N/A Any new configuration API has an equivalent XML API (with the corresponding XSD extension) 
- [x] Changes are backport compatible: they do **NOT** break ABI nor change library core behavior. 
- [x] Changes are API compatible. 
- N/A New feature has been added to the `versions.md` file (if applicable).
- N/A New feature has been documented/Current behavior is correctly described in the documentation. 
    
- N/A Applicable backports have been included in the description.

## Reviewer Checklist

- [x] The PR has a milestone assigned.
- [x] The title and description correctly express the PR's purpose.
- [x] Check contributor checklist is correct.
- _N/A_ If this is a critical bug fix, backports to the critical-only supported branches have been requested.
- [x] Check CI results: changes do not issue any warning.
- [x] Check CI results: failing tests are unrelated with the changes.

PR #6333, opened 2026-03-15
Fix XML parser null-dereference bugs on 2.6.x

PavelGuzenfeld:
## Summary

Fixes multiple null-dereference bugs in the XML parser on the 2.6.x branch that can cause crashes when loading XML configuration files, including the default XML file at startup via `loadDefaultXMLFile()`.

These are the 2.6.x equivalents of issues found by oss-fuzz on master (PRs #4961, #5007, #5080, #5565, #5668). The master fixes cannot be cherry-picked because the Dynamic Types API was completely rewritten between 2.6.x and 2.10.x, so these are manually written fixes for the same class of bugs.

### 1. `parseXMLConsumer()` in `XMLParser.cpp` (4 locations)

`TinyXML2::GetText()` can return `nullptr` when XML elements have no text content (e.g., `<class/>`, `<name></name>`). The return value was passed directly to `std::string` constructors, which is undefined behavior (typically segfault). Added null checks with error logging before each `std::string` construction.

- Line 1514: `<class>` element text
- Line 1548: StdoutErrConsumer `<name>` element text
- Line 1571: StdoutErrConsumer `<value>` element text
- Line 1651: FileConsumer `<name>` element text

### 2. `parseXMLUnionDynamicType()` in `XMLDynamicParser.cpp`

`Attribute(TYPE)` can return `nullptr` when the `<discriminator>` element lacks a `type` attribute. The null `const char*` was passed to `getDiscriminatorTypeBuilder()` whose parameter is `const std::string&`, causing undefined behavior. Added null check before use.

### 3. `extractXxxProfile()` methods in `XMLProfileManager.cpp` (6 locations)

`dynamic_cast` results were dereferenced without null checks. If the cast fails, the resulting `nullptr` is immediately used to call `getAttributes()`. Added null checks after each `dynamic_cast`.

- `extractParticipantProfile()` (line 523)
- `extractPublisherProfile()` (line 556)
- `extractSubscriberProfile()` (line 589)
- `extractTopicProfile()` (line 679)
- `extractRequesterProfile()` (line 712)
- `extractReplierProfile()` (line 740)

## Files Changed

- `src/cpp/rtps/xmlparser/XMLParser.cpp`
- `src/cpp/rtps/xmlparser/XMLDynamicParser.cpp`
- `src/cpp/rtps/xmlparser/XMLProfileManager.cpp`

## References

- Issue: #6331
- Related oss-fuzz fixes on master: #4961, #5007, #5080, #5565, #5668

### DIFF
--- a/src/cpp/rtps/xmlparser/XMLDynamicParser.cpp
+++ b/src/cpp/rtps/xmlparser/XMLDynamicParser.cpp
@@ -868,17 +868,24 @@ XMLP_ret XMLParser::parseXMLUnionDynamicType(
     if (p_element != nullptr)
     {
         const char* disc = p_element->Attribute(TYPE);
-        p_dynamictypebuilder_t discriminator = getDiscriminatorTypeBuilder(disc);
-        if (discriminator == nullptr)
+        if (disc == nullptr)
         {
-            logError(XMLPARSER,
-                    "Error parsing union discriminator: Only primitive types allowed (found type " << disc << ").");
+            logError(XMLPARSER, "Error parsing union discriminator: 'type' attribute not found.");
             ret = XMLP_ret::XML_ERROR;
         }
         else
         {
-            p_dynamictypebuilder_t typeBuilder = types::DynamicTypeBuilderFactory::get_instance()->create_union_builder(
-                discriminator);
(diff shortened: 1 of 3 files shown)


PavelGuzenfeld:
Ping for review — no activity since submission. Happy to rebase or adjust if needed.

PR #6425, opened 2026-06-09
bound key material lengths in KeyMaterialCDRDeserialize

alhudz:
## Description

`KeyMaterialCDRDeserialize` parses the `dds.cryp.keymat` binary property carried in a `CryptoToken` sent by a remote participant, so every length byte in it crosses a trust boundary.

`Repro:` join a secured domain and send a participant/datawriter/datareader `CryptoToken` whose first sequence length byte (`master_salt`) is `0xFF` followed by enough bytes.
`Cause:` each `key_len` is read straight off the wire and handed to `memcpy` into the 32-byte `std::array` members (`master_salt`, `master_sender_key`, `master_receiver_specific_key`) with no check, and the CDR buffer size is never validated before indexing, so a `key_len` up to 255 writes past the destination (ASan reports a `stack-buffer-overflow`, `WRITE of size 255`).
`Fix:` validate the available CDR size at each step and reject any `key_len` larger than its destination array, returning `false` so the three `set_remote_*_crypto_tokens` callers drop the malformed token.

The validation lives in the deserialiser itself rather than the callers, since the callers only have the opaque token bytes.

## Contributor Checklist

### DIFF
--- a/src/cpp/security/cryptography/AESGCMGMAC_KeyExchange.cpp
+++ b/src/cpp/security/cryptography/AESGCMGMAC_KeyExchange.cpp
@@ -140,7 +140,12 @@ bool AESGCMGMAC_KeyExchange::set_remote_participant_crypto_tokens(
     }
 
     KeyMaterial_AES_GCM_GMAC keymat;
-    KeyMaterialCDRDeserialize(keymat, &plaintext);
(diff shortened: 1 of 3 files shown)


alhudz:
gentle ping

PR #6477, opened 2026-07-20
Guard IOKit main port usage on macOS

Tobias-Fischer:
This is part of an effort to contribute RoboStack downstream patches back upstream.

Origin: RoboStack patch/ros-rolling-fastdds.osx.patch, authored by Daisuke Nishimatsu.

Apple documents both kIOMasterPortDefault and kIOMainPortDefault in IOKit. kIOMainPortDefault is the current spelling for newer macOS SDKs, while kIOMasterPortDefault is still needed when targeting older macOS versions. Guarding the constant with MAC_OS_X_VERSION_MIN_REQUIRED avoids referencing the newer symbol for older deployment targets while retaining the deprecation fix for modern builds.

References:
- https://developer.apple.com/documentation/iokit/kiomasterportdefault
- https://developer.apple.com/documentation/iokit/kiomainportdefault

### DIFF
--- a/src/cpp/utils/Host.cpp
+++ b/src/cpp/utils/Host.cpp
@@ -24,6 +24,7 @@
 #include <windows.h>
 #include <process.h>
 #elif defined(__APPLE__)
+#include <AvailabilityMacros.h>
 #include <IOKit/IOKitLib.h>
 #else
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


Tobias-Fischer:
FYI @wep21 @traversaro, this was opened from the RoboStack upstreaming tracker: https://github.com/RoboStack/robostack.github.io/issues/16

Source patch: https://github.com/RoboStack/ros-rolling/blob/main/patch/ros-rolling-fastdds.osx.patch

PR #6487, opened 2026-07-23
Fix FTBFS against OpenSSL 4.0

### LINKED ISSUE #6432: Build failure with OpenSSL 4.0
This bug was filed using automation as part of the Debian OpenSSL 4.0 migration effort. Please ignore if the build issue is already fixed.

More details and build logs in the Debian bug report:
https://bugs.debian.org/1138444

amrelsh:
1. `<asio/ssl.hpp>` pulls in the deprecated `rfc2818_verification`, which no
   longer compiles on OpenSSL 4.0. so include only the
   `asio/ssl` headers we need.

2. `X509_get_subject_name()`  returns a `const` pointer, so store its
   result in a `const` pointer to avoid conversion errors.

Fixes #6432

@Mergifyio backport 3.3.x

### DIFF
--- a/src/cpp/rtps/transport/TCPAcceptorSecure.h
+++ b/src/cpp/rtps/transport/TCPAcceptorSecure.h
@@ -20,7 +20,7 @@
 #endif // ifdef OPENSSL_API_COMPAT
 #define OPENSSL_API_COMPAT 10101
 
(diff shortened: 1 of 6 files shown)


amrelsh:
@MiguelCompany Hello, whenever you got the time can you take a look at this? Thanks!

PR #6490, opened 2026-07-29
[25538] Fix datasharing infinite loop

MiguelCompany:
## Description

Fixes an infinite spin in `ReaderPool::get_next_unread_payload()`: when a payload's sequence number no longer looked newer than the last one delivered, the code discarded it without advancing the read pointer, so it kept re-examining the same shared-memory slot forever, hanging `DataSharingListener::stop()`'s thread join on reader teardown.

Also adds a defensive `is_running_` check in the listener's drain loop, plus a deterministic blackbox regression test that reproduces the hang via a rejected sample and verifies the fix.

The meaning of the return value of `process_data_msg` has been changed, with `false` now meaning "Try again later".
The only place where the return value is checked is in `DataSharingListener::process_new_data()`, and the changed method is part of an internal interface, so there is no user behavior change.

With this change, we can now break the loop when the reader informs us to "Try again later", and we don't advance the payload index in that case. Please note the subtle change from `>=` to `>` in `ReaderPool.hpp`, which allows processing the same sequence number several times.

Includes an unrelated one-line fix for a missing include in the TypeObjectRegistry GTest mock that only surfaces as a build failure on newer GCC/libstdc++ (reproduced with GCC 15.2.0).

@Mergifyio backport 3.2.x 2.14.x

## Contributor Checklist

## Reviewer Checklist

### DIFF
--- a/src/cpp/rtps/DataSharing/DataSharingListener.cpp
+++ b/src/cpp/rtps/DataSharing/DataSharingListener.cpp
@@ -162,7 +162,7 @@ void DataSharingListener::process_new_data ()
 
         uint64_t last_payload = pool->end();
         bool has_new_payload = true;
-        while (has_new_payload)
+        while (has_new_payload && is_running_.load())
         {
             CacheChange_t ch;
             SequenceNumber_t last_sequence = c_SequenceNumber_Unknown;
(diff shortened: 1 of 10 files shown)


zesk1999 (approved):
LGTM with green CI.

MiguelCompany:
@Mergifyio backport 3.2.x 2.14.x

